PR #7, opened 2017-06-13
Webpack src alias

Austio:
Noticed in tests that we were digging up quite a few levels to grab files from source directory.

Any thoughts on adding same for `test/resources`?


eddyerburgh:
Actually, hold fire on adding aliases for resources. The aliases aren't working correctly when I try to debug Karma - https://github.com/karma-runner/karma/issues/2708

eddyerburgh:
Ok, ignore the previous post - it's working after restarting chrome 👍

PR #13, opened 2017-07-04
docs(getting-started): Initial draft of getting-started guide

codebryo:
This would be my initial draft of the getting started guide, based on the API documentation provided. I aimed for making it easy to read and accessible.

There are some questions though:
- Is the language fine? Or do we aim to make it more formal? (I personally prefer a very communicative writing style)
- The async DOM update section is based on the fact that `html()` takes care of the async operation (Vue.nextTick) for us. Not sure this is entirely true for the initial release.

Also, let me know if there should be something in this guide I am totally missing.
I tried to follow through on the `chai` syntax for expectations.

Best,
Roman

### DIFF
--- a/docs/en/getting-started.md
+++ b/docs/en/getting-started.md
@@ -1,1 +1,137 @@
 # Getting Started
+
+Get up and running on with the official `vue-test-utils` is straight forward.
+Use either `npm` or `yarn` to install the package:
+
+```shell
+# npm
+$ npm install --save vue-test-utils
+
+# yarn
+$ yarn add vue-test-utils
+```
+
+Let's test a simple Vue component like a counter:
+
+```js
+// counter.js
+
+export default {
+  template: `
+    <div>
+      {{ count }}
+      <button @click="increment">Increment</button>
+    </div>
+  `,
+
+  data () {
+    return {
+      count: 0
+    }
+  },
+
+  methods: {
+    increment() {
+      this.count++
+    }
+  }
+}
+
+```
+
+We have a bunch of nice things we could test here, so let's see what `vue-test-utils` can do for you. The general concept of interacting with a your Vue component through the utils is a wrapper, with a lot of useful helpers under its belt.
+
+**1. How to Wrap a component**
+
+```js
+// Import the mount() method from the test utils
+// and the component you want to test
+
+import { mount } from 'vue-test-utils'
+import Counter from './counter'
+
+// Now mount the component and there you have your wrapper
+const wrapper = mount(Counter)
+```
+
+So that was quite easy to do right, so let's start with our first spec.
+Maybe you are curious about how the rendered HTML output of the component looks like.
+
+**2. Test rendered HTML output of the component**
+
+```js
+// For getting the html output, the wrapper provides
+// a html() method for you.
+
+describe('Counter', () => {
+  it('renders the correct markup', () => {
+    expect(wrapper.html()).to.equal('<div>0<button>Increment</button></div>')
+  })
+
+  // it's also easy to check for the existence of elements
+  it('has a button', () => {
+    expect(wrapper.contains('button')).to.equal(true)
+  })
+})
+```
+
+That was easy as well, so let's step up the game.
+
+**3. Component data**
+
+Changing the data of the component can be quite useful for efficient testing.
+
+```js
+
+describe('Data interactions', () => {
+  beforeEach(() => {
+    wrapper.setData({count: 10})
+  })
+
+  it('should be set to 10', () => {
+    // Get the current data of the Component with the data() method
+    expect(wrapper.data().count).to.equal(10)
+  })
+})
+
+```
+
+**4. Interactions**
+
+This section will introduce two important functions.
+`find()` to find an HTML element inside you components HTML output and `trigger()` will allow for firing general events.
+
+```js
+
+describe('Trigger an event', () => {
+  it('button should increment the count', () => {
+    expect(wrapper.data().count).to.equal(0)
+    let button = wrapper.find('button')
+    button.trigger('click')
+    expect(wrapper.data().count).to.equal(1)
+  })
+})
+
+```
+
+**5. Handle async DOM updates**
+
+As you probably know, Vue updates the DOM based on an internal `tick` to prevent unnecessary rerenders if a bunch of data gets changed. That's why `Vue.nextTick(...)` allows you to verify that DOM changes have been applied correctly. Luckily this is handled for you by the `.html()` method.
+
+```js
+
+describe('DOM updates', () => {
+  it('html() should account for async DOM updates', () => {
+    expect(wrapper.html()).to.equal('<div>0<button>Increment</button></div>')
+    wrapper.setData({ count: 23 })
+    expect(wrapper.html()).to.equal('<div>23<button>Increment</button></div>')
+  })
+})
+
+```
+
+## What Next
+
+Of course there is a lot more discover for you in `vue-test-utils` and it's recommended to go through the list of the [provided API](SUMMARY.md)
+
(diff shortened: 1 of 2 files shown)


docs/en/getting-started.md
```diff
@@ -1 +1,137 @@
 # Getting Started
+
+Get up and running on with the official `vue-test-utils` is straight forward.
```
callumacrae: Maybe try "Getting up and running with the official `vue-test-utils` package is straight forward." or "Getting up and running with `vue-test-utils` is straight forward." - there's a couple words missing as it is now

docs/en/getting-started.md
```diff
@@ -1 +1,137 @@
...
+  methods: {
+    increment() {
+      this.count++
+    }
+  }
+}
+
+```
+
+We have a bunch of nice things we could test here, so let's see what `vue-test-utils` can do for you. The general concept of interacting with a your Vue component through the utils is a wrapper, with a lot of useful helpers under its belt.
```
callumacrae: "interacting with a your" there's an extra word here

docs/en/getting-started.md
```diff
@@ -1 +1,137 @@
...
+    expect(wrapper.data().count).to.equal(10)
+  })
+})
+
+```
+
+**4. Interactions**
+
+This section will introduce two important functions.
+`find()` to find an HTML element inside you components HTML output and `trigger()` will allow for firing general events.
```
callumacrae: "you components HTML" -> "your component's HTML"

docs/en/getting-started.md
```diff
@@ -1 +1,137 @@
...
+    button.trigger('click')
+    expect(wrapper.data().count).to.equal(1)
+  })
+})
+
+```
+
+**5. Handle async DOM updates**
+
+As you probably know, Vue updates the DOM based on an internal `tick` to prevent unnecessary rerenders if a bunch of data gets changed. That's why `Vue.nextTick(...)` allows you to verify that DOM changes have been applied correctly. Luckily this is handled for you by the `.html()` method.
```
callumacrae: Generally better to avoid using phrases like "As you probably know": makes people feel stupid if they don't!

https://spin.atomicobject.com/2014/09/09/never-use-the-word-obviously/

docs/en/getting-started.md
```diff
@@ -1 +1,137 @@
...
+  })
+})
+
+```
+
+## What Next
+
+Of course there is a lot more discover for you in `vue-test-utils` and it's recommended to go through the list of the [provided API](SUMMARY.md)
+
+The `vue-test-utils` aim to play nice with your testrunner, should you run into any issues though, you'll find the official repository here: [vue-test-utils](https://github.com/vuejs/vue-test-utils)
```
callumacrae: Thank you for not making "here" the link 😎

eddyerburgh:
Looks great Roman.

The tests have failed because the linting on the docs has failed. A few of the errors are legitimate, but the no-undef errors should be fixed. 

Do you mind updating the `.eslintrc` inside /docs.

If you add a globals section with wrapper and expect included, that will fix most of the errors:

```
"globals": {
  "wrapper": true,
  "expect": true
}
```

For the `no-unused-vars` error on line 54, could you add a small example of using a wrapper underneath? Maybe just:
```
Wrapper.is(Container) // returns true!
```

If you feel this breaks the flow, then we could consider overriding this rule in `/docs/.eslintrc`, although I think this is useful

TheJaredWilcurt:
### Comment on usage of Markdown

* * *

If you use 
```md
1. asdf
1. qwer
```
it will create
```html
<ol>
  <li>asdf</li>
  <li>qwer</li>
</ol>
```

* * *

If you use
```md
**1. asdf**

**1. qwer**
```
it will create
```html
<p><strong>1. asdf</strong></p>
<p><strong>1. qwer</strong></p>
```

* * *

If you use
```md
#### 1. asdf
#### 1. qwer
```
it will create
```html
<h4 id="1-asdf">1. asdf</h4>
<h4 id="1-qwer">1. qwer</h4>
```

Which means, that if someone is in a chatroom or help site trying to understand something related to this project, someone can link them to:

* https://github.com/vuejs/vue-test-utils/documentation/getting-started.md#1-qwer

And it will load the page and scroll down automatically to the relevant data.

* * *

However it would be better if you don't include a number in the header, since things innevitably get reorganized, added, removed, shifted around. And that means things will need to be renumbered, and that will break links.

```md
#### asdf
#### qwer
```

```html
<h4 id="asdf">asdf</h4>
<h4 id="qwer">qwer</h4>
```

This way if you move the content around or update it you'll have this:

```md
#### asdf
#### zxcv
#### qwer
```
and not 
```
#### 1. asdf
#### 3. zxcv
#### 2. qwer
```

codebryo:
Thank you for the details. I usually prefer the `strong` tags for highlighting, but you make a good point regarding linking to parts of the documentation by using `#`.
I will bake that in as well :)

docs/en/getting-started.md
```diff
@@ -1 +1,154 @@
...
+
+```
+
+### Handle async DOM updates
+
+Vue updates the DOM based on an internal `tick` to prevent unnecessary rerenders if a bunch of data gets changed. That's why `Vue.nextTick(...)` allows you to verify that DOM changes have been applied correctly and to pass in a callback to run right after that. Luckily this async DOM update is taked care of by the `.html()` method.
+
+```js
+
+describe('DOM updates', () => {
```
eddyerburgh: vue-test-utils has an `update` method that is called inside setData.  `html` doesn't update the DOM.

So we should use `update` as an example instead. Here's the test- https://github.com/vuejs/vue-test-utils/blob/master/test/integration/specs/mount/Wrapper/update.spec.js#L6 . There's probably a better example to use than that though

docs/en/getting-started.md
```diff
@@ -1 +1,154 @@
...
+      this.count++
+    }
+  }
+}
+
+```
+
+This component is very basic in it's capabilities and scope,
+but never the less it allows for things we could test here. The general concept of interacting with a Vue component through the `vue-test-utils` is using a wrapper. This wrapper comes with a lot of useful helpers under its belt that should make writing even complex specs a breeze. But let's start simple:
+
```
eddyerburgh: There should probably be a section on setting up with JSDOM/ running the tests in a browser. As they won't work if not in a browser

docs/en/getting-started.md
codebryo: Okay, that's good to know thanks. Will work that in.

docs/en/getting-started.md
codebryo: Sounds good, I was expecting something like that. Will try to find a way to make that clear :)

codebryo:
@eddyerburgh Baked in your feedback. I hope that covered your references. :)

docs/en/getting-started.md
codebryo: I added a **Notice** block that links to some testrunners. I'd think about adding some specific guides extra. Setting up with jest, Karma, ava... you name it :)

eddyerburgh: Yeah, I think we should add guides and link to them here when we have some set up.

docs/en/getting-started.md
```diff
@@ -1 +1,158 @@
...
+```shell
+# npm
+$ npm install --save vue-test-utils
+
+# yarn
+$ yarn add vue-test-utils
+```
+
+**Notice:** To test Vue components and to make use of the `vue-test-utils` run the specs in a browser with a testrunner like [Karma](https://karma-runner.github.io/1.0/index.html) or node based runners that support virtual DOMs like `jsDOM` (e.g.: [jest](https://facebook.github.io/jest/), [ava](https://github.com/avajs/ava))
```
eddyerburgh: I think the wording could be improved to make it clear that any test runner in node can be used with jsDOM, but that's a minor detail

codebryo: yeah, I ment those 2 as examples. 
> node based runners that support virtual DOMs like `jsDOM`

should tell em, but yes. Always improve :)

PR #14, opened 2017-07-08
Feat: add callback function to trigger method

sqal:
Sometimes when testing our components, default Vue event modifiers may be insufficient so we need to have a way to modify event object before it gets dispatched.

### DIFF
--- a/docs/en/api/wrapper-array/trigger.md
+++ b/docs/en/api/wrapper-array/trigger.md
@@ -1,7 +1,8 @@
-# trigger(eventName)
+# trigger(eventName, process)
 
 - **Arguments:**
-  - `{string} eventName
+  - `{string} eventName`
+  - `{string} process` - optional callback allowing to modify event object
 
 - **Usage:**
 
--- a/docs/en/api/wrapper-array/trigger.md
+++ b/docs/en/api/wrapper-array/trigger.md
@@ -2,7 +2,7 @@
 
 - **Arguments:**
   - `{string} eventName`
-  - `{string} process` - optional callback allowing to modify event object
+  - `{Function} process` - optional callback allowing to modify event object
 
 - **Usage:**
 
--- a/docs/en/api/wrapper/trigger.md
+++ b/docs/en/api/wrapper/trigger.md
@@ -1,7 +1,8 @@
-# trigger(eventName)
+# trigger(eventName, process)
 
 - **Arguments:**
   - `{string} eventName`
+  - `{string} process` - optional callback allowing to modify event object
 
 - **Usage:**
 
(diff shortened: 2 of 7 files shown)


eddyerburgh:
What are the advantages of using a callback over an object to augment the event object?

```js
button.trigger('mousedown', (e) => {
  e.button = 0
})
```

```js
button.trigger('mousedown', {
  button: 0
})
```

sqal:
@eddyerburgh I think if we wanted to extend the event object, we won't be able to do something like this:

`eventObject = Object.assign(eventObject, objToExtend)`

`trigger(document, 'click', { defaultPrevented: true })`

won't work, it will throw - Uncaught TypeError: Cannot assign to read only property 'defaultPrevented' of object '#<Event>'

Also, i think we should replace `new Event(event)` with `new Event(event, { bubbles: true, cancelable: true })`   because currently doing

```
triggerEvent(document, 'click', (e) => {
  e.preventDefault()
})
```

does not prevent the default action, reproduction here -> https://jsfiddle.net/z01xhwcu/

eddyerburgh:
I think that those read only properties can't be overwritten either when edited in a callback or extending the event object. 

Just ran this test in the browser -

```js
function editEvent(cb) {
  const event = new Event('click')
  cb(event)
  console.log(event)
}

editEvent(e => e.defaultPrevented = true)
```

Does not overwrite `defaulyPrevented`.

Can you get a working example where the callback API does something that extending with an object can't do?

> Also, i think we should replace new Event(event) with new Event(event, { bubbles: true, cancelable: true }) because currently doing

Sounds good, can you make the change in this PR?

sqal:
> I think that those read only properties can't be overwritten either when edited in a callback or extending the event object.

Correct.

> Can you get a working example where the callback API does something that extending with an object can't do?

The example with preventDefault() :) Let's say I have method in my component:

```
someMethod(event) {
  if (event.defaultPrevented) return;
  // otherwise, do something
}
```

with the callback fn you can call `e.preventDefault()` to test this method, it won't be possible without callback since you cannot override defaultPrevented value.

eddyerburgh:
Ah yeah of course 😛 

So personally I prefer an object API over the callback API. We have more control over the input, and we can throw custom errors for common properties that are read only. 

Although the callback API you've suggested is a lot less maintenance. But I can imagine getting lots of questions about not being able to set target. We can warn that target can't be set if we use an options object

For the preventDefault() example, we could have something like this:

```js
triggerEvent(document, 'click', {
  preventDefault: true
})
```

What do you think?

sqal:
Sure, sounds good  and i am fine with this :) So, I guess we're done with this PR and you'll implement these changes on your own?

eddyerburgh:
You could add some more commits to this PR, or create a new one.

I won't have time to get around to this in the next few days so a PR would really be appreciated! 😄

eddyerburgh:
I can add this tomorrow if you aren't working on it

sqal:
I am sorry, but I will be busy at work, so go ahead and add it yourself. Thank you <3

eddyerburgh:
closed with 6ced351ee35e8c8abb54532f90b649a220185124

Thanks for the PR 😊

PR #19, opened 2017-07-20
Documentation: Replace references to data() with vm

codebryo:
As `data()` is no longer valid, I replaced those references with `vm`.
I also updated the content of the getting started guide accordingly.

Best,
Roman


eddyerburgh:
Ah no, sorry @codebryo. I've added a dev branch, so this will get overwritten in master when we release.

Would you be able to resubmit your PR against dev?

PR #21, opened 2017-07-23
docs: replace .data() with .vm and updated the getting started guide

### LINKED ISSUE #16: data() is not available
Not sure this is on purpose, but it seems that 
`wrapper.data()` is no longer valid.

Is it preferred to use the `vm` object directly?
It's still in the API docs of `setData()` that's why I'm asking.

I checked, and it's no longer in the source code, so I guess it was removed.
Should I update the documentation accordingly?

posva:
I **just cherry picked the commit from #19**
It should close #16

### DIFF
--- a/docs/en/getting-started.md
+++ b/docs/en/getting-started.md
@@ -91,7 +91,7 @@ That was easy as well, so let's step up the game.
 
 ### Component data
 
-Changing the data of the component can be quite useful for efficient testing. The method `setData({...})` is ment for changing the data and `data()` returns the current data object of the component.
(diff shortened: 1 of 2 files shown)


docs/en/getting-started.md
```diff
@@ -91,7 +91,7 @@ That was easy as well, so let's step up the game.
 
 ### Component data
 
-Changing the data of the component can be quite useful for efficient testing. The method `setData({...})` is ment for changing the data and `data()` returns the current data object of the component.
+Changing the data of the component can be quite useful for efficient testing. The method `setData({...})` is ment for changing the data on the instance. You can interact with the instance directly using the `vm` key. As Vue automatically sets all data values and computed properties as getters on the root instance, we can access those values straight away.
```
eddyerburgh: Just spotted a typo:

> is ment for

PR #23, opened 2017-08-01
Added guide to get started with jest

codebryo:
Added a guide to get started with Jest.
Also added a reference to my repo from the vue conf 2017 talk with Jest examples, slides and more details around it. Hope that's alright.

Thank you,
Roman

### DIFF
--- a/docs/en/guides/using-with-jest.md
+++ b/docs/en/guides/using-with-jest.md
@@ -52,13 +52,13 @@ That's pretty much everything necessary to do before writing the first specs.
 
 ### Where should my tests live
 
-By default, jest will pick up all files that have a `.spec.js` or `.test.js` extension. If this is not fitting your needs, it's possible [to chang the testRegex](https://facebook.github.io/jest/docs/en/configuration.html#testregex-string) in the config section in the `package.json` file.
+By default, jest will pick up all files that have a `.spec.js` or `.test.js` extension. If this does not fit your needs, it's possible [to chang the testRegex](https://facebook.github.io/jest/docs/en/configuration.html#testregex-string) in the config section in the `package.json` file.
 
 Jest recommends to create a `__spec__` folder, but feel free to do as you like. Just know ahead of time, that when using the [snapshot](https://facebook.github.io/jest/docs/en/snapshot-testing.html#content) feature, snapshots will get stored in a `__snapshot__` folder.
 
 ### Example Spec
 
-If you are familiar with Jasmin, or similar test libraries you should feel at home in Jest right away. Many useful assertions are in place, so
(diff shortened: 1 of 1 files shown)


docs/en/guides/using-with-jest.md
```diff
@@ -0,0 +1,81 @@
...
+    }]
+  ]
+}
+```
+
+That's pretty much everything necessary to do before writing the first specs.
+
+### Where should my tests live
+
+By default, jest will pick up all files that have a `.spec.js` or `.test.js` extension. If this is not fitting your needs, it's possible [to chang the testRegex](https://facebook.github.io/jest/docs/en/configuration.html#testregex-string) in the config section in the `package.json` file.
```
eddyerburgh: *is not fitting*, should be *does not fit* 👍

docs/en/guides/using-with-jest.md
```diff
@@ -0,0 +1,81 @@
...
+
+Jest recommends to create a `__spec__` folder, but feel free to do as you like. Just know ahead of time, that when using the [snapshot](https://facebook.github.io/jest/docs/en/snapshot-testing.html#content) feature, snapshots will get stored in a `__snapshot__` folder.
+
+### Example Spec
+
+If you are familiar with Jasmin, or similar test libraries you should feel at home in Jest right away. Many useful assertions are in place, so
```
eddyerburgh: small typo - Jasmin should be *Jasmine*

docs/en/guides/using-with-jest.md
```diff
@@ -0,0 +1,81 @@
...
+enjoy writing specs.
+
+```js
+import { mount } from 'vue-test-utils'
+import Component from './component'
+
+describe('Component', () => {
+  test('it's a Vue instance', () => {
```
eddyerburgh: you're breaking the string with the apostrophe in it's - maybe change to *is a Vue instance*, or some string that doesn't have an apostrophe

There's a script that lints the docs - `npm run lints:docs`

eddyerburgh:
Great, thanks Roman 😃

I think it's fine adding a reference to your example 👍

docs/en/guides/using-with-jest.md
codebryo: Happy you are a native speaker :D

codebryo:
Addressed feedback. Thank you Edd :)

eddyerburgh:
Ah, this PR is against master. You need to PR against the dev branch.

Can you remake the PR please?

Also note the commit convention - https://github.com/vuejs/vue/blob/dev/.github/CONTRIBUTING.md

`docs: description goes here` is valid
`(docs): description goes here` is not valid

I'll squash the PR and add a valid commit message 👍

PR #38, opened 2017-09-03
fix(docs): correct docs/api links

fritx:
Before this PR, the links are 404.

### DIFF
--- a/docs/en/api/README.md
+++ b/docs/en/api/README.md
@@ -1,3 +1,3 @@
-## [mount](/docs/en/api/mount.md)
-## [shallow](/docs/en/api/shallow/README.md)
-## [selectors](/docs/en/api/selectors.md)
+## [mount](/api/mount.md)
+## [shallow](/api/shallow.md)
+## [selectors](/api/selectors.md)


docs/en/api/README.md
```diff
@@ -1,3 +1,3 @@
-## [mount](/docs/en/api/mount.md)
-## [shallow](/docs/en/api/shallow/README.md)
-## [selectors](/docs/en/api/selectors.md)
+## [mount](/api/mount.md)
```
posva: It would be even better to have `api/mount.md` (relative links). That way it'd be easier to build translations (and it's shorter 😆 )

fritx: Oh yes I also thought about that, but I noticed the other links were all `/api/xxx.md`, 😆 
so I had to keep them consistent.

image

fritx: I served the gitbook, and found it just worked, without problems.

fritx: Seems the other links were all absolute paths, so I kept them consistent.

image

PR #56, opened 2017-09-29
fix bug

webfrontzhifei:
the bug is : when you test the style a arbitrary name，like "margin-top333", the old function will return true, because the value of 'elStyle' and 'mockNodeStyle' are all ''undefined"

### DIFF
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -133,7 +133,7 @@ export default class Wrapper implements BaseWrapper {
 
     const elStyle = window.getComputedStyle(this.element)[style]
     const mockNodeStyle = window.getComputedStyle(mockNode)[style]
-    return elStyle === mockNodeStyle
+    return elStyle && mockNodeStyle && elStyle === mockNodeStyle
   }
 
   /**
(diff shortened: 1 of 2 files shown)


webfrontzhifei:
it('Message component has style margin-top2: 10', () => {
     expect(cmp.find(Message).hasStyle('margin-top2', '10')).toBe(true)
   })

   The test Result is :
   ✓ Message component has style margin-top2: 10 (35ms)

eddyerburgh:
Can you add that to the tests/unit/specs/Wrapper/hasStyle.spec.js in this PR

webfrontzhifei:
ok, i have added it.

src/wrappers/wrapper.js
```diff
@@ -133,7 +133,7 @@ export default class Wrapper implements BaseWrapper {
 
     const elStyle = window.getComputedStyle(this.element)[style]
     const mockNodeStyle = window.getComputedStyle(mockNode)[style]
-    return elStyle === mockNodeStyle
+    return elStyle && mockNodeStyle && elStyle === mockNodeStyle
```
eddyerburgh: You need to wrap this whole return statement in parentheses and cast them to a Boolean:
```
return !!(elStyle && mockNodeStyle && elStyle === mockNodeStyle)
```

webfrontzhifei: ok, i have fixed it

PR #64, opened 2017-10-02
fix #63 isEmpty on element with empty slot

### LINKED ISSUE #63: isEmpty returns false for element containing an empty slot
Elements with slots always return false from isEmpty even if there is no content. i.e. 
```
<div>
    <slot></slot>
</div>
```

philstavri:
closes #63

Mounting element with slot with no content sets vnode children to [], which makes isEmpty return false (currently checks only for undefined). This duplicates behaviour in [Avoriaz](https://github.com/eddyerburgh/avoriaz/pull/128).

### DIFF
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -248,7 +248,7 @@ export default class Wrapper implements BaseWrapper {
    * Checks if node is empty
    */
   isEmpty (): boolean {
-    return this.vnode.children === undefined
+    return this.vnode.children === undefined || this.vnode.children.length === 0
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/wrappers/wrapper.js
```diff
@@ -248,7 +248,7 @@ export default class Wrapper implements BaseWrapper {
    * Checks if node is empty
    */
   isEmpty (): boolean {
-    return this.vnode.children === undefined
+    return this.vnode.children === undefined || this.vnode.children.length === 0
```
lusarz: Isn't enough to:
`return !this.vnode.children`

Issue #67, opened 2017-10-02
Find / FindRef: wrapper function to find element based on ref


matt-oconnell:
**Suggestion for additional functionality**

Currently, there is no functionality built into `vue-test-utils` to select elements within a component based on `$refs` and return them within a Wrapper. Our test suite is built around the idea that we can easily tag elements we need to hook onto within tests using `$ref` rather than some arbitrary "marker class."

I think it might be helpful to include a specific method on the wrapper like `findRefs`

I'm not sure if this is a widely used strategy but we've found it helpful. Template structures change often and class selectors can easily break or target the wrong element when these changes happen.

eddyerburgh:
I think this is a good idea, but I don't think we should add an extra method that only finds refs.

We could extend find, so that it takes an object. 

```js
find({
  selector: 'refs',
  value: 'value' 
})
```

Then we could extend it to other selectors, like `props`.

```js
find({
  selector: 'props',
  value: 'value' 
})
```

matt-oconnell:
I wonder how many alternate use cases we would have for a selector other than CSS, component name, or $ref. I'm not sure if component with prop: `x`would ever be used as a selector, but I could be wrong.

matt-oconnell:
But, if we do want to go with the selector/value find/findAll functions, I could start a PR. Initially it could look for $props / $refs. I guess the tough part would be matching `$props` because the values could be `any`

eddyerburgh:
I think we should just add support for refs at the moment ☺

matt-oconnell:
@eddyerburgh To make it a little more concise, how do you feel about:

```js
find({ ref: 'myRefName' })
```

eddyerburgh:
Yep that looks good 👍

matt-oconnell:
Great, updating the PR to reflect this. Currently, the validation error messages for find/findAll display: `wrapper.find/findAll() must be passed a valid CSS selector or a Vue constructor`. As we update the API to accept this new options object, how should we refer to it? 

If we're going to start with refs for now, we could use:

`wrapper.find() must be passed a valid CSS selector, Vue constructor, or ref object`

Or `wrapper.find() must be passed a valid CSS selector, Vue constructor, or options object`

My only concern with the latter is that it may be too vague.

eddyerburgh:
Ah that's a good point.

If we add this then we need to:
- Update all methods that take a selector: `is`, `contains`, `find`, `findAll`
- Update the selectors page to include the refs object
[...]

PR #68, opened 2017-10-02
Wrapper: find/findAll accepts ref options object

matt-oconnell:
This PR relates to https://github.com/vuejs/vue-test-utils/issues/67

It adds the `.ref('someRefName')` method to the main `wrapper`.

This returns a `WrapperArray` with wrappers for any `vnode`s that have the specified ref. Additionally, `.ref` will only match `vnodes` with refs that exist on the `wrapper.vm.$refs`.

1) Do we want to just have a single `ref` function or do we want to split it into `findRefs` / `findAllRefs`? Right now, this only returns a wrapper array.

2) I haven't implemented tests yet because I wanted to make sure I was going in the right direction first.

3) The find-matching-vnodes-by-ref.js file that is added here uses some of the functionality from the existing find-matching-vnodes.js file. Should I pull out some of those functions into a more generic vnode traversal file or something?


eddyerburgh:
1. I'm not sure on the API.  Ideally we would be able to extend find and findAll to handle refs. Let's discuss in #67
2. That's fine, but we'll need some before me merge
3. Yes 🙂

src/lib/find-vnodes-by-ref.js
```diff
@@ -0,0 +1,15 @@
...
+import { removeDuplicateNodes, findAllVNodes } from './vnode-utils'
+
+function nodeMatchesRef (node: VNode, refName: string): boolean {
+  return node.data && node.data.ref === refName
+}
+
+export default function findVNodesByRef (vNode: VNode, refName: string): Array<VNode> {
+  const nodes = findAllVNodes(vNode)
+  const refFilteredNodes = nodes.filter(node => nodeMatchesRef(node, refName))
+  // Only return ref matches from top level vNode
```
kaicataldo: For my own understanding, why do we not do a recursive check here? Might be helpful to document in the comment.

matt-oconnell: Updated. This is because we want to emulate the exact selector functionality of `wrappedComponent.$refs.myRef`.

If we do a recursive check, we may return a general result of `wrappedComponent.$refs.someComponent.$refs.someComponent.$refs.myRef` when we only want to grab ref elements on the top-level wrapper itself. If you need a sub-component ref, you'd need to find the subcomponent and get the ref off of it explicitly.

src/lib/find-vnodes-by-selector.js
```diff
@@ -0,0 +1,13 @@
...
+
+function nodeMatchesSelector (node: VNode, selector: string): boolean {
+  return node.elm && node.elm.getAttribute && node.elm.matches(selector)
+}
+
+export default function findVNodesBySelector (vNode: VNode, selector: string): Array<VNode> {
+  const nodes = findAllVNodes(vNode)
+  const filteredNodes = nodes.filter(node => nodeMatchesSelector(node, selector))
+  return removeDuplicateNodes(filteredNodes)
+}
```
matt-oconnell: This file includes all non-shared functionality from `find-matching-vnodes`

src/lib/vnode-utils.js
```diff
@@ -0,0 +1,28 @@
...
+export function removeDuplicateNodes (vNodes: Array<VNode>): Array<VNode> {
+  const uniqueNodes = []
+  vNodes.forEach((vNode) => {
+    const exists = uniqueNodes.some(node => vNode.elm === node.elm)
+    if (!exists) {
+      uniqueNodes.push(vNode)
+    }
+  })
+  return uniqueNodes
+}
```
matt-oconnell: This file includes all shared functionality from `find-vnodes-by-selector` and `find-vnodes-by-ref`

src/lib/find-vnodes-by-ref.js
kaicataldo: Cool, thanks for the 💯 explanation!

matt-oconnell:
Ok, @eddyerburgh, so I know we're waiting on the discussion to resolve in #67 about whether or not the find by $ref functionality should be implemented for all methods that take a "Selector," but this PR implements that functionality in the case that we do decide to move forward with that.

A few notes:

1) Importing flow-types from `vue` does not seem to work correctly. Selector was defined as
```js
declare type Selector = string | Component
```
but it looks like the global imports in `flow/vue.js` may not actually get imported. So, anywhere that accepted a Selector type was actually accepting `any`. What should be done about this?

2) `is` throws an error if an options object `{ ref: 'someName' }` is passed in because we can't assert that an element matches the $ref selector without access to the component that has that $ref property. (it would be like asking: `"is this component a ref of this component"`)

3) We throw errors when when trying to select by $ref on non-Vue components.

4) The naming of this new selector type is currently `RefSelector`. There may a better way to refer to it

5) Will need to update docs

eddyerburgh:
As you said, we should hold off until we've had more input from the community. But it looks great so far 🙂

1. I'll look into this
2. OK that makes sense
3. Good
4. I think that's fine
5. OK, are you able to do this?

matt-oconnell:
re: 5. Yep, this is in the PR now.

kaicataldo:
@eddyerburgh Do you have a link to the discussion that is happening around this? Would love to see this land!

eddyerburgh:
@kaicataldo here — https://github.com/vuejs/vue-test-utils/issues/67

eddyerburgh:
Hey @matt-oconnell , are you able to fix the merge conflicts? Then I'll merge

matt-oconnell:
Ok, resolved 👍

eddyerburgh:
Thanks @matt-oconnell 😀

Issue #76, opened 2017-10-05
Wrong link to docs


ankurk91:
This link is 404

https://github.com/vuejs/vue-test-utils/blob/b9b0aa9dde34b0e1b38dd0a485ad734c9c856ac5/src/mount.js#L18

PR #77, opened 2017-10-05
fix: link to docs, #76

ankurk91:
This PR fixe a link to docs.
Existing URL was throwing 404.

Correct URL should be -
https://vue-test-utils.vuejs.org/en/guides/common-tips.html

May be we should setup a test to traverse such links in code and make sure they respond in 200.

Ref #76


eddyerburgh:
Thanks for the PR 🙂

I think the test to traverse links would be a good idea. Are there solutions available already?

ankurk91:
I have used this one in past -
https://www.npmjs.com/package/markdown-link-check
Not sure about js files.

PR #87, opened 2017-10-11
Restore temporary change to original config after adding attrs/listeners

### LINKED ISSUE #75: Mocha-webpack setup, can't suppress Vue logging
Using the mocha-webpack setup, I have a required file `start.js` with the following code:

```
require('jsdom-global')();
const Vue = require('vue');

// disable Vue logging
Vue.config.productionTip = false;
Vue.config.devtools = false;
Vue.config.silent = true;

global.assert = require('chai').assert;
global.expect = require('chai').expect;
````

However, Vue is still logging out to the console,

image

Have I missed something?

re-fort:
This PR resolves #75.

### DIFF
--- a/src/lib/add-attrs.js
+++ b/src/lib/add-attrs.js
@@ -1,11 +1,12 @@
 import Vue from 'vue'
 
 export default function addAttrs (vm, attrs) {
+  const originalVueConfig = Vue.config
   Vue.config.silent = true
   if (attrs) {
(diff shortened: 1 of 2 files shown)


eddyerburgh:
Ah good spot. I realized I'd accidentally overwritten this the other day, but forgot to make the changes.

The failing tests are coming from a PR I made, so I'll merge and fix the tests.

PR #93, opened 2017-10-13
docs: Update readme

ankurk91:
* Add badges
     - ~~This will help people like me to know if there is a new version available~~
     - Add CI badge only
* Update install commands to get latest version
    - This will always download the latest version, even if it is beta


ankurk91:
@eddyerburgh
Done, dont forget to squash commits

Issue #99, opened 2017-10-14
Using setProps and wrapper.props return error not a function.


alexhernandez:
I was following the example on: https://vue-test-utils.vuejs.org/en/api/wrapper/setProps.html

And my code returns: `TypeError: wrapper.props is not a function`;

```javascript
import { mount } from 'vue-test-utils';
import { expect } from 'expect';
import Foo from './Foo.vue';

const wrapper = mount(Foo);
wrapper.setProps({ foo: 'bar' });
expect(wrapper.props().foo).toBe('bar');
```

For Debugging:

```javascript
{
  "scripts": {
    "test": "mocha-webpack --webpack-config ./node_modules/laravel-mix/setup/webpack.config.js --require spec/setup.js spec/**/*.spec.js"
  },
  "devDependencies": {
    "babel-plugin-transform-object-rest-spread": "^6.26.0",
    "cross-env": "^5.0.1",
    "expect": "^21.2.1",
    "jsdom": "^11.3.0",
    "jsdom-global": "^3.0.2",
    "laravel-mix": "^1.0",
    "mocha": "^4.0.1",
    "mocha-webpack": "^0.7.0",
    "vue-test-utils": "^1.0.0-beta.2"
  },
  "dependencies": {
    "vue": "^2.5.2",
    "vue-router": "^3.0.1",
    "vuelidate": "^0.6.1",
    "vuex": "^3.0.0",
    "vuex-router-sync": "^5.0.0"
  }
}

```

alexhernandez:
Just noticed if you change `wrapper.props()` to `wrapper.vm._props.foo` it works.

```javascript
const wrapper = mount(Foo);
wrapper.setProps({ foo: 'bar' });
expect(wrapper.props().foo).toBe('bar');
```

lmiller1990:
Is your prop is `required`, you may get an error, because the prop is not present on the initial mount.

You can also pass the props when you call `mount` like so:
``` js
// Your component
<script>
export default {
    props: {
      user: { 
        type: Object, 
        required: true
    }
}
```

``` js
// Test your component with required prop
const mockUser = {
  displayname: 'Test user'
}

const wrapper = mount(UserConversationDisplay, {   
  propsData: {
    user: mockUser
  }
})
```

eddyerburgh:
Thanks guys.

You're right, we need a page on using props. If anyone wants to make a PR, it would be greatly appreciated ☺.

FYI - `props` isn't a method in vue-test-utils

[...]

PR #102, opened 2017-10-15
Fix example and add example using setProps

lmiller1990:
In response to #99 

This fixes the otherwise not working example.

It also adds another example showing how to use `propsData` for a component that might have `required` props.

### DIFF
--- a/docs/en/api/wrapper/setProps.md
+++ b/docs/en/api/wrapper/setProps.md
@@ -19,7 +19,7 @@ wrapper.setProps({ foo: 'bar' })
 expect(wrapper.vm.foo).to.equal('bar')
 ```
 
-You can also pass a `propsData` object, which will initialize the Vue instance with passed values. This is useful if you have a component or instance that has props which are `required`.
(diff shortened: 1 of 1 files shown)


docs/en/api/wrapper/setProps.md
```diff
@@ -16,5 +16,33 @@ import Foo from './Foo.vue'
 
 const wrapper = mount(Foo)
 wrapper.setProps({ foo: 'bar' })
-expect(wrapper.props().foo).toBe('bar')
+expect(wrapper.vm.foo).to.equal('bar')
+```
+
+You can also pass a `propsData` object, which will initialize the Vue instance with passed values. This is useful if you have a component or instance that has props which are `required`.
```
eddyerburgh: Could you remove this section please"
```
This is useful if you have a component or instance that has props which are `required`.
```

I think the users will understand what it's useful for.

lmiller1990:
@eddyerburgh  Sure. I removed it

PR #106, opened 2017-10-16
Added support for CSS modules in hasClass

nblackburn:
Checks the className against $style where available.

- Allows for checking classes that were converted with css modules without needing to know what it was converted to.
- Only works when $style exists on the vm and a mapping is available for the target class.

> **DISCLAIMER**: Not a fan of defining a function argument but it's the cleanest solution.

### DIFF
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -98,15 +98,15 @@ export default class Wrapper implements BaseWrapper {
    * Asserts wrapper has a class name
    */
   hasClass (className: string) {
-    let targetClass = className;
+    let targetClass = className
 
     if (typeof targetClass !== 'string') {
       throwError('wrapper.hasClass() must be passed a string')
     }
 
-    // if $style is available and has a matching target, use that instead. 
-    if (this.vm && this.vm.$style && this.vm.$style[targetClass]) { 
-      targetClass = this.vm.$style[targetClass];
+    // if $style is available and has a matching target, use that instead.
+    if (this.vm && this.vm.$style && this.vm.$style[targetClass]) {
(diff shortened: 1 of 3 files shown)


nblackburn:
Made some tweaks so that the function variables are not getting redefined at the cost of consiceness.

nblackburn:
Sure, i was going to add one but didn't have time yesterday.

nblackburn:
Added a unit test

nblackburn:
No idea why that failed, just downloaded and ran it myself again and everything passed.

eddyerburgh:
The compatibility tests are failing, it looks like $style is implemented differently in Vue 2.0.x

To debug you can run the `npm run test:compatibility`

nblackburn:
Should be good to go now, sorry about that.

> **FYI**: You can supress all of the stuff being spewed out on console in the karma runner with the `client.captureConsole` property.

eddyerburgh:
And can you revert the test to the previous test, where you had a class but had a scoped style block

nblackburn:
@eddyerburgh I am not convinced the test was correct as to be able to use css modules, the class needs to be that of the mapped (found in `$style`) one rather than the original which to my understanding is not done automatically.

eddyerburgh:
The test was correct, you can see the example in the vue loader docs —https://vue-loader.vuejs.org/en/features/scoped-css.html

nblackburn:
@eddyerburgh this feature is for CSS modules, not scopped css.

eddyerburgh:
Ah sorry 👍 

You're right then, the test you have now looks correct

nblackburn:
Cool, thanks 👍

PR #112, opened 2017-10-19
fix: add support for vue.extended components + add test

feedthejim:
Hi,

Here's the PR. I haven't had time to do much testing. 

Cheers :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "sinon": "^2.3.2",
     "sinon-chai": "^2.10.0",
     "typescript": "^2.4.1",
-    "vue": "^2.4.4",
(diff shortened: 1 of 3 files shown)


eddyerburgh:
We run compatibility tests (npm run test:compatibility) against each major version of Vue, and the test you added failed on version 2.0.8

feedthejim:
Ok @eddyerburgh after investigating, this.$props came only after Vue 2.2.1, so i fixed my test

feedthejim:
@eddyerburgh when are you planning a new release ?

eddyerburgh:
I'll release on Sunday, I'm away from my laptop until then unfortunately

Issue #118, opened 2017-10-23
Support multiple classes in `hasClass`


lmiller1990:
Hi,

I noticed that `hasClass()` only works for single word classes. For example, I did something like:

``` js
<template>
  <div :class="[ todo.isComplete ? 'complete todo' : incomplete todo']">
    Text
  </div>
</template>
```

And because `hasClass()` does 

`return !!(this.element && this.element.classList.contains(className))`

You cannot find multiword classes. This was a little inconvenient since I mainly use semantic-ui, which uses longer multiword names like `ui eight wide compact grid`. Being able to assert the full class name makes my tests more understandable.

My proposal is to all `hasClass` to work with multiword classes. This seems like a pretty easy change and I'd like to tackle it if the maintainer(s) are okay with it.

callumacrae:
Multi-word classes are… not a thing. It's just multiple class names! With semantic-ui, are `class="complete todo"` and `class="todo complete"` the same thing?

If they're not the same thing, try `blabla.is('[class*="complete todo"]');`

lmiller1990:
@callumacrae  Thanks for the reply. In many cases the order does have an impact in semantic-ui, but I think those are mostly weird edge cases, probably not a good use case to argue for this feature.

I don't understand what your second comment means - I haven't seen that syntax, how can I use it to assert my wrapper has the class "complete todo" with the existing tools in vue-test-utils? Maybe I could pull another library in to do the comparison but other than doing two `hasClass()` calls, I don't see how I can achieve what I'm trying to.

If it's outside of the scope of vue-test-utils that's okay too, but I feel like I should be able to do `wrapper.hasClass('multiple classes')` and return true.

callumacrae:
`[class*="complete todo"]` is a CSS selector that would match any element where the `class` attribute contains `complete todo` (but not `todo complete`): https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

I agree that it would be useful for `hasClass` to match multiple classes - I was just pointing out that if you want to check for order, there's already a way to do that. A selector like the above is also probably how semantic-ui deals with it as well.

lmiller1990:
Right, thanks for the clarification. I was more concerned about checking for the full two classes being present, not so much the order - in this case 'incomplete todo' vs 'complete todo'. I know I can just workaround by checking for `incomplete` or `complete` in this case, but I still think it'd be a nice addition.

eddyerburgh:
[...]

PR #123, opened 2017-10-24
Allow `hasClass` to compare multiple classes

lmiller1990:
Addressing #118.

Instead of making a new variable, I could do the `.split` and `.every` check in the return statement, but I feel splitting it up is more readable. Thoughts?

### DIFF
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -111,8 +111,7 @@ export default class Wrapper implements BaseWrapper {
 
     const containsAllClasses = targetClass
       .split(' ')
-      .map(target => this.element.classList.contains(target))
-      .reduce((a, b) => a === b ? a : NaN)
+      .every(target => this.element.classList.contains(target))
 
     return !!(this.element && containsAllClasses)
   }
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -114,6 +114,7 @@ export default class Wrapper implements BaseWrapper {
       .every(target => this.element.classList.contains(target))
 
     return !!(this.element && containsAllClasses)
+    // return !!(this.element && this.element.classList.contains(targetClass))
   }
 
   /**
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -114,7 +114,6 @@ export default class Wrapper implements BaseWrapper {
       .every(target => this.element.classList.contains(target))
 
     return !!(this.element && containsAllClasses)
-    // return !!(this.element && this.element.classList.contains(targetClass))
   }
 
   /**
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -104,18 +104,16 @@ export default class Wrapper implements BaseWrapper {
       throwError('wrapper.hasClass() must be passed a string')
     }
 
-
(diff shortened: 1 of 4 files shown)


src/wrappers/wrapper.js
```diff
@@ -109,7 +109,11 @@ export default class Wrapper implements BaseWrapper {
       targetClass = this.vm.$style[targetClass]
     }
 
-    return !!(this.element && this.element.classList.contains(targetClass))
+    const containsAllClasses = targetClass
```
eddyerburgh: I think containsAllClasses should be a function, otherwise it will error if this.element is undefined, whereas at the moment the method returns false

test/unit/specs/mount/Wrapper/hasClass.spec.js
```diff
@@ -46,4 +46,10 @@ describe('hasClass', () => {
 
     expect(wrapper.hasClass('color-red')).to.equal(true)
   })
+
+  it('returns true when the element contains multiple classes', () => {
```
eddyerburgh: Can you add a test for when wrapper.element is undefined. We want to either handle the error, or returns false. Previous behavior was to return false

lmiller1990:
Thanks for the review @eddyerburgh .

I tried adding `wrapper.element = undefined` like test, but it doesn't seem to have the effect I expected - `wrapper.element` is still very much defined when `hasClass` executes. Any ideas?

I also noticed the fourth test in `hasClass.spec.js`:

``` js
it('returns false if wrapper does not have an element', () => {
  const compiled = compileToFunctions('<div />')
  const wrapper = mount(compiled)
  wrapper.element = null
  expect(wrapper.hasClass('not-class-name')).to.equal(false)
})
```

Does not seem to be doing what is expected either - removing the line `wrapper.element = null` does not seem to have any impact, that tests still passes either way.

eddyerburgh:
Sorry, element on a VueWrapper is a getter. You can do this by setting element on an element Wrapper to null - https://github.com/vuejs/vue-test-utils/blob/dev/test/unit/specs/mount/Wrapper/trigger.spec.js#L77

lmiller1990:
I'll give this a try now. What does this mean for the already existing `it('returns false if wrapper does not have an element'` test - isn't doing `wrapper.element = null` basically testing nothing?

eddyerburgh:
Yep, you're right. Can you delete that test?

src/wrappers/wrapper.js
```diff
@@ -114,6 +114,7 @@ export default class Wrapper implements BaseWrapper {
       .every(target => this.element.classList.contains(target))
 
     return !!(this.element && containsAllClasses)
+    // return !!(this.element && this.element.classList.contains(targetClass))
```
eddyerburgh: Can you remove this comment

src/wrappers/wrapper.js
lmiller1990: Won't it still throw an error if this.element is undefined (since `this.element.className.contains` will throw the "Cannot find .contains of null" type error, even if containsAllClasses is a function?

lmiller1990:
Updated!

- remove the old test 
- restore package.json to original state before checkout
- add comment regarding changing containsAllClasses to a function (I don't quite understand why).

Issue #117, opened 2017-10-23
Functional component context is undefined if not defined in mount() option argument


rssfrncs:
Version: 1.0.0-beta.2

If you mount a functional component without passing a context object in the options argument,
```
it('Example mocha test', () => {
  const wrapper = mount(FunctionalComp)
  // assertions
})
```

when mocha executes the test case, context is undefined ,
![image](https://user-images.githubusercontent.com/7808978/31879833-8d9063f2-b7d6-11e7-9595-df9e560a4c43.png)

Passing `{ context: {} }` to the options of mount() is a workaround but not ideal.

eddyerburgh:
Hi, this is in my list of todos. Would you be able to make a PR that implements this? If not, I'll be adding this feature at some point this week

rssfrncs:
@eddyerburgh sure i'll take a look this evening

rssfrncs:
@eddyerburgh Evening, I'm having a few issues running the test scripts.

run test:unit / test:unit:karma issue
![image](https://user-images.githubusercontent.com/7808978/31903152-30bde61a-b81f-11e7-812e-0343e14fd07a.png)

and run test issue

![image](https://user-images.githubusercontent.com/7808978/31903239-76c5df78-b81f-11e7-949a-a85fe684de28.png)

eddyerburgh:
Yes let's talk about it in a PR. There's a Vue version mismatch. Confusing because the circle tests are passing. Potentially clearing your npm cache will solve the issue

PR #126, opened 2017-10-24
Looking to fix issue #117

rssfrncs:
@eddyerburgh Evening Edd. 

I might be looking in the wrong place, but to me it seemed that functional components context was undefined due to the condition on line 17. What do you think?

I've tested in my own project and it is working as expected. I should look at adding a test case.

### DIFF
--- a/src/lib/create-instance.js
+++ b/src/lib/create-instance.js
@@ -14,20 +14,18 @@ import { compileTemplate } from './compile-template'
 export default function createConstructor (component: Component, options: Options): Component {
   const vue = options.localVue || Vue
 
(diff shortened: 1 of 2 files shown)


eddyerburgh:
This looks great. Like you said, please add a regression test, then I'll merge 🙂

rssfrncs:
@eddyerburgh test case added.

eddyerburgh:
Thanks 🙂

I just noticed you're living in Gloucestershire. I'm from Tetbury originally 🙌

rssfrncs:
@eddyerburgh no way, what a small world!

PR #129, opened 2017-10-25
Feat destroy wrapper method

vouill:
This PR adds

- `.destroy()` to `Wrapper`.

Heavily inspired from avoriaz :)


src/wrappers/error-wrapper.js
```diff
@@ -99,4 +99,8 @@ export default class ErrorWrapper implements BaseWrapper {
   update (): void {
     throwError(`find did not return ${this.selector}, cannot call update() on empty Wrapper`)
   }
+
+  destroy (): void {
+    throwError(`find did not return ${this.selector}, cannot call update() on empty Wrapper`)
```
eddyerburgh: should be `... cannot call destroy()`

test/unit/specs/wrappers/error-wrapper.spec.js
```diff
@@ -154,4 +154,11 @@ describe('ErrorWrapper', () => {
     const error = new ErrorWrapper(selector)
     expect(() => error.update()).to.throw().with.property('message', message)
   })
+
+  it('destroy throws error when called', () => {
+    const selector = 'div'
+    const message = `[vue-test-utils]: find did not return ${selector}, cannot call update() on empty Wrapper`
```
eddyerburgh: should be `cannot call destroy()`

vouill:
@eddyerburgh PR updated !

vouill:
Alright, thanks for walking me through this ! PR updated !

Issue #139, opened 2017-10-30
Support for classes(), etc


tim-hutchinson:
This came up in #118. We've found it easier to work with tests that use the assertion library's comparison functions, rather than the comparison functions like `hasClass`, `hasProp`, etc. 

I'd like to propose adding helpers that are more like `html()` and `text()` that simply return a consumable form of the element feature requested. IE, `classes()` would return an array of class names on the element. I've got an implementation of `classes` and can submit a PR for it and others that make sense, but wanted to see if it made sense before going further.

eddyerburgh:
I agree. Value assertions are more useful than boolean assertions.

We could replace hasClass, hasAttribute, and hasProp with `classes`, `attributes`, and `props`. 

We've got a good opportunity to remove these before the official release.

The one method I think is useful is hasStyle, because it allows you to check `hasStyle('color', 'white')` instead of returning `{color: 'rgb(255, 255, 255)}`

What do people think?

lmiller1990:
I haven't had much occasion to use `attributes` or `props`, but I'd be for including them - `props` seems like it might be handy. `classes` is solved in #118 , and should probably be renamed `classes` in light of the updated functionality.

I agree that `hasStyle` should stay, I have used  it quite a lot (changing button colors, opacities, and the such). I don't see a `styles` helper being very useful.

vouill:
I prefer personnaly prefer the notation `getClasses()` but other than that i like it too.

eddyerburgh:
I also prefer the notation `getClasses`.

What do other people think?

tim-hutchinson:
The `getX` syntax seems different from the conventions used in this library already (`html`, `text`). 

Alternatively, it could go fully the other way and implement the functions as getters, so they would expose a property-like API `expect(wrapper.classes).to.eql(['foo', 'bar'])`

eddyerburgh:
I've changed my mind, as you said we already have html and text methods.

eddyerburgh:
This has been released in 1.0.0-beta.6

Happy to continue discussing API, but thinking about it I think:

```js
wrapper.props().someProp
```

Is more elegant that

```js
wrapper.getProps().someProp
```

Plus we can extend the `props` method by adding an optional param:

```js
wrapper.props('someProp')
```

PR #141, opened 2017-11-01
Support for classes inspection

tim-hutchinson:
This PR has an implementation for #139 . I based the structure on the existing `html` and `text` methods that only work on a single instance, not an array.

Classes has a few tricky edge cases with CSS Modules. I don't use them in my own projects, so I don't have expertise on the right expected behavior, and couldn't find an official inverse function for them. With a CSS Module class that composes another class, both classes end up on the output. Right now, I'm returning both the extension and base classes back, but I don't know if this is the expected behavior. I'm also assuming that the base will always be defined in `$style`, but I haven't experience more exotic examples of CSS Modules (ie extending globals, etc).


eddyerburgh:
Thanks @tim-hutchinson . I'd like to wait for some more feedback before we merge, but let's keep this open, and I'll merge once we've decided.

eddyerburgh:
Would you be able to add `attributes` and `props` to this PR?

tim-hutchinson:
Sure, I'll get them added.

I noticed before there are tests for `WrapperArray` in both `test/unit/specs/mount/WrapperArray` and `test/unit/specs/wrappers/wrapper-array.spec.js`. Do you prefer tests in both locations or just one?

eddyerburgh:
Prefer tests in both locations

tim-hutchinson:
One question on the spec for `props`. If there are no props, should it return `undefined` or `{ }`? The component itself actually has it undefined, but an empty object/array is the behavior on the other inspection functions.

eddyerburgh:
@tim-hutchinson can you fix the merge conflict?

eddyerburgh:
Thanks @tim-hutchinson 😀

I'll add deprecation notices to hasClass, hasProp, and hasAttribute, and release later this week

PR #154, opened 2017-11-05
Propogate thrown errors by configuring error handler and attaching in…

### LINKED ISSUE #147: Expected Behavior of Error Propogation
Hey sorry for the multiple issues but saw this while driving out a test.

This may be more of a discussion, but what is the intended behavior of error propogation in vue-test-utils?  I created an example below but what i'm seeing in production is a mounted function in throwing an error due to some bad assumptions and that causes a white screen.  In my test when i reproduce by throwing an error in the mounted function I see the error in the console but I am not able to access it in vue-test-utils, i've tried numerous 'mocha' like ways to get the error to surface but was unable to.

Any thoughts on how to get access to thrown errors?

Here is a repo with reproduction.
https://github.com/Austio/vue-test-utils-mocha-webpack-example/pull/1/files

Austio:
… mount function resolves #147 

### High level overview

Creates an errorHandler function which we assign to Vue's config.errorHandler.  This errorHandler will be called with the error, the vm and additional info from vue.  The error handler handles
 - thrown Error objects `throw(new Error('error'))`
 - thrown string object `throw('error')`
 - appending `info` to the error message if it is provided

### Background/Justification
Currently vue-test-utils silently swallows errors that are thrown in components.  This can lead to false green test suits.  Proof of issue is in this repo pull request https://github.com/Austio/vue-test-utils-mocha-webpack-example/pull/1

### DIFF
--- a/src/lib/error-handler.js
+++ b/src/lib/error-handler.js
@@ -1,28 +1,19 @@
-function errorMessage(msg, info) {
+function errorMessage (msg, info) {
   if (info) {
     return `${msg} : additional info ${info}`
   }
 
-  return msg;
+  return msg
 }
 
-function setVueErrorHandler(vue) {
-  vue.config.errorHandler = errorHandler
-}
-
-function errorHandler(err, _vm, info) {
+export default function errorHandler (err, _vm, info) {
   if ((typeof err === 'object') && err.message) {
     if (info) {
-      err.message = errorMessage(err.message, info);
+      err.message = errorMessage(err.message, info)
     }
 
-    throw err;
+    throw err
   }
 
-  throw new Error(errorMessage(err, info));
-}
-
-export {
-  errorHandler,
-  setVueErrorHandler
+  throw new Error(errorMessage(err, info))
 }
--- a/src/lib/error-handler.js
+++ b/src/lib/error-handler.js
@@ -1,0 +1,28 @@
+function errorMessage(msg, info) {
+  if (info) {
+    return `${msg} : additional info ${info}`
+  }
+
+  return msg;
+}
+
+function setVueErrorHandler(vue) {
+  vue.config.errorHandler = errorHandler
+}
+
+function errorHandler(err, _vm, info) {
+  if ((typeof err === 'object') && err.message) {
+    if (info) {
+      err.message = errorMessage(err.message, info);
+    }
+
+    throw err;
+  }
+
+  throw new Error(errorMessage(err, info));
+}
+
+export {
+  errorHandler,
+  setVueErrorHandler
+}
--- a/src/mount.js
+++ b/src/mount.js
@@ -7,10 +7,10 @@ import createInstance from './lib/create-instance'
 import cloneDeep from 'lodash/cloneDeep'
 import createElement from './lib/create-element'
 import './lib/matches-polyfill'
-import { setVueErrorHandler } from './lib/error-handler'
+import errorHandler from './lib/error-handler'
 
 Vue.config.productionTip = false
-setVueErrorHandler(Vue)
+Vue.config.errorHandler = errorHandler
 
 export default function mount (component: Component, options: Options = {}): VueWrapper {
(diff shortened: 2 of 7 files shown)


test/unit/specs/lib/error-handler.spec.js
```diff
@@ -0,0 +1,38 @@
+import { errorHandler } from '../../../../src/lib/error-handler'
+
+describe('errorHandler', () => {
+  const errorString = 'errorString'
+  const info = 'additional info provided by vue'
+
+  describe('with Error thrown', () => {
```
eddyerburgh: Can you remove the nested describe blocks? Everywhere else the describe blocks aren't nested

src/lib/error-handler.js
```diff
@@ -0,0 +1,28 @@
...
+      err.message = errorMessage(err.message, info);
+    }
+
+    throw err;
+  }
+
+  throw new Error(errorMessage(err, info));
+}
+
+export {
```
eddyerburgh: Can you export the functions rather than an object that contains the functions

src/mount.js
```diff
@@ -7,8 +7,10 @@ import createInstance from './lib/create-instance'
 import cloneDeep from 'lodash/cloneDeep'
 import createElement from './lib/create-element'
 import './lib/matches-polyfill'
+import { setVueErrorHandler } from './lib/error-handler'
 
 Vue.config.productionTip = false
+setVueErrorHandler(Vue)
```
eddyerburgh: I think we should export errorHandler and call Vue.config.errorHandler = errorHandler, instead of calling a function. Especially since we set productionTip false just above the function

Austio: Hey, on this one we are going to need to do this config.errorHandler in each place that we have an entry into vue-test-utils (shallow, mount and create local vue)  will update with individual imports

Side thought for now or future, it may make sense to have a custom version of Vue that we import from so that we can isolate any global things like this to a single point.  

```import Vue from '.lib/base-vue'```

src/lib/error-handler.js
Austio: 👍

test/unit/specs/lib/error-handler.spec.js
Austio: 👍

Austio:
@eddyerburgh should be good

I think something to consider is having a TestUtilsVue and import Vue from there when we are implementing things like mount, shallow, etc.  It would allow us to centralize things that need to be set on each Vue instance globally (like the error handler)

eddyerburgh:
When you import/ require in node it caches the instance, so every import to vue is the same vue constructor as every other require.

// some-file.js 
import Vue from 'vue'
Vue.config.errorHandler = true

// some-other-file.js 
import Vue from 'vue'
console.log(Vue.config.errorHandler) // true

So if I understood correctly, there's no need to centralize a Vue singleton, because node requires are always singletons (unless you explicitly delete the cache)

eddyerburgh:
Ah I just read your code, I see what you mean. We could add it in the createInstance—https://github.com/vuejs/vue-test-utils/blob/dev/src/lib/create-instance.js#L18. This is the same variable for localVue and for a regular Vue

eddyerburgh:
I'm happy to merge as it is

Austio:
Cool let's go ahead and merge and if we need the central in future can add then.

Austio:
@eddyerburgh hey hey, wanted to ping you on this one, good to merge when you are ☮️

eddyerburgh:
The tests are failing on my machine for some reason. The custom error is thrown, but it doesn't propagate 🤔

I'm going to skip them for the moment and have a look later, but do you have any idea why that might be?

Austio:
Hey nothing off top of head.  Any specific reproduction steps?   Should be able to gander here in a few hours.

Austio:
@eddyerburgh patched #170

PR #173, opened 2017-11-12
Cannot pass content to slots in functional components

znck:
```js
const wrapper = mount(FunctionalComponent, { slots: { default: '<p>some content</p>' } })

wrapper.contains('p') // false
```


eddyerburgh:
Thanks @znck, I've added the test and fix in my local branch. I'll push tomorrow. I didn't want to merge into develop with broken tests.

znck:
It's okay. I wanted to present the issue with an example. 

👍 Thanks for quick fix.

PR #181, opened 2017-11-17
Emit api

lmiller1990:
Add an alternative syntax for `wrapper.emitted()`.

Also updated docs to include new syntax (ja and en).

### DIFF
--- a/docs/en/api/wrapper/emitted.md
+++ b/docs/en/api/wrapper/emitted.md
@@ -31,3 +31,16 @@ expect(wrapper.emitted().foo.length).toBe(2)
 // assert event payload
 expect(wrapper.emitted().foo[1]).toEqual([123])
 ```
+
+You can also write the as follows:
+
+```js
+// assert event has been emitted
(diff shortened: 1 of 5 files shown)


lmiller1990:
In response to #181 

@eddyerburgh , I'm sure I need to add something for Typescript somewhere for this. 

It turned out adding the different syntax was trivial and I was overthinking it. I'm sure you have some feedback or change requests.

Also perhaps @kazupon will check the Japanese translation.

lmiller1990:
Thanks @kazupon .

Also for some reason `createLocalVue Router should work properly with local Vue` was failing locally, but it passes on the CI. Eh.

eddyerburgh:
Are you able to add types? They're in the types/index.d.ts https://github.com/vuejs/vue-test-utils/blob/dev/types/index.d.ts

I can help you if you're stuck 😄

lmiller1990:
I tried to add the types. I think it is just one type, the new optional event string. I haven't used typescript before, is it a big game changer? Should I be? 

Let me know if I should change anything else, @eddyerburgh .

PR #182, opened 2017-11-17
Fix case vue.nextTick to Vue.nextTick

ChangJoo-Park:
It seems to be uppercase https://vuejs.org/v2/api/#Vue-nextTick


ChangJoo-Park:
Thanks kazupon :)

PR #196, opened 2017-11-22
Update testing-SFCs-with-jest.md

Alex-Sokolov:
Rename `jest-vue` to `vue-jest` because above we installing `npm install --save-dev vue-jest`


eddyerburgh:
Thanks 🙂

Issue #132, opened 2017-10-27
Mounted functional component has undefined context.children


rssfrncs:
Version: beta 3

Mounting a functional component passing options.context.children, results in context.children to be undefined in the components render function.

![image](https://user-images.githubusercontent.com/7808978/32097204-9e0eb56c-bb00-11e7-8dd1-9a7fa5629aa9.png)

Console output

![image](https://user-images.githubusercontent.com/7808978/32097226-b40b19d2-bb00-11e7-9eb1-e86d73681f53.png)

What is curious is that data.children contains the expected array.

rssfrncs:
Taking a look at `lib/create-instance.js` , if options.context.children is passed as a [third argument](https://github.com/vuejs/vue-test-utils/blob/3c563d12735b8eab7529d32494f0955e0c51a294/src/lib/create-instance.js#L27) then context.children is defined, i'm not exactly sure why this works or what `h` in the functional components render context refers to?

eddyerburgh:
h is an alias for the [createElement](https://vuejs.org/v2/guide/render-function.html#createElement-Arguments) function

Can you create a failing test?

rssfrncs:
@eddyerburgh Ah nice. Yeah will do when i'm home from work.

rssfrncs:
@eddyerburgh sorry for the delay, i've added a test case to this [branch](https://github.com/rssfrncs/vue-test-utils/tree/context-children-undefined) in this [file](https://github.com/rssfrncs/vue-test-utils/blob/a89c42dee7641293d46179b13bbef7512b58e119/test/unit/specs/mount/options/context.spec.js#L69).

eddyerburgh:
Fixed in 5fc1c57bf5d14d0425b748f9a3f42d6eb59791e2

rssfrncs:
@eddyerburgh I noticed at work this morning, that whilst _text_ children are rendered correctly, components like `h => <div>hello</div>` are not. This is partly my fault, as the [original test case](https://github.com/rssfrncs/vue-test-utils/blob/a89c42dee7641293d46179b13bbef7512b58e119/test/unit/specs/mount/options/context.spec.js#L69) I raised only included a text child.

[...]

PR #198, opened 2017-11-22
#132 additional work

rssfrncs:
Adds to #132, now renders children depending on createElement correctly. Updated tests for text and added one for createElement child.


eddyerburgh:
Thanks 🙂

Issue #88, opened 2017-10-12
Add config object


eddyerburgh:
I think we should add a config object.

We could use it to store mount options, and other options at a later date.

The main use I see right now is to set transition and transition-group Components to be stubbed by default.

This means the global default config would look like this:

```js
console.log(vueTestUtils.config) // {
  stubs: {
    'transition': TransitionStub,
    'transition-group': TransitionGroupStub
  }
}
```

The options in config would be overwritten by mount options passed to Wrapper.

Having a config would save adding large mount options in tests.

What are peoples thoughts?

petridw:
Is there currently a way to have stubs apply to child components of the originally mounted component?

e.g. `mount(ComponentWithChildren, { stubs: { transition: TransitionStub } }` where the transition stub would also be applied recursively to any child components of `ComponentWithChildren`.

Seems having this global config option would be one option for enabling this behavior.

eddyerburgh:
There isn't currently a way for doing this
 We could add it ☺

eddyerburgh:
Closed with 7962b88f7535ce8f278b35b0ead1dcb6c2c0981c

PR #199, opened 2017-11-23
Add missing types to type definition

elevatebart:
When using typescript on a vue site that has animation, it becomes quite vital to be able to check that transitions are called. 
In order to do just that in TS too, I added the missing types.
This way I can spy on render and check that animations are run.

This is connected with the following issue:
https://github.com/vuejs/vue-test-utils/issues/88


eddyerburgh:
Thanks, I'll release in beta.7 later this week 🙂

Issue #208, opened 2017-11-27
Triggering events in IE


ccamarat:
I'm running unit tests in browser using Karma. When I use `wrapper.trigger('click')`, I get the error message `Object doesn't support this action`. This was [resolved in Avoriaz](https://github.com/eddyerburgh/avoriaz/issues/81) by [providing a fallback for IE](https://github.com/eddyerburgh/avoriaz/commit/c3457770f88cf414b8e82dff1c10e4648b99ae55) in Wrapper.js.

In the docs I notice you recommend against running tests in browsers, but, well, we are. Are you willing to accept a PR to fix this, or do you consider this kind of problem beyond the scope of this project?

FWIW it's possible to work around this issue by providing an Event constructor before running tests, so we'll be fine either way:
```JavaScript
let oEvent;
beforeEach(() => {
    if (typeof (window.Event) !== 'function') {
        oEvent = window.Event
        window.Event = function (arg0) {
            const eventObject = document.createEvent('Event')
            eventObject.initEvent(arg0, true, true)
            return eventObject
        }
    }
})

afterEach(() => {
    if (oEvent) {
        window.Event = oEvent
        oEvent = null
    }
})
```

eddyerburgh:
Hi, thanks. Of course we're willing to accept a PR! It would be greatly appreciated.

You can reuse the code from avoriaz I believe. If you could test manually, that would be great!

ccamarat:
Done. Thanks!

PR #209, opened 2017-11-27
Add workaround for IE in trigger

ccamarat:
Addresses #208. Reran tests locally and they all passed (there are some linter failures but not in the code changed here). I built and manually verified this fix allows events to trigger in IE.


eddyerburgh:
Thanks 😀

PR #225, opened 2017-12-04
Add VueWrapper.exists()

38elements:
That `Wrapper.exists()` returns `false` after a `wrapper` is destoryed is more appropriate.

### DIFF
--- a/src/wrappers/vue-wrapper.js
+++ b/src/wrappers/vue-wrapper.js
@@ -29,8 +29,4 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
 
     logEvents(vm, this._emitted, this._emittedByOrder)
   }
-
-  exists (): boolean {
-    return !!this.vm && !this.vm._isDestroyed
-  }
 }
--- a/src/wrappers/vue-wrapper.js
+++ b/src/wrappers/vue-wrapper.js
@@ -29,4 +29,8 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
 
     logEvents(vm, this._emitted, this._emittedByOrder)
   }
+
+  exists (): boolean {
+    return !!this.vm && !this.vm._isDestroyed
+  }
 }
--- a/src/wrappers/wrapper-array.js
+++ b/src/wrappers/wrapper-array.js
@@ -39,7 +39,7 @@ export default class WrapperArray implements BaseWrapper {
   }
 
   exists (): boolean {
-    return this.wrappers.some(wrapper => wrapper.exists())
+    return this.length > 0 && this.wrappers.every(wrapper => wrapper.exists())
   }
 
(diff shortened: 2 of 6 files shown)


eddyerburgh:
Why did you close the PR? It looks good 😀

38elements:
Thank you for reply.
I think that this changes  not only `wrapper.exists()` but also  `wrapperArray.exists()`.
I will change  `wrapperArray.exists()`.

38elements:
`VueWrapper` has `vm` property.
`Wrapper` does not have `vm` property.
This does not work.

src/wrappers/vue-wrapper.js
```diff
@@ -29,4 +29,8 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
 
     logEvents(vm, this._emitted, this._emittedByOrder)
   }
+
+  exists (): boolean {
+    return !!this.vm && !this.vm._isDestroyed
```
eddyerburgh: Can you move this method to wrapper.js and return true if this.isVueComponent is false. A node Wrapper will always exist. If find couldn't find anything, it would return an ErrorWrapper instead

src/wrappers/wrapper-array.js
```diff
@@ -37,8 +37,9 @@ export default class WrapperArray implements BaseWrapper {
 
     return this.wrappers.every(wrapper => wrapper.contains(selector))
   }
+
   exists (): boolean {
-    return this.wrappers.length > 0
+    return this.wrappers.filter(wrapper => wrapper.exists()).length > 0
```
eddyerburgh: You can use .every instead—
```
return this.wrappers.every(wrapper => wrapper.exists())
```

38elements:
```
[].every(a => a)
// => true
```

38elements: I think that it is appropriate to return `true` if there is at least one element matching the search condition.

38elements: IMHO, I think that it is better to use `some`.

```
return this.wrappers.length > 0 && this.wrappers.some(wrapper => wrapper.exists())
```

38elements: Thank you for reviewing.
I think the following code using `some` is equivalent.

```
return this.wrappers.some(wrapper => wrapper.exists())
```

eddyerburgh: some would return true if only one wrapper.exists() returns true, every returns true only if every Wrapper exists, which is the behavior of other methods like hasStyle

38elements: Thank you for reply.
I will use `every`.
I will change this.

38elements: I changed this.

Issue #229, opened 2017-12-05
Passing text to default slots 


WebDevLuke:
Hey all, so I have a very simple component with a default slot, like so:

```
<template>
	<button>
		<slot></slot>
	</button>
</template>

<script>
export default {
	name: 'app-button'
}
</script>
```

And would use like so:

```
<app-button>My Button</app-button>
```

How can I test that the `My Button` text passed to the component's default slot is rendering correctly?

I have the below unit test currently, which is working only because I've wrapped the `My Button` testing text in a span element, as otherwise I'm getting a `Component template requires a root element, rather than just text.` error.

```
import { mount } from 'vue-test-utils'
import appButton from '@/components/app-button/app-button'

describe('app-button.vue', () => {
	it('values passed to slots should render correctly', () => {
		const wrapper = mount(appButton, {
			slots: {
				default: '<span>My Button</span>'
			}
		})
		expect(wrapper.find('span').text()).toBe('My Button')
	})
})
```

How would I modify this so I can simply just test that `<app-button>My Button</app-button>` works, without having to wrap `My Button` string in a span? I apologize if this seems a stupid question, but I've read the docs and can't quite find a way to cover this very simple case. Thanks.

eddyerburgh:
There actually isn't a way to do this right now. Perhaps we should think about adding it to the slots API. What do people think?

WebDevLuke:
In that case, I look forward to seeing this implemented. Passing text to slots is quite a common use case, especially with simple, dumb components. 😃

eddyerburgh:
This has been added and will be released in the next beta version 🙂

eddyerburgh:
Released in 1.0.0-beta.7 🙌

eddyerburgh:
we need to add this to docs

38elements:
I will add document about this within 3 days.

PR #233, opened 2017-12-06
Passing text to slots

38elements:
This is related to #229.

This support only text as below.

```js
const wrapper = mount(ComponentWithSlots, { slots: { default: 'foobar' }})
```

This does not work for the text like one as below.

```js
const wrapper = mount(ComponentWithSlots, { slots: { default: 'foo<span>bar</span>' }})
```

```js
const wrapper = mount(FooComponent, { slots: { default: 'foo {{ bar }}' }})
```

### DIFF
--- a/test/unit/specs/mount/options/slots.spec.js
+++ b/test/unit/specs/mount/options/slots.spec.js
@@ -1,5 +1,4 @@
 import { compileToFunctions } from 'vue-template-compiler'
-import Vue from 'vue'
 import mount from '~src/mount'
 import Component from '~resources/components/component.vue'
 import ComponentWithSlots from '~resources/components/component-with-slots.vue'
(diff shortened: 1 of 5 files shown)


eddyerburgh:
Great, thanks 🙂

The tests failed runnning Vue 2.0.8. It must not add v as an alias. Ideally we'll have a workaround that works in Vue 2.0.8, if you can't find one we can add an error that text slot values aren't supported in Vue 2.0.x

38elements:
I look for ways to solve at `Vue 2.0.8`.

38elements:
I could not find a way to solve it.
I added a warning.

There is another problem.
The following code does not work.

```js
const wrapper = mount(FooComponent, { slots: { default: 'foo {{ bar }}' }})
```

test/unit/specs/mount/options/slots.spec.js
```diff
@@ -25,6 +26,18 @@ describe('mount.slots', () => {
     expect(wrapper.contains('span')).to.equal(true)
   })
 
+  it('mounts component with default slot if passed string in slot object', () => {
+    const vm = new Vue()
+    if ('_v' in vm) {
```
eddyerburgh: Could you find which Vue version supports _v and then import { vueVersion } from ~resources/test-utils. 

So instead of this:
```js
if ('_v' in vm) {
```
use this:
```js
if (vueVersion > 2.2) {
```
That way it makes more sense from other developers.

38elements: https://github.com/vuejs/vue/blob/68abdc3c3c86e01089c32b434e662dc15a3a9eda/src/core/instance/render.js

[Vue.js 2.1.5](https://github.com/vuejs/vue/releases/tag/v2.1.5) supports  `vue._v()`.

38elements: https://github.com/vuejs/vue/commit/79e1058799dcde0b7e341ef44b590ecbf1e3a722#diff-5b337525bf69cfea5590728194e8d92f

[Vue.js 2.1.5](https://github.com/vuejs/vue/releases/tag/v2.1.5) supports  `vue._v()`.

I will change this.

38elements: I changed this.

Issue #215, opened 2017-11-29
Events emitted on mount not captured


fergaldoyle:
Events emitted immediately in the `mounted` hook are not captured by `wrapper.emitted()`. With something like the below
- event `something` is not captured
- event `something-else` is not captured on first emit in the immediate watcher, but is captured on subsequent watcher callbacks
- event `toggle` is captured if the `toggle` method is called after mount, e.g. on user click

```javascript
{
  mounted() {
    this.$emit('something');
    
    this.$watch('foo', (v) => {
      this.$emit('something-else');
    }, { 
      immediate: true 
    });
    
  },
  methods: {
    toggle() {
      this.$emit('toggle');
    }
  }
}
```

wtho:
This happens, because the [wrapper attaches to the component](https://github.com/vuejs/vue-test-utils/blob/8904ddd407eade79611f6ffaaa79ad544d0a7e0f/src/mount.js#L32) **after** the component was [created and mounted](https://github.com/vuejs/vue-test-utils/blob/8904ddd407eade79611f6ffaaa79ad544d0a7e0f/src/mount.js#L20).

I do not know if there is a way to change this, but I have the feeling if we do not attach the wrapper earlier there might be more edge cases where the wrapper is not ready when the component gets created.

One possibility I see is to install the wrapper earlier is via `beforeCreate`-mixin. It would be called by the Vue-framework after the vm-object is instiated but before the Vue-component is created and mounted, and so we could do the setup then already. We might have to split the actions done in `vue-wrapper.js` in several hooks.

But maybe there are other ways as well.

eddyerburgh:
I was thinking we could add a listener object before mount and pass it to VueWrapper when it's created and merge that listener object with the current listener object.

wtho:
Ok, so what worked for me was this (using the `logEvents` method in `src/lib/log-events.js`):
```js
localVue.mixin({
  beforeCreate: function() {
    this.$_emitted = {}
    this.$_emittedByOrder = []
    logEvents(this, this.$_emitted, this.$_emittedByOrder)
  }
})
```
It does catch all events from `beforeCreate` on in a variable of the component (in this case `$._emitted`). In the constructor we could then copy these self-tracked events to the wrapper. The mixin could be added in every `localVue` instance when creating it.

This approach is quite naive and the first listener should be removed after the wrapper is created. It also seems a bit hacky and not too clean to use a field on the component.

[...]

PR #243, opened 2017-12-09
Fix event capturing in beforeCreate lifecycle hook

wtho:
The event capturing was installed in the wrapper constructor, after the component (vm) was created and mounted, which is too late to catch events emitted in lifecycle hooks like `beforeCreate`.

Two additional tests were written, in `emitted.spec.js` and `emittedByOrder.spec.js`, and an existing one was adjusted. Notice I had to treat Vue v2.0.8 differently to pass the compatibility tests (in `emittedByOrder.spec.js`), as that version emits lifecycle hooks events, whereas the others do not.

Implementation:
* A simple event captor object is created when `mount` is called before the `vm` is created.
* The captor gets passed to the `createInstance` method, where it will get installed as a mixin on the used Vue instance. The installation uses the same method as the wrapper used before. The vm creation continues as usual
* When the wrapper gets created in the `mount` method, the captor gets passed as well. Instead of installing the capturing on its own, the wrapper will use the event captor object.
* Flow type definitions were added for the EventCaptor, as it is passed in two methods.

This solves #215.

### DIFF
--- a/src/lib/create-instance.js
+++ b/src/lib/create-instance.js
@@ -13,13 +13,22 @@ import extractOptions from '../options/extract-options'
 import deleteMountingOptions from '../options/delete-mounting-options'
 import createFunctionalComponent from './create-functional-component'
 
+import { logEvents } from '../lib/log-events'
+
 export default function createConstructor (
   component: Component,
-  options: Options
+  options: Options,
+  eventCaptor: EventCaptor
 ): Component {
   const mountingOptions = extractOptions(options)
 
   const vue = mountingOptions.localVue || createLocalVue()
+
+  vue.mixin({
+    beforeCreate: function () {
+      logEvents(this, eventCaptor.emitted, eventCaptor.emittedByOrder)
+    }
+  })
 
   if (mountingOptions.mocks) {
     addMocks(mountingOptions.mocks, vue)
--- a/src/lib/create-instance.js
+++ b/src/lib/create-instance.js
@@ -5,6 +5,7 @@ import addMocks from './add-mocks'
 import addAttrs from './add-attrs'
 import addListeners from './add-listeners'
 import addProvide from './add-provide'
+import { addEventLogger } from './log-events'
 import { stubComponents } from './stub-components'
 import { throwError } from './util'
 import { compileTemplate } from './compile-template'
@@ -13,22 +14,13 @@ import extractOptions from '../options/extract-options'
 import deleteMountingOptions from '../options/delete-mounting-options'
 import createFunctionalComponent from './create-functional-component'
 
-import { logEvents } from '../lib/log-events'
-
 export default function createConstructor (
   component: Component,
-  options: Options,
-  eventCaptor: EventCaptor
+  options: Options
 ): Component {
   const mountingOptions = extractOptions(options)
 
   const vue = mountingOptions.localVue || createLocalVue()
-
-  vue.mixin({
-    beforeCreate: function () {
-      logEvents(this, eventCaptor.emitted, eventCaptor.emittedByOrder)
-    }
-  })
 
   if (mountingOptions.mocks) {
     addMocks(mountingOptions.mocks, vue)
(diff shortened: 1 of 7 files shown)


src/wrappers/vue-wrapper.js
```diff
@@ -24,9 +23,9 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
     }))
     this.vm = vm
     this.isVueComponent = true
-    this._emitted = Object.create(null)
-    this._emittedByOrder = []
+    this._emitted = options.eventCaptor.emitted
+    this._emittedByOrder = options.eventCaptor.emittedByOrder
 
-    logEvents(vm, this._emitted, this._emittedByOrder)
+    // logEvents(vm, this._emitted, this._emittedByOrder)
```
eddyerburgh: We can delete this line

src/lib/create-instance.js
```diff
@@ -13,14 +13,23 @@ import extractOptions from '../options/extract-options'
...
+  options: Options,
+  eventCaptor: EventCaptor
 ): Component {
   const mountingOptions = extractOptions(options)
 
   const vue = mountingOptions.localVue || createLocalVue()
 
+  vue.mixin({
+    beforeCreate: function () {
+      logEvents(this, eventCaptor.emitted, eventCaptor.emittedByOrder)
```
eddyerburgh: I actually think we should use `this.__emitted` and `this.__emittedBy`, because otherwise (I think) all events logged by any of the child components will be added to the eventCaptor. Using `this.__emitted` means we can support child component emitted, which won't work at the moment.

src/wrappers/vue-wrapper.js
```diff
@@ -24,9 +23,9 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
     }))
     this.vm = vm
     this.isVueComponent = true
-    this._emitted = Object.create(null)
-    this._emittedByOrder = []
+    this._emitted = options.eventCaptor.emitted
+    this._emittedByOrder = options.eventCaptor.emittedByOrder
```
eddyerburgh: We should add this._emitted (maybe this.__emitted to be safe) in the mixin, then every component will have them added. Could you write a test to make sure this is true? Every child component should have a clean emitted object that only refers to events emitted by itself

src/lib/create-instance.js
wtho: Yeah, I think you are right, if a `localVue` instance will be reused, the captor might be added several times. I thought it might be better to keep it non-intrusive, but I will do it using `this.__emitted` then.

wtho: > Using this.__emitted means we can support child component emitted, which won't work at the moment.

How would that work? Is there anything required to implement this as well? Right now this does not get captured by `wrapper.emitted()`:
```js
    const wrapper = mount({
      template: '<child-component></child-component>',
      components: {
        'child-component': {
          beforeCreate: function () {
            this.$emit('childEvent')
          }
        }
      }
    })
```

eddyerburgh: Sorry, that wasn't very clear. I meant that at the moment (in this PR) only root components will support emitted, child components won't, because when the VueWrapper is created with them, they won't have an eventCaptor to pass in. 

This should work:
```js
    const wrapper = mount({
      template: '<child-component></child-component>',
      components: {
        'child-component': {
          beforeCreate: function () {
            this.$emit('childEvent')
          }
        }
      }
    })
wrapper.find(childComponent).vm.emitted('childEvent') // returns array

Thinking about the underscore, I think it's fine to have a single underscore.

wtho:
Applied the changes. `this.__emitted` and `this.__emittedByOrder` is used to capture the events. These two fields will then be used by the wrapper, when it is created and they will not be removed.

I removed previously added type definitions in flow, that were not required anymore.

I created an `addEventLogger` method in `logEvents.js` to keep `createInstance.js` clean. It violates the `add-something.js`-pattern, as it does not use the default export, but I prefer it being with the `logEvents` method (that is nowhere else used at the moment than in `addEventLogger`).

I added one more test to check for side effects on localVue with several wrappers / components. The events emitted by children are not captured by the parent, but on the child component. I might add another test to show this.

eddyerburgh:
Thanks so much @wtho  🙂

PR #254, opened 2017-12-11
Wrong propsData definition

ThomasKlessen:
propsData has to be a object with key(/prop) and value. 
At least this is how it works for me.

My first commit so i hope to meet the requirements.
Otherwise let me know where to read and i will improve this PR.


eddyerburgh:
Yes, good catch! Thanks for the PR 🙂

PR #257, opened 2017-12-13
Add document for passing text to slots

38elements:
This is related to #229 .

### DIFF
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -66,15 +66,15 @@ expect(wrapper.find('div')).toBe(true)
 You can pass text to `slots`.  
 There are two limitations to this.
 
-This works at vue@2.2+.
+This works with Vue 2.2+.
 
-This works for the text as below.
+This works for the text below.
 
 ```js
 const wrapper = mount(ComponentWithSlots, { slots: { default: 'foobar' }})
 ```
 
-This does not work for the text as below.
+This does not work for the text below.
 
 ```js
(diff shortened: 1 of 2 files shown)


docs/en/api/options.md
```diff
@@ -54,12 +54,33 @@ const wrapper = shallow(Component, {
...
 ```
 
+#### Passing text
+
+You can pass text to `slots`.  
+There are two limitations to this.
+
+This works at vue@2.2+.
+
+This works for the text as below.
```
eddyerburgh: >This works for the text as below.

to 

>This works for the text below.

docs/en/api/options.md
```diff
@@ -54,12 +54,33 @@ const wrapper = shallow(Component, {
...
+
+```js
+const wrapper = mount(ComponentWithSlots, { slots: { default: 'foobar' }})
+```
+
+This does not work for the text as below.
```
eddyerburgh: This does not work for the text as below.

>

This does not work for the text below.

docs/en/api/options.md
```diff
@@ -54,12 +54,33 @@ const wrapper = shallow(Component, {
...
 })
 expect(wrapper.find('div')).toBe(true)
 ```
 
+#### Passing text
+
+You can pass text to `slots`.  
+There are two limitations to this.
+
+This works at vue@2.2+.
```
eddyerburgh: > This works at vue@2.2+

to

> This works with Vue 2.2+

38elements:
Thank you for reviewing.
I changed this.

Issue #263, opened 2017-12-14
attributes() method always return an Object{undefined: undefined}


tianyong90:
Today , I upgrade to beta.8, and I tried to using `attributes()` , but found that it always return `Object{undefined: undefined}`.

![image](https://user-images.githubusercontent.com/9278461/34002085-ab59c1e2-e12c-11e7-8831-0eabf989f8c5.png)

eddyerburgh:
This will be released in beta.9

PR #265, opened 2017-12-14
Fix wrapper.attributes() method

tianyong90:
#263

### DIFF
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -37,11 +37,12 @@ export default class Wrapper implements BaseWrapper {
    * Returns an Object containing all the attribute/value pairs on the element.
    */
   attributes (): { [name: string]: string } {
-    const attributes = [...this.element.attributes] // NameNodeMap is not iterable
+    const attributes = this.element.attributes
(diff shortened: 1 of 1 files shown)


tianyong90:
Sure, I am starting working on it. 😄

tianyong90:
@eddyerburgh Hi, after reviewing the tests, I found that these needed unit tests are already exists and there's no need to change them，none of them failed.
image

eddyerburgh:
Hi @tianyong90 . These are bad tests if they passed when the code was buggy. Would you like to investigate why they were passing when the code was returning undefined? If not, I can merge and investigate myself. Thanks for the PR 😀

tianyong90:
image

Aha, Finally found out that it seems that it's a bug in the `dist` but not in `src`, perhaps there's something wrong with the bounding tools or configuration. Just like shown in the picture above, the spread operator is transformed to `[].concat()`, this is the reason why unit tests all passed, but broken in my actual use.
To fix it, you can just merge the PR, or make some change with the building process.

eddyerburgh:
Great find, thanks for the PR 🙂

Issue #244, opened 2017-12-09
Passing text to slots in mounting options


38elements:
The `slots`  in `mounting options` does not work for the text like one as below.

```js
const wrapper = mount(ComponentWithSlots, { slots: { default: 'foo<span>bar</span>' }})
```

```js
const wrapper = mount(FooComponent, { slots: { default: 'foo {{ bar }}' }})
```

PR #274, opened 2017-12-17
Improve passing text to slots

38elements:
This is related to #244 .

The text works below.

```js
const wrapper1 = mount(ComponentWithSlots, { slots: { default: '1{{ foo }}2' )
const wrapper2 = mount(ComponentWithSlots, { slots: { default: '<p>1</p>{{ foo }}<p>2</p>' )
const wrapper3 = mount(ComponentWithSlots, { slots: { default: '<p>1</p>{{ foo }}' )
const wrapper4 = mount(ComponentWithSlots, { slots: { default: '123' )
const wrapper5 = mount(ComponentWithSlots, { slots: { default: '1<p>2</p>{{ foo }}3' )
const wrapper6 = mount(ComponentWithSlots, { slots: { default: '<p>1</p><p>2</p>' }})
const wrapper7 = mount(ComponentWithSlots, { slots: { default: '1<p>2</p>3' }})
```

### DIFF
--- a/src/lib/add-slots.js
+++ b/src/lib/add-slots.js
@@ -22,7 +22,7 @@ function addSlotToVm (vm: Component, slotName: string, slotValue: Component | st
     if (_slotValue[0] === '<' && _slotValue[_slotValue.length - 1] === '>' && document.body.childElementCount === 1) {
       elem = vm.$createElement(compileToFunctions(slotValue))
     } else {
-      const compiledResult = compileToFunctions(`<div>${slotValue}</div>`)
+      const compiledResult = compileToFunctions(`<div>${slotValue}{{ }}</div>`)
       const _staticRenderFns = vm._renderProxy.$options.staticRenderFns
       vm._renderProxy.$options.staticRenderFns = compiledResult.staticRenderFns
       elem = compiledResult.render.call(vm._renderProxy, vm.$createElement).children
--- a/src/lib/add-slots.js
+++ b/src/lib/add-slots.js
@@ -13,6 +13,9 @@ function addSlotToVm (vm: Component, slotName: string, slotValue: Component | st
     if (!compileToFunctions) {
       throwError('vueTemplateCompiler is undefined, you must pass components explicitly if vue-template-compiler is undefined')
     }
+    if (window.navigator.userAgent.match(/PhantomJS/i)) {
+      throwError('option.slots does not support PhantomJS. Please use Puppeteer')
+    }
     const domParser = new window.DOMParser()
     const document = domParser.parseFromString(slotValue, 'text/html')
     const _slotValue = slotValue.trim()
--- a/src/lib/add-slots.js
+++ b/src/lib/add-slots.js
@@ -1,6 +1,5 @@
 // @flow
 
-import Vue from 'vue'
 import { compileToFunctions } from 'vue-template-compiler'
 import { throwError } from './util'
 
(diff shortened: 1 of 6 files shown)


src/lib/add-slots.js
```diff
@@ -10,19 +9,21 @@ function isValidSlot (slot: any): boolean {
 
 function addSlotToVm (vm: Component, slotName: string, slotValue: Component | string | Array<Component> | Array<string>): void {
   let elem
-  const vueVersion = Number(`${Vue.version.split('.')[0]}.${Vue.version.split('.')[1]}`)
   if (typeof slotValue === 'string') {
     if (!compileToFunctions) {
       throwError('vueTemplateCompiler is undefined, you must pass components explicitly if vue-template-compiler is undefined')
     }
-    if (slotValue.trim()[0] === '<') {
+    const domParser = new window.DOMParser()
```
eddyerburgh: DOMParser isn't supported in phantomJs, so we need an alternative that will work in phantom

38elements: Thank you for reviewing.
Is there choice not to support PhantomJS ?
I think it is better to use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).

eddyerburgh: A lot of users still use phantom. We could throw an error, or a warning that phantom doesn't support markup in slots, and handle it gracefully

38elements: Thank you for reply.
I will look for other ways.

38elements:
I changed this.
This throws error when the UserAgent is PhantomJS.

test/unit/specs/mount/options/slots.spec.js
```diff
@@ -4,6 +4,18 @@ import Component from '~resources/components/component.vue'
...
 
 describe('mount.slots', () => {
+  let _window
+
+  beforeEach(() => {
+    _window = window
+  })
+
+  afterEach(() => {
+    /* eslint-disable */
```
eddyerburgh: Unless I'm mistakenm you can just add // eslint-disable-line to the line, instead of 2 comments. Even better, you can add `//eslint-disable-line name-of-rule` to only disable eslint for the rule which is failing

src/lib/add-slots.js
eddyerburgh: Can you change the error message to—`option.slots does not support strings PhantomJS. Please use Puppeteer, or pass a component`

test/unit/specs/mount/options/slots.spec.js
38elements: Thank you for pointing out.
I fixed it.

38elements:
I am sorry.
There were a few mistakes in document.

38elements:
I am sorry.
There is a bug.

```vue
<template>
  <div class="container" @click='change'>
    <header>
      <slot name="header"></slot>
    </header>
    <main>
      <slot></slot>
    </main>
    <footer>
      <slot name="footer"></slot>
    </footer>
  </div>
</template>

<script>
  export default {
    name: 'component-with-slots',
    data () {
      return {
        'foo': 'bar'
      }
    },
    methods: {
      change () {
        this.foo = 'baz'
      }
    }
  }
</script>
```


```js
    const wrapper5 = mount(ComponentWithSlots, { slots: { default: '1{{ foo }}2' }})
    wrapper5.find('.container').trigger('click')
    expect(wrapper5.find('main').html()).to.equal('<main>1baz2</main>')
   // => <main>undefined</main>
```

38elements:
I think it is resolved.

src/lib/add-slots.js
38elements: I am sorry.
I did not notice your comment.
I fixed the error message.

src/lib/add-slots.js
```diff
@@ -10,27 +9,40 @@ function isValidSlot (slot: any): boolean {
...
 function addSlotToVm (vm: Component, slotName: string, slotValue: Component | string | Array<Component> | Array<string>): void {
   let elem
-  const vueVersion = Number(`${Vue.version.split('.')[0]}.${Vue.version.split('.')[1]}`)
   if (typeof slotValue === 'string') {
     if (!compileToFunctions) {
       throwError('vueTemplateCompiler is undefined, you must pass components explicitly if vue-template-compiler is undefined')
     }
-    if (slotValue.trim()[0] === '<') {
+    if (window.navigator.userAgent.match(/PhantomJS/i)) {
+      throwError('option.slots does not support strings PhantomJS. Please use Puppeteer, or pass a component')
```
eddyerburgh: > option.slots does not support strings PhantomJS. Please use Puppeteer, or pass a component

to

> option.slots does not support strings in PhantomJS. Please use Puppeteer, or pass a component

38elements: I changed it.

Issue #246, opened 2017-12-09
To do before release


eddyerburgh:
Hi guys, vue-test-utils has been in beta for a couple months now, and the feedback we've got has been really helpful. 

There are a few things left to do before we release 1.0.0:

- [x] ~~Add deprecation notice to `hasClass`, `hasAttribute`, and `hasProp`, in favor of `classes`, `attributes`, and `props`~~
- [x] Add support for async components via Code Splitting #61
- [x] ~~Fix bug with keyed transitions~~ #52 
- [x] ~~Add docs for testing components that have asynchronous calls #151~~
- [x] ~~Add support for inheritAttrs #146~~
- [x]  ~~Add support for testing scoped slots #156~~
- [x] ~~Improve support for vue.extends #248~~ 
- [x] ~~Improve find to find Vue components without names~~
- [x] ~~Handle functional components in find~~ #296 
- [x] ~~Add setChecked, setValue, and setSelected methods~~ #530
- [ ] Improve support for extended components
- [ ] Improve scopedSlots option. Issue #809
- [ ] Fix synchronous bugs 

If anyone has any additional items they think should be added to the list, leave a comment 🙂

lmiller1990:
I think you have the wrong issue reference for `Add docs for testing components that have asynchronous calls` ?

I am interested in writing docs for this! This basically refers to anything using `async` or a Promise, right? 

Maybe we can put it under the guides section.

eddyerburgh:
Yes, I think we should add it to the guides section. That would be great if you could write it 🙂

lmiller1990:
I would like to do this, I will have time after the 19th.

Also, I think we should add something about _what_ to test. A lot of people have problems, when they are testing things using vue test utils that it shouldn't be used for (like testing other people's components).

Not directly related to this library, but I think providing an official library that will become a big part of the ecosystem, some solid guides on _how_ to use it, and not just what it does but what it does not/should not do would be valuable.

I gave a presentation last night about vue-test-utils at v-meetup in Tokyo, and talked to a bunch of people, and one of the main feedback points was "better ways to select things, like calendar inputs".

I think what people are doing is unit testing components that have child components which wrap <select> and <input type="date">, which I don't think should be the job of vue-test-utils (unless of course you are authoring a input wrapper component). The line between unit tests and integration tests is pretty blurred for a lot of people.

eddyerburgh:
Yeah I agree, I'll add it to the list

38elements:
[...]

Issue #213, opened 2017-11-28
Handling asynchronous errors


callumacrae:
I was having problems with some warnings being output to the console after the test had already passed, which was causing a lot of confusion. Sometimes I was also having issues—see <https://github.com/vuejs/vue/issues/7140>—with my actual tests not reporting the error in the correct way, instead the error being logged to the console and then the test timing out later.

I've ended up rewriting all of my tests, including the synchronous ones, to basically look like this:

```js
  it('is something', (done) => {
    Vue.config.errorHandler = done;

    // do test

    localVue.nextTick(done);
  });
```

(plus logic to make sure that `Vue.config.errorHandler` was set back to what it was supposed to be in `before` and `afterEach`)

There seem to be a lot of errors that are logged to the console in the next tick—for example, if you're using `<router-link>` and have `VueRouter` installed but haven't actually passed in a router—and these were causing a lot of issues in my tests because they were just being logged to the console.

Until I figured out that I had to do this let my test run for another tick and set up `Vue.config.errorHandler`, writing tests was really tricky! I was getting errors in the console, but they weren't failing the tests, and sometimes they were actually being output several tests later.

The stack trace also looked like this, but this might be a setup issue on my behalf:

```
ERROR LOG: TypeError{stack: 'render@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:51854:21
_render@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:10218:26
updateComponent@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:8532:28
get@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:8875:29
Watcher@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:8864:15
mountComponent@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:8536:28
$mount@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:13596:24
init@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:9811:19
createComponent@http://localhost:9876/base/index.js?b4bc4f750f67930cd8fde8390c76caa6f9a4df51:11263:10
[...]

PR #282, opened 2017-12-20
Add async guide

lmiller1990:
As discussed in #246 , some more guides for common pitfalls might be nice.

@eddyerburgh  this is just my first draft, I'd like some input on the content, style, etc. Thanks!

### DIFF
--- a/docs/en/guides/testing-async-components.md
+++ b/docs/en/guides/testing-async-components.md
@@ -30,6 +30,7 @@ One common cases is components that use `watch`, which updates asynchronously. B
     }
   }
 </script>
+```
 
 This component conditionally renders a submit button, based on whether the user entered some text. Let's see how we might test this:
 
@@ -92,7 +93,8 @@ export default {
 ```
 
 The below component makes an API call when a button is clicked, then assigns the response to `value`.
-```
+
+``` js
 <template>
   <button @click="fetchResults" />
 </template>
--- a/docs/en/guides/testing-async-components.md
+++ b/docs/en/guides/testing-async-components.md
@@ -72,9 +72,9 @@ test('Foo', () => {
 })
 ```
 
-The reason `$nextTick` or `setTimeout` allow the test to pass is because the microtask queue where promise callbacks are processed run efore the task queue, where `$nextTick` and `setTimeout` are processed. This means by the time the `$nexTick` and `setTimeout` run, any promise callbacks on the microtask queue will have been executed. See [here](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) for a more detailed explanation.
+The reason `$nextTick` or `setTimeout` allow the test to pass is because the microtask queue where promise callbacks are processed run before the task queue, where `$nextTick` and `setTimeout` are processed. This means by the time the `$nexTick` and `setTimeout` run, any promise callbacks on the microtask queue will have been executed. See [here](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) for a more detailed explanation.
 
-Another solution is to use the npm package, `flush-promises`, which can be used to flush all pending resolved promise handlers, and prepend the test with the ES7 'async' keyword. The ES7 `await` keyword with `flushPromises` can now be used to immediately resolve the API call.
+Another solution is to use an `async` function and the npm package `flush-promises`. `flush-promises` flushes all pending resolved promise handlers. You can `await` the call of `flushPromises` to flush pending promises and improve the readability of your test.
 
 The updated test looks like this:
 
--- a/docs/en/guides/testing-async-components.md
+++ b/docs/en/guides/testing-async-components.md
@@ -1,0 +1,6 @@
+# Testing Asynchronous Components
+
+> An example project will be made available.
+
+To simplify testing, `vue-test-utils` applies updates _synchronously_. However, there are some techniques you need to be aware ofwhen testing a component with asynchronous behavior such as callbacks or promises.
+
--- a/docs/en/guides/testing-async-components.md
+++ b/docs/en/guides/testing-async-components.md
@@ -2,7 +2,7 @@
 
 > An example project will be made available.
 
-To simplify testing, `vue-test-utils` applies updates _synchronously_. However, there are some techniques you need to be aware ofwhen testing a component with asynchronous behavior such as callbacks or promises.
+To simplify testing, `vue-test-utils` applies updates _synchronously_. However, there are some techniques you need to be aware of when testing a component with asynchronous behavior such as callbacks or promises.
 
 One common cases is components that use `watch`, which updates asynchronously. Below in an example of a component that renders some content based on a boolean value, which is updated using a watcher:
 
@@ -133,9 +133,9 @@ test('Foo', () => {
 })
 ```
 
-This test currently fails, because the assertion is called before the promise resolves. One solution is to use the npm package, `flush-promises`. which immediately resolve any unresolved promises. This test is also asynchronous, so like the previous example, we need to let the test runner know to wait before making any assertions. If you are using Jest, there are a few options, such as passing a `done()` callback, as shown above. 
+This test currently fails, because the assertion is called before the promise resolves. One solution is to use the npm package, `flush-promises`. which immediately resolve any unresolved promises. This test is also asynchronous, so like the previous example, we need to let the test runner know to wait before making any assertions. 
 
-Another is to prepend the test with the ES7 'async' keyword. We can now use the the ES7 `await` keyword with `flushPromises()`, to immediately resolve the API call.
+If you are using Jest, there are a few options, such as passing a `done()` callback, as shown above. Another is to prepend the test with the ES7 'async' keyword. We can now use the the ES7 `await` keyword with `flushPromises()`, to immediately resolve the API call.
 
 The updated test looks like this:
 
--- a/docs/en/guides/testing-async-components.md
+++ b/docs/en/guides/testing-async-components.md
@@ -1,86 +1,8 @@
-# Testing Asynchronous Components
-
-> An example project will be made available.
+# Testing Asynchronous Behavior
 
 To simplify testing, `vue-test-utils` applies updates _synchronously_. However, there are some techniques you need to be aware of when testing a component with asynchronous behavior such as callbacks or promises.
 
-One common cases is components that use `watch`, which updates asynchronously. Below in an example of a component that renders some content based on a boolean value, which is updated using a watcher:
-
-``` js
-<template>
-  <div>
-    <input v-model="text" />
-    <button v-if="showButton">Submit</button>
-  </div>
-</template>
-
-<script>
-  export default {
-    data () {
-      return {
-        text: '',
-        showButton: false
-      }
-    },
-
-    watch: {
-      text () {
-        this.showButton = true
-      }
-    }
-  }
-</script>
-```
-
-This component conditionally renders a submit button, based on whether the user entered some text. Let's see how we might test this:
-
-``` js
-import { shallow } from 'vue-test-utils'
-import Foo from './Foo'
-
-describe('Foo', () => {
-  it('renders a button async when the user enters text', () => {
-    expect.assertions(2)
-    const wrapper = shallow(Foo)
-
-    wrapper.find('input').element.value = 'Value'
-    wrapper.find('input').trigger('input')
-
-    expect(wrapper.vm.showButton).toBe(true)
-    expect(wrapper.find('button').exists()).toBe(true)
-  })
-})
-```
-
-The above test fails - however, when you test it by hand in a browser, it seems fine!  The reason the test is failing is because properties inside of `watch` are updated asynchronously. In this test, the assertion occurs before `nextTick()` is called, so the `button` is still not visible.
-
-Most unit test libraries provide a callback to let the runner know when the test is complete. Jest and Karma both use `done()`. We can use this, in combination with `nextTick()`, to test the component.
-
-
-``` js
-import { shallow } from 'vue-test-utils'
-import Foo from './Foo'
-
-describe('Foo', () => {
-  it('renders a button async when the user enters text', (done) => {
-    expect.assertions(2)
-    const wrapper = mount(Foo)
-
-    wrapper.find('input').element.value = 'Value'
-    wrapper.find('input').trigger('input')
-
-    wrapper.vm.$nextTick(() => {
-      expect(wrapper.vm.showButton).toBe(true)
-      expect(wrapper.find('button').exists()).toBe(true)
-      done()
-    })
-  })
-})
-```
-
-Now the test passes! 
-
-Another common asynchronous behavior is API calls and Vuex actions. The following examples shows how to test a method that makes an API call. This example is using Jest to run the test and to mock the HTTP library `axios`.
+One of the most common asynchronous behaviors is API calls and Vuex actions. The following examples shows how to test a method that makes an API call. This example is using Jest to run the test and to mock the HTTP library `axios`.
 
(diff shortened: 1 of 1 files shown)


docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,159 @@
...
+describe('Foo', () => {
+  it('renders a button async when the user enters text', (done) => {
+    expect.assertions(2)
+    const wrapper = mount(Foo)
+
+    wrapper.find('input').element.value = 'Value'
+    wrapper.find('input').trigger('input')
+
+    wrapper.vm.$nextTick(() => {
+      expect(wrapper.vm.showButton).toBe(true)
```
callumacrae: If this assertion throws, will the error be caught, or will the test time out? In my experience, it's the latter.

callumacrae: Mind you, I've been using mocjha. might be different with jest

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,159 @@
+# Testing Asynchronous Components
```
LinusBorg: This headline might be confused with ["Async components"](https://vuejs.org/v2/guide/components.html#Async-Components).

Maybe: "Testing Asynchronous Behaviour" ?

docs/en/guides/testing-async-components.md
lmiller1990: I think it will time out @callumacrae . What is the best thing to do in this situation?

docs/en/guides/testing-async-components.md
lmiller1990: Good catch. I think your suggested title is better. Mintor, but In my day to day life I spell it behaviour, should use use American English (no u) or English/Europe/everywhere else English?

lmiller1990:
@eddyerburgh  good idea about mentioning `update`. What exactly does `update` do, just force `nextTick`?

docs/en/guides/testing-async-components.md
LinusBorg: Hm. I think we should probably stick do American english as that's the bigger "market"?

OTOH @eddyerburgh is british so ..., :-P

eddyerburgh: Indeed I am British 🇬🇧

eddyerburgh:
@lmiller1990 At the moment, it forces a component re-render.

I think we should also make it run all the component and child component watchers. I'm trying to think whether this would have unintended consequences 🤔

docs/en/guides/testing-async-components.md
lmiller1990: I am Australian (English colony) so we use the same style. But I agree we should use the American English since I believe the official docs would use that.

docs/en/guides/testing-async-components.md
callumacrae: not sure. i made an issue about it here: https://github.com/vuejs/vue-test-utils/issues/213

docs/en/guides/testing-async-components.md
eddyerburgh: Yeah, we should use US English

lmiller1990:
Removed `watch` example, since that'll be solved in #283 .

Will look add adding more about `update`. Can you think of any good examples where running `update` is important for async tests @eddyerburgh ? I haven't had to use it much. Or were you just thinking to add something so users are more aware of how `vue-test-utils` works, in a synchronous fashion?

38elements:
I seem that `flushPromises` does not work for asynchronous processing executed in `Promise`.
IMHO, I think it may be is better to introduce `sleep` as below.

```js
export default function sleep (ms) {
  return new Promise(resolve => setTimeout(resolve, ms))
}
```

eddyerburgh:
In fact, calling a synchronous method that updates state will also require calling `update`.
Any time state is changed, you need to call `update`. In all vue-test-utils methods that change or can change state, like `trigger`, `setData`, `setProps`, `update` is called to trigger a re render.

If state is changed outside of vue-test-utils, by mutating the state directly, or by triggering a method, `update` must be called to update the render tree.

@lmiller1990 An example for this guide would be when you trigger an async method by calling it on the `vm`:

```js
const Component = {
  template: `<div>{{someData}}</div>`,
  data() {
    return {
      someData: 0
    }
  },
  methods: {
    asyncMethod() {
     Promise.resolve(() => {
      this.someData++
     })
    }
  }
}

const wrapper = mount(Component)
wrapper.vm.asyncMethod()
await flushPromises()
wrapper.update()
expect(wrapper.text()).toBe(1)
```

lmiller1990:
@eddyerburgh  I haven't actually ran into this issue - calling the methods on the wrapper seems to actually implicitly call `update` as well. The following (basically what you wrote above) actually passes.

``` js
<template>
  <div>
    {{someData}}
  </div>
</template>

<script>
  export default {
    data () {
      return {
        someData: 0
      }
    },
    methods: {
      asyncMethod () {
        Promise.resolve(() => {
        }).then(() => {
          this.someData += 1
        })
      }
    }
  }
</script>
```

``` js
import {shallow} from 'vue-test-utils'
import Update from './Update'
import flushPromises from 'flush-promises'

describe('Update', () => {
  it('works', async () => {
    const wrapper = shallow(Update)
    wrapper.vm.asyncMethod()
    await flushPromises()
    expect(wrapper.vm.someData).toBe(1)
    expect(wrapper.text()).toBe('1')
  })
})

```

I haven't actually hit any situations where I've had to call `update`. Do you have another example?

eddyerburgh:
Ah—when flush-promises has finished, the component will have re rendered.

Looks like we don't need to include `updates` in this guide 😂

lmiller1990:
@eddyerburgh  ok. Is there anything else you think I should change or improve? Would be nice to get this in.

Also, I fixed up the failing tests regarding `no trailing whitespace`, but it appears to be actually linting the example code, and complaining that `await is a reserved keyword` and it also doesn't like `@click`. Any way to ignore these?

wtho:
Just a thought:. 
What if the `asyncMethod` is not promise-based, but e. g. callback-based? I know it is less often and maybe bad style, but it should still be possible to test this code.

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,81 @@
...
+
+The below component makes an API call when a button is clicked, then assigns the response to `value`.
+
+``` js
+<template>
+  <button @click="fetchResults" />
+</template>
+
+<script>
+  import axios from 'axios'
```
eddyerburgh: Could you add a line break after the import statement

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,81 @@
...
+    methods: {
+      async fetchResults () {
+        const response = await axios.get('mock/service')
+        this.value = response.data
+      }
+    }
+  }
+</script>
+```
+A test can be written like this:
```
eddyerburgh: Can you add a link to the Jest manual mocks page—https://facebook.github.io/jest/docs/en/manual-mocks.html#content

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,81 @@
...
+test('Foo', () => {
+  it('fetches async when a button is clicked', () => {
+    const wrapper = shallow(Foo)
+    wrapper.find('button').trigger('click')
+    expect(wrapper.vm.value).toEqual('value')
+  })
+})
+```
+
+This test currently fails, because the assertion is called before the promise resolves. One solution is to use the npm package, `flush-promises`. which immediately resolve any unresolved promises. This test is also asynchronous, so like the previous example, we need to let the test runner know to wait before making any assertions.
```
eddyerburgh: > which immediately resolve any unresolved promises
to
> which can be used to flush all pending resolved promise handlers

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,81 @@
...
+The updated test looks like this:
+
+``` js
+import { shallow } from 'vue-test-utils'
+import flushPromises from 'flush-promises'
+import Foo from './Foo'
+jest.mock('axios')
+
+test('Foo', () => {
+  it('fetches async when a button is clicked', () => {
```
eddyerburgh: You need to add async to the function, that's why there's a linting error:
`it('fetches async when a button is clicked', () => {`

to

` it('fetches async when a button is clicked', async () => {`

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,81 @@
...
+    const wrapper = shallow(Foo)
+    wrapper.find('button').trigger('click')
+    expect(wrapper.vm.value).toEqual('value')
+  })
+})
+```
+
+This test currently fails, because the assertion is called before the promise resolves. One solution is to use the npm package, `flush-promises`. which immediately resolve any unresolved promises. This test is also asynchronous, so like the previous example, we need to let the test runner know to wait before making any assertions. 
+
+If you are using Jest, there are a few options, such as passing a `done` callback, as shown above. Another is to prepend the test with the ES7 'async' keyword. We can now use the the ES7 `await` keyword with `flushPromises`, to immediately resolve the API call.
```
eddyerburgh: >such as passing a `done` callback, as shown above

It isn't shown above. Can you add an example using `done`. Instead of using flushPromises you can just use a setTimeout:

```js
test('Foo', (done) => {
   it('fetches async when a button is clicked', () => {
     const wrapper = shallow(Foo)
     wrapper.find('button').trigger('click')
     setTimeout(() => {
          expect(wrapper.vm.value).toEqual('value')
          done()
     })
   })
 })
```

(note I haven't tested this code)

We should add a note on why this works, with a link to this blog post—https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/. Basically, the microtask queue, which processes promise callbacks, runs before the task queue, which processes timeout callbacks. So when a setTimeout callback is executed, any resolved promises will have been executed.

lmiller1990:
Thanks for the feedback @eddyerburgh .

Updates:

1. Add example using `done`
2. Include the option to use `$nextTick` or `setTimeout` (both seem fine, I ran the tests. Do you have a preference? I used `$nextTick` in the example code, but mention both in the explanation.
3. Include link about why `setTimeout` works. I didn't know why, that was a really interesting read.
4. Improve some wording.
5. Run all tests locally (was doing so anyway, but just to check).

docs/en/guides/testing-async-components.md
```diff
@@ -55,9 +57,24 @@ test('Foo', () => {
...
+    wrapper.find('button').trigger('click')
+    wrapper.vm.$nextTick(() => {
+      expect(wrapper.vm.value).toEqual('value')
+      done()
+    })
+  })
+})
+```
+
+The reason `$nextTick` or `setTimeout` allow the test to pass is because the microtask queue where promise callbacks are processed run efore the task queue, where `$nextTick` and `setTimeout` are processed. This means by the time the `$nexTick` and `setTimeout` run, any promise callbacks on the microtask queue will have been executed. See [here](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) for a more detailed explanation.
```
eddyerburgh: small typo — `efore`

docs/en/guides/testing-async-components.md
```diff
@@ -55,9 +57,24 @@ test('Foo', () => {
...
 
-If you are using Jest, there are a few options, such as passing a `done` callback, as shown above. Another is to prepend the test with the ES7 'async' keyword. We can now use the the ES7 `await` keyword with `flushPromises`, to immediately resolve the API call.
+Another solution is to use the npm package, `flush-promises`, which can be used to flush all pending resolved promise handlers, and prepend the test with the ES7 'async' keyword. The ES7 `await` keyword with `flushPromises` can now be used to immediately resolve the API call.
```
eddyerburgh: Can we reword this:
>Another solution is to use the npm package, `flush-promises`, which can be used to flush all pending resolved promise handlers, and prepend the test with the ES7 'async' keyword. The ES7 `await` keyword with `flushPromises` can now be used to immediately resolve the API call.

To something like this:
>Another solution is to use an `async` function and the npm package `flush-promises`. `flush-promises` flushes all pending resolved promise handlers. You can `await` the call of `flushPromises` to flush the promises and improve the readability of your test.

lmiller1990:
Done! Thanks for the feedback.

doun:
@lmiller1990 Would u add an example for catching exceptions:
`
try{
await a();
expect(CONDITION);
await b();
expect(CONDITION);
}
catch(e){
}
`
be careful that exception throw by expect() should NOT be caught, or it will pass all tests

lmiller1990:
@doun  the current test shouldn't swallow any exceptions, but throw an error. I haven't experienced what your describe before, can you link to an example/post some sample code? Or do you mean I should add an example showing how to test if an error was throw from some async code?

doun:
It's not about current sample code.
I just did this days ago, sample code:
```javascript
test("expect fail", async ()=>{
try{
  var resultA = await TaskA()
  expect(resultA).toBe(true)
  var resultB = await TaskB()
  expect(true).toBe(false)
}
//all exceptions caught, the test will pass
catch(e){
}
})
```
I wanted to catch only async exceptions, but the test framework(jest) also use exceptions to catch expect failures.

lmiller1990:
@doun Thanks for the sample. It appears you are just testing regular JavaScript methods, not Vue component specific ones? Or can you provide the component and some more context?

I'm not sure this belongs here -- of course it is important, but it doesn't feel specific enough to Vue components to include.

One good practise can be to use `expect.assertions` at the top of your tests - that way, if an error occurs and Jest skips to the `catch` block, because `expect.assertions` is not the correct amount, the test will fail.

doun:
No, it’s about async test, not about vue component test. Your idea is awesome, thanks!

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,98 @@
...
+test('Foo', () => {
+  it('fetches async when a button is clicked', () => {
+    const wrapper = shallow(Foo)
+    wrapper.find('button').trigger('click')
+    expect(wrapper.vm.value).toEqual('value')
+  })
+})
+```
+
+This test currently fails because the assertion is called before the promise in `fetchResults` resolves. Most unit test libraries provide a callback to let the runner know when the test is complete. Jest and Karma both use `done`. We can use `done` in combination with `$nextTick` or `setTimeout` to ensure any promises resolve before the assertion is made.
```
eddyerburgh: > Jest and Karma both use 

should be 

Jest and Mocha both use

docs/en/guides/testing-async-components.md
```diff
@@ -0,0 +1,98 @@
+# Testing Asynchronous Behavior
+
+To simplify testing, `vue-test-utils` applies updates _synchronously_. However, there are some techniques you need to be aware of when testing a component with asynchronous behavior such as callbacks or promises.
```
eddyerburgh: > applies updates _synchronously_

to

>  applies DOM updates _synchronously_

lmiller1990:
Thanks @eddyerburgh , I added those changes.

eddyerburgh:
Thanks @lmiller1990 🙂

lmiller1990:
Thank you @eddyerburgh , the part about how `setTimeout` and task/microtasks was new to me. OSS is great. I feel like I've learned more doing OS in the last year than like 4 years at school at 2 in the workforce.

Ray-56:
if
image had setTimeout 1500s, expected is error. 
how do i handle it?

lmiller1990:
@guokangf can you post your question on stackoverflow with `vue-test-utils` tag and the test you trying to write? I can help you there. This repo is just for issues with `vue-test-utils`.

Thanks.

PR #288, opened 2017-12-22
Add sleep

38elements:


eddyerburgh:
Thanks for the PR, but I don't think we should add something like sleep to the library. You can start an issue and see what others think, but my opinion is that anything not related to Vue shouldn't be included in the library

38elements:
I think that when asynchronous processing is executed in `Promise`, it needs to sleep asynchronously.

38elements:
I see.
I will make issue.

38elements:
I think it is necessary because it is only 3 lines.

PR #293, opened 2017-12-25
Add passing false to options.stubs in options.md

38elements:
This adds the document for `options.stubs` when `options.stubs` is false.

### DIFF
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -85,14 +85,16 @@ mount(Component, {
   stubs: ['registered-component']
 })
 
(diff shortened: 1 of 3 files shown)


38elements:
I do not understand why it becomes error when it is executed in browser.


```js
  it('does not stub components when passed false', () => {
    config.stubs.transition = TransitionStub
    const testComponent = {
      template: `
        <div>
          <transition><p /></transition>
        </div>
      `
    }
    const wrapper = mount(testComponent, {
      stubs: {
        'transition': false
      }
    })
    expect(wrapper.contains(TransitionStub)).to.equal(false)
  })
```

Issue #302, opened 2017-12-27
Incorrect coverage result


38elements:
Since there are the tests using the files in `src` directory like `test/unit/specs/error-wrapper.spec.js`,  the coverage result is incorrect.

https://github.com/vuejs/vue-test-utils/blob/063cd8f0e24c098dc6f6eca0a86f8981d24b46f2/test/unit/specs/wrappers/error-wrapper.spec.js#L1

Since the test is executed to check `dist`, I think that using for `src` is not appropriate.

eddyerburgh:
Agreed.

The wrapper-array test also imports from the `src` directory—https://github.com/vuejs/vue-test-utils/blob/dev/test/unit/specs/wrappers/wrapper-array.spec.js

38elements:
I will send the pull request for `wrapper-array.spec.js`.

PR #303, opened 2017-12-27
refactor test/unit/specs/wrappers/error-wrapper.spec.js

38elements:
This is related to #302 .


eddyerburgh:
Thanks 🙂

the wrapper-array spec file also imports from src—https://github.com/vuejs/vue-test-utils/blob/dev/test/unit/specs/wrappers/wrapper-array.spec.js

Issue #165, opened 2017-11-10
How to do code coverage ?


lionel-bijaoui:
Hello,

I have followed the guide to use [vue-test-utils with mocha](https://vue-test-utils.vuejs.org/en/guides/testing-SFCs-with-mocha-webpack.html).
It work great, no problem.

Then I wanted to add code coverage, since my initial config (based on [the official vue+webpack template]( https://github.com/vuejs-templates/webpack)) came with it.
But I can't seem to make it work. I have followed [mocha-webpack guide on code coverage](https://github.com/zinserjan/mocha-webpack/blob/master/docs/guides/code-coverage.md) and it doesn't work.

Can you add an example of code coverage in [the official example](https://github.com/vuejs/vue-test-utils-mocha-webpack-example) ?

Thanks a lot for your hard work !

eddyerburgh:
Hi @lionel-bijaoui 

I agree, it would be nice to add these to the docs 😀

zhouweicsu:
[Vue-test-utils with jest](https://vue-test-utils.vuejs.org/en/guides/testing-SFCs-with-jest.html) also need an official example 😀

bdeo:
@zhouweicsu What kind of example are you interested in? Project structure? package.json jest config?

robcresswell:
I can document the process with Jest if you like; I've got a running setup locally. Would that help, @eddyerburgh ?

robcresswell:
I've opened a PR for this repo; I'll add similar code to the example repo too, and link here when it's done.

robcresswell:
^^ As promised, similar code in the `vue-test-utils-jest-example` repo. Please comment and review, and enjoy the holidays! 🎅

eddyerburgh:
I'm reopening, because we still need to add docs on code coverage with mocha-webpack

robcresswell:
I won't be able to pick that part up; I've no experience with Mocha, I'm afraid, and I'm sure there are others better positioned to explain it's usage.

eddyerburgh:
I think we could just add a link to the mocha-webpack docs—https://github.com/zinserjan/mocha-webpack/blob/master/docs/guides/code-coverage.md, I'll make a PR

38elements:
The mocha-webpack document is an explanation for testing `js file`.
IMHO,
I think the method to test `vue file` is necessary.

eddyerburgh:
We've added code coverage to the official repo, thanks @38elements 🙂

[...]

PR #309, opened 2017-12-28
Update testing-SFCs-with-mocha-webpack.md

38elements:
`vuejs-templates/webpack` can setup to the code coverage for vue file.
I think it is the great example for this solution. 
This is related to #165 .


eddyerburgh:
We need a section on setting up code coverage for mocha-webpack, the link you posted is for karma-webpack which use karma to run the tests, instead of mocha-webpack.

38elements:
Thank you for reply.

PR #321, opened 2018-01-02
docs: update testing-SFCs-with-mocha-webpack.md

38elements:


eddyerburgh:
Thanks @38elements .

I don't think we need to add this link, since we already have a link to the example project in the resource section.

Issue #327, opened 2018-01-04
How to check an element is displayed (v-show) ?


Toilal:
I try to assert that elements are displayed or not. Those elements are children (or even grand-children) of a v-show node.

Consider the following, I wish to assert display of `.item` elements.

```html
<div>
  <div v-show="modelAttrA">
    <span class="item">Ax</span>
    <span class="item">Ay</span>
  </div>
  <div v-show="modelAttrB">
    <span class="item">Bx</span>
    <span class="item">By</span>
  </div>
<div>
``` 

It seems a common use case, but I can't find any way in vue-test-utils to do that. I can't rely on DOM element properties as I run tests with JEST (JSDOM doesn't layout elements and can't tell if an element is displayed neither). Maybe there's something possible with VueJS VNode API ?

I would expect some kind of `displayed` method in the Wrapper API.
```
expect(wrapper.findAll('.item').at(3).displayed()).toBeTruthy()
```

lmiller1990:
From the Vue.js docs,

`v-show`: simply applies `display: none`
`v-if`: removes element from dom entirely.

Here is an example:
```html
<template>
  <button v-show="false" />
</template>
```
```
// testing v-show
expect(wrapper.find('button').hasStyle('display', 'none')).toBe(true)

// testing v-if
expect(wrapper.find('button').exists()).toBe(false)
```

I think this kind of q and a is best for stackoverflow, though. Does that help?
  
Edit: I just realized you are looking at children elements of `v-show`. It's the same thing, though. If the parent is not shown, than the children are not shown, either. So just assert parent is not shown.

Toilal:
I understand, but I can't find `parent()` nor `ancestor()` method to check for `display: none` style on ancestors of element to assert, and I don't think it's possible to build a CSS selector that would match elements with such ancestors.

It's more a feature request than a Q&A, as a (new) user of vue-test-utils, I just expect to find a `visible()` method on Wrapper to assert that an element is visible.

Toilal:
Here's a possible implementation (Working with JEST/JSDOM) :

``` typescript
function isVisible(wrapper: Wrapper<Vue>) {
  const computedStyle = window.getComputedStyle(wrapper.element)
  return computedStyle.display !== 'none'
}
```

May I open a pull request that adds a `displayed()` method on wrapper with this implementation ? it seems displayed is the best name, as it relies on `display: none` CSS style.

Toilal:
[...]

PR #328, opened 2018-01-04
feat: export createWrapper and add createWrapperArray functions

Toilal:
In some use case, it can be handy to manually create Wrapper and WrapperArray instances.

For example, you can find elements with `findAll` method, then filter them manually with `wrappers` property and Array.filter method, and finally create a new filtered WrapperArray instance.

```javascript
const wrapper = mount(SomeComponent)
const filteredSubComponents = createWrapperArray(wrapper.findAll(SubComponent).wrappers.filter(w => !!w.vm.v))
...
```

### DIFF
--- a/src/wrappers/create-wrapper-array.js
+++ b/src/wrappers/create-wrapper-array.js
@@ -1,0 +1,9 @@
+// @flow
+
+import WrapperArray from './wrapper-array'
+import type Wrapper from './wrapper'
+import type VueWrapper from './vue-wrapper'
+
+export default function createWrapperArray (wrappers: Array<Wrapper | VueWrapper>) {
+  return new WrapperArray(wrappers)
+}
--- a/src/index.js
+++ b/src/index.js
@@ -1,12 +1,16 @@
 import shallow from './shallow'
 import mount from './mount'
 import createLocalVue from './create-local-vue'
+import createWrapper from './wrappers/create-wrapper'
+import createWrapperArray from './wrappers/create-wrapper-array'
 import TransitionStub from './components/TransitionStub'
 import TransitionGroupStub from './components/TransitionGroupStub'
 import config from './config'
 
 export default {
   createLocalVue,
+  createWrapper,
+  createWrapperArray,
   config,
   mount,
   shallow,
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-import Vue, { VNodeData, ComponentOptions, FunctionalComponentOptions } from 'vue'
+import Vue, { VNode, VNodeData, ComponentOptions, FunctionalComponentOptions } from 'vue'
 
 // TODO: use core repo's Component type after https://github.com/vuejs/vue/pull/7369 is released
 export type Component =
@@ -126,6 +126,8 @@ interface VueTestUtilsConfigOptions {
 }
 
 export declare function createLocalVue (): typeof Vue
+export declare function createWrapper (node: VNode | Component, update?: Function, options?: WrapperOptions): Wrapper<Vue>
+export declare function createWrapperArray<T extends Vue> (wrappers: Array<Wrapper<T>>): WrapperArray<T>
 export declare let config: VueTestUtilsConfigOptions
 
 export declare function mount<V extends Vue> (component: VueClass<V>, options?: ThisTypedMountOptions<V>): Wrapper<V>


emileber:
Could we export the types as well? `Wrapper` and `WrapperArray`

eddyerburgh:
@emileber we could definitely export the types, would you like to make a PR? 

Personally I'm hesitant to export these methods.  I appreciate it could be useful to create a wrapper array for some users, but it's a low-level function. Most users won't be aware of how to create WrapperArrays, and I don't think we should make them aware.

We only expose a few methods as methods to use:
- mount
- shallow
- createLocalVue

(We also expose components, but they aren't featured in the top-level API in the docs).

I feel like createWrapper and createWrapperArray add a lot of noise, for a niche use-case.

That's my thoughts. If others feel like this is a really valuable feature then we could add, but I'm against adding it to the API.

Toilal:
By exporting types, do you mean exporting classes ? Because [types are already exported for TypeScript users](https://github.com/vuejs/vue-test-utils/blob/dev/types/index.d.ts#L71-L101).

If classes Wrapper and WrapperArray are exported it will be just same as exporting createArray and createWrapperArray, but user will have to use new keyword and constructor parameters are less obvious than those factory methods.

But anyway,  instead of this pull request, would you consider adding a filtering feature on all methods involving a selector (find/findAll/...) by adding an optional predicate function parameter ? It would really help for my use case (filtering out non-visible elements: #327)

Toilal:
and maybe a filter method on WrapperArray ?

```javascript
const wrapper = mount(SomeComponent)
const filteredSubComponents = wrapper.findAll(SubComponent).filter(w => !!w.vm.v)
```

38elements:
How about changing `wrapperArray.length` from a fixed value to `getter` ?

Currently, `wrapperArray.length` is a fixed value.

https://github.com/vuejs/vue-test-utils/blob/fce6e6e8dbc51a5a4effb1ff3f447ec33461daaa/src/wrappers/wrapper-array.js#L7-L14

`wrapperArray.length` becomes `getter`, as below.

```js
export default class WrapperArray implements BaseWrapper {
  wrappers: Array<Wrapper | VueWrapper>;
  length: number;

  constructor (wrappers: Array<Wrapper | VueWrapper>) {
    this.wrappers = wrappers || []
  }
  
  get length() {
    return this.wrappers.length
  }
  
  // ...
}
```


```js
const wrapper = mount(SomeComponent)
const wrapperArray = wrapper.findAll(SubComponent)
wrapperArray.wrappers = wrapperArray.wrappers.filter(w => !!w.vm.v)
```

Toilal:
You may have a look to #334 as an alternative. I can't see any other use case than "filtering" for manually creating WrapperArray.

emileber:
@Toilal By exporting the types, I just meant exposing the `Wrapper` and `WrapperArray` classes outside of this lib. Making it possible for devs to manually wrap an existing Vue instance and use the wrapper utilities.

Unless it's already possible and I just didn't see where in the doc?

eddyerburgh:
@emileber I thought you meant exposing the flow types. I don't think we should expose the Wrapper and WrapperArray classes. What is your use case? Also, please make an issue so we can discuss this further

eddyerburgh:
@38elements I like your suggestion 😀

38elements:
The following code works.

```js
const wrapper = mount(ComponentWithProps)
const foo = wrapper.findAll('div') 
const WrapperArray = foo.constructor
const bar = new WrapperArray(foo.wrappers)
expect(bar.length).to.equal(3)
```

emileber:
@38elements While it could work, it could eventually fail if the return value of `findAll` was to change. Since it's not in the public API, we should avoid such _hacks_.

eddyerburgh:
Thinking about it, I think this could be the best solution to the problems you've had (we could also add the `filter` method as you suggested @Toilal )

phanan:
Chiming in here. Exported `Wrapper` and `WrapperArray` can be useful to create additional helpers. Think plugins/utils for vue-test-utils.

emileber:
@phanan What do you have in mind?

phanan:
@emileber I'm writing a helper library on top of vue-test-utils which provides additional methods or wrappers – those may come in handy but don't belong to the core.

emileber:
@phanan Something like vue-test-utils-_utils_? hehe See #337 which initially was about exposing `Wrapper` and `WrapperArray`.

src/wrappers/create-wrapper-array.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+import WrapperArray from './wrapper-array'
+import type Wrapper from './wrapper'
+import type VueWrapper from './vue-wrapper'
+
+export default function createWrapperArray (wrappers: Array<Wrapper | VueWrapper>) {
```
eddyerburgh, line 7: Can you write a few tests for this method and add validation (so check that Wrappers is an array of wrappers and throw an error if not)

eddyerburgh (commented):
We need to add validation to createWrapper and createWrapperArray when we expose them. Can you add tests that check createWrapper and createWrapperArray work correctly, and that they validate the input. The tests should be top-level (e.g. test/unit/specs/createWrapper.spec.js)

eddyerburgh:
@Toilal are you able to make the changes?

38elements:
https://github.com/vuejs/vue-test-utils/pull/328#issuecomment-358378538

> Chiming in here. Exported Wrapper and WrapperArray can be useful to create additional helpers. Think plugins/utils for vue-test-utils.

I agree with this comment.
I think providing Wrapper class and WrapperArray class is better than `createWrapper()`.

PR #334, opened 2018-01-08
Add filter feature on Selector functions and WrapperArray

Toilal:
This is an alternative to #328 to allow user to filter elements from WrapperArray.

### DIFF
--- a/docs/en/api/README.md
+++ b/docs/en/api/README.md
@@ -43,6 +43,7 @@
 * [WrapperArray](./wrapper-array/README.md)
   * [at](./wrapper-array/at.md)
   * [contains](./wrapper-array/contains.md)
+  * [filter](./wrapper-array/filter.md)
   * [exists](./wrapper/exists.md)
   * [destroy](./wrapper-array/destroy.md)
   * [hasStyle](./wrapper-array/hasStyle.md)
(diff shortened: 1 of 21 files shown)


eddyerburgh:
Thanks for the PR. 

I'm not sure we should add this feature. We talked about adding array methods a few months ago and decided against it as it would make the API to large. Perhaps we could make an exception for `filter`, but my concern would be that users would expect other array methods to be implemented.

eddyerburgh:
I'm also not sure about adding a predicate to methods that take a selector. We should start a feature request issue and get some feedback.

Toilal:
I need either this feature or #337, or I just can't use vue-test-utils for my component.

eddyerburgh:
You can use the suggestion 38elements gave:

```js
wrapperArr = findAll('div')
wrapperArr.wrappers = wrapperArr.wrappers.filter(w => w.element.style.visibility === 'hidden')
```

Toilal:
On TypeScript, `wrappers` is marked as readonly so it won't compile. And I think `length` property will not be updated too.

eddyerburgh:
Ah, well we could potentially update the length property. I understand your use case, and we'll find a solution,  but we need to discuss it first.

eddyerburgh (changes requested):
We'll add the filter method, but we're not going to add a predicate to selector methods for now. Can you remove the predicate from this PR, or create a new PR that adds the filter method?

eddyerburgh:
Thanks for the PR.

I've added the `filter` method in another PR, so we can close the issue.

Issue #65, opened 2017-10-02
Method that return parent


lusarz:
In one of my test scenario I need to get parent of element, something like:

`wrapper.find(selector).parent().hasClass('disabled')`

What do you think about adding `parent` method to wrapper ?

eddyerburgh:
I would prefer to keep the API small. This would require adding a `parent` property to the Wrapper class.

I'd like hear what other people think 🙂

Austio:
I agree @eddyerburgh.  Initial thoughts are that this would add a fair amount of complexity and create some ambiguity on whether it is the parent vue component or parent html node.

eddyerburgh:
I'm closing this as we don't intend to implement this at the moment

emileber:
Since `Wrapper` isn't exposed, we can't just create a new `Wrapper` from `wrapper.vm.parent`. 

#337 would help in this case.

eddyerburgh:
This feature request is for a parent method for a vnode, not a Vue instance. You wouldn't be able to access a vnode parent, the only way would be to add a parent property to each vnode inside vue-test-utils.

anorborg:
Is there any recommended workaround to access the parent?

natchiketa:
@anorborg Here's my workaround, fwiw:

> TL;DR: the Wrapper instance exposes an `element` property, which is a regular ol' JavaScript DOM `Element`. You can use that to get to a parent, sibiling, etc.

## Use case
I have a event calendar component that is a week or month calendar, and I'm testing that the event fixture from the tests and the link to go to the route for just that calendar day match. The markup looks like this:

```vue
        <li v-for="(day, index) in calendarDays" :key="index" class="day">
          <div class="date">{{ day.date.format('DD') }}</div>
          <div v-for="(event, index) in eventsOnDay(day.date)" :key="index" class="event"> {{ // stuff }}</div>
          <router-link :to="detailPath(day.date)" class="router-link" />
        </li>
```

I want to find the `.event` element (there's only one event in this test), then go back up to its parent so I can get to `.event`'s sibiling, the `<router-link />`, and make sure that the link's url matches the event time. Yada yada, real-world use case.

```js
  it('has links to the detail view', async function() {
    const wrapper = await mount(EventCalendar, { localVue, store, router })
    const eventWrapper = wrapper.find('.event') // this returns a Wrapper instance
    // parentElement is the <li />, then you can use querySelector to find the sibling, so
    // detailLink is also an Element instance
[...]

PR #335, opened 2018-01-08
feat(visible): Add visible() method on Wrapper (#327)

Toilal:
Ready ! :)


eddyerburgh:
Thanks for the PR 😀

I don't think we should add a parent method as part of this PR, or at all —https://github.com/vuejs/vue-test-utils/issues/65

You should use the element parentNode/parentElement properties to traverse the parent tree. 

Like you found, most vnodes have an undefined parent. Functional vnodes use the parent property.

Toilal:
I just removed `parent()`, implement `visible()` using HTMLElement, and added tests. It seems to work now :p. 

`visible()` is also consistent with `exists()` (if `exists() == false` then `visible() == false`)

package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "build": "node build/build.js",
-    "build:test": "NODE_ENV=test node build/build.js",
+    "build:test": "cross-env NODE_ENV=test node build/build.js",
```
eddyerburgh, line 13: Thanks, good spot

src/wrappers/wrapper.js
```diff
@@ -120,6 +120,30 @@ export default class Wrapper implements BaseWrapper {
     return true
   }
 
+  /**
+   * Utility to check wrapper is visible. Returns false if a parent element has display: none or visibility: hidden style.
+   */
+  visible (): boolean {
+    if (!this.exists()) {
+      console.log('NOT EXISTS')
```
eddyerburgh: We should throw an error here using the `throwError` function:
`throwError('cannot call visible on a wrapper without an element')`

src/wrappers/wrapper.js
```diff
@@ -120,6 +120,30 @@ export default class Wrapper implements BaseWrapper {
     return true
   }
 
+  /**
+   * Utility to check wrapper is visible. Returns false if a parent element has display: none or visibility: hidden style.
+   */
+  visible (): boolean {
+    if (!this.exists()) {
```
eddyerburgh: You should check for an element here instead of exists—`if(!this.element)`. That's how we check for an element in other tests (I think this might solve the flow error)

src/wrappers/wrapper.js
```diff
@@ -120,6 +120,30 @@ export default class Wrapper implements BaseWrapper {
...
+    while (element) {
+      if (element.style && (element.style.visibility === 'hidden' || element.style.display === 'none')) {
+        return false
+      }
+      if (!element.parentElement) {
+        break
+      }
+      element = element.parentElement
+    }
+    console.log('no parent')
```
eddyerburgh: We should remove these console logs

eddyerburgh (changes requested):
Looks good, just a couple of changes. Also we need the flow error to be fixed

eddyerburgh:
Are you able to make these changes @Toilal ?

Toilal:
Oops, I just forget this one, i'll do ASAP

Toilal:
Done

38elements:
I think it is necessary to add this to the document.

eddyerburgh:
Yes this should be added to the documentation before merging

eddyerburgh:
@Toilal Can you add a page on visible to docs/en/wrapper/api/visible.md, docs/en/api/wrapper-array/visible.md, and add a links to both pages in docs/en/SUMMARY.

Toilal:
Rebased and docs added.

Toilal:
Can you restard the build ?

docs/en/api/wrapper/visible.md
```diff
@@ -0,0 +1,21 @@
+# exists()
```
Alex-Sokolov: title missed

Toilal: Oops

Toilal: Fixed

docs/en/SUMMARY.md
```diff
@@ -63,6 +64,7 @@
     * [setProps](api/wrapper-array/setProps.md)
     * [trigger](api/wrapper-array/trigger.md)
     * [update](api/wrapper-array/update.md)
+    * [visible](api/wrapper/visible.md)
```
eddyerburgh: This should be wrapper-array

docs/en/api/wrapper/visible.md
```diff
@@ -0,0 +1,21 @@
+# visible()
+
+Assert `Wrapper` or `WrapperArray` is visible.
+
+Returns false if an ancestor element has `display: none` or `visibility: hidden` style.
+
```
eddyerburgh, line 6: We should add that this is useful for testing whether v-show is working correctly

docs/en/SUMMARY.md
Toilal: I've made the same as exists() with use the same page.

docs/en/api/wrapper/visible.md
```diff
@@ -0,0 +1,21 @@
...
+- **Example:**
+
+```js
+import { mount } from 'vue-test-utils'
+import { expect } from 'chai'
+import Foo from './Foo.vue'
+
+const wrapper = mount(Foo)
+expect(wrapper.visible()).toBe(true)
+expect(wrapper.find('is-not-visible').visible()).toBe(false)
```
eddyerburgh: is-not-visible should be a class selector, not a tag selector

src/wrappers/error-wrapper.js
```diff
@@ -36,6 +36,10 @@ export default class ErrorWrapper implements BaseWrapper {
     return false
   }
 
+  visible (): boolean {
+    return false
```
eddyerburgh: This should throw an error, like the other methods.

Only exists returns false, everything else throws an error to indicate that find was not succesful

test/unit/specs/mount/Wrapper/visible.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    expect(wrapper.find('.child.ready').visible()).to.equal(false)
+  })
+
+  it('returns true if one element is present', () => {
+    const wrapper = mount(ComponentWithVIf)
+    wrapper.vm.$set(wrapper.vm, 'ready', true)
+    wrapper.update()
+    expect(wrapper.find('.child.ready').visible()).to.equal(true)
+  })
+})
```
eddyerburgh, line 137: Nice tests 👌

docs/en/SUMMARY.md
eddyerburgh: I didn't realize exists uses the same page, this is an error with the exists page.

The wrapper-array pages have different examples that use a wrapper-array

src/wrappers/error-wrapper.js
Toilal: Don't you think it should match the behavior of `exists()` method ? Having this return an error would make assertions on non-visible elements more verbose.

docs/en/api/wrapper/visible.md
Toilal: Done

docs/en/SUMMARY.md
Toilal: Should I fix exists page too ? Or another pull request ?

src/wrappers/error-wrapper.js
eddyerburgh: No, I think it should throw an error. `exists` checks if find is successful, so it makes sense to return false. If find('div') is not successful, exists should return false.

Every other method checks something on the Wrapper element/ instance.

We need to make a distinction between a node being visible and a node being present.

Toilal:
I just squashed everything and make visible() method fails for non-existent elements. I think it's know OK :p

docs/en/api/wrapper/visible.md
```diff
@@ -0,0 +1,23 @@
...
+
+```js
+import { mount } from 'vue-test-utils'
+import { expect } from 'chai'
+import Foo from './Foo.vue'
+
+const wrapper = mount(Foo)
+expect(wrapper.visible()).toBe(true)
+expect(wrapper.find('.is-not-visible').visible()).toBe(false)
+expect(wrapper.findAll('div').visible()).toBe(true)
```
eddyerburgh: Can you remove the findAll examples please

Toilal: Done :)

eddyerburgh (approved):

eddyerburgh:
Thanks @Toilal 🙂

PR #359, opened 2018-01-19
Hotfix: Fix error being thrown with `Vue.extend` when functional

blake-newman:
If using `functional` component with `Vue.extend` an error is thrown,
this is avoided by checking the options object

### DIFF
--- a/test/unit/specs/mount/options/context.spec.js
+++ b/test/unit/specs/mount/options/context.spec.js
@@ -1,7 +1,8 @@
+import Vue from 'vue'
 import { mount } from '~vue-test-utils'
 import { vueVersion } from '~resources/test-utils'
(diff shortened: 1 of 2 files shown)


test/unit/specs/mount/options/context.spec.js
```diff
@@ -1,7 +1,8 @@
+import Vue from 'vue'
 import { mount } from '~vue-test-utils'
 import { vueVersion } from '~resources/test-utils'
 
-describe('context', () => {
+describe.only('context', () => {
```
eddyerburgh: Can you remove this .only 🙂

eddyerburgh:
Thanks, just need the .only removed and I'll merge

PR #370, opened 2018-01-22
Fix code highlight in gitbook.

fimars:
image

😄


fimars:
Maybe should open a PR into master.

eddyerburgh (approved):
Thanks 🙂

fimars:
vue-test-utils is really great, I should thank you for it.

PR #411, opened 2018-02-05
fix: allow plugins to be re-installed in localVue instance (fixes #406)

### LINKED ISSUE #406: Install issue when VueRouter is already used on Vue
### Version
1.0.0-beta.11

### Reproduction link
[https://codesandbox.io/s/p7x5mqkx2q](https://codesandbox.io/s/p7x5mqkx2q)

### Steps to reproduce
This likely applies to other libraries, but when VueRouter is already installed on Vue (i.e `Vue.use(VueRouter)`) and then subsequently used with localVue, it errs with 
```
Error: [vue-router] not installed. Make sure to call `Vue.use(VueRouter)` before creating root instance.
```

### What is expected?
No error.

```
Error: [vue-router] not installed. Make sure to call `Vue.use(VueRouter)` before creating root instance.
```

---
Adding 
```
VueRouter.install.installed = true
```
after declaring a localVue instance fixes the issue.

kellym:
This fixes an issue that occurs when the global Vue object has had a plugin installed to it. 

In this scenario, installing the same plugin into the `localVue` instance created via `createLocalVue()` will mark the plugin as not installed and fails to reinstall it because of an additional check inside the `Vue.use` method.

The core source seems to be the process of cloning the original Vue object, in that plugins that are classes are of type `function` instead of type `object`, so the original objects are passed through.

### DIFF
--- a/src/create-local-vue.js
+++ b/src/create-local-vue.js
@@ -32,6 +32,7 @@ function createLocalVue (): Component {
   instance.options._base = instance
 
   // compat for vue-router < 2.7.1 where it does not allow multiple installs
+  instance._installedPlugins = []
   const use = instance.use
   instance.use = (plugin, ...rest) => {
     if (plugin.installed === true) {
--- a/src/create-local-vue.js
+++ b/src/create-local-vue.js
@@ -32,7 +32,9 @@ function createLocalVue (): Component {
   instance.options._base = instance
 
   // compat for vue-router < 2.7.1 where it does not allow multiple installs
-  instance._installedPlugins = []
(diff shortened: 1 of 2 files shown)


eddyerburgh:
Thanks for the PR 😀

The tests are failing for Vue 2.0.x. Are you able to debug and fix?

kellym:
Hmm, definitely. I'll take a look.

eddyerburgh (approved):

eddyerburgh:
Thanks 😀

kellym:
No, thank _you_. :) This package makes testing a dream compared to what I'm used to.

PR #424, opened 2018-02-11
fix: don't fire event on disabled element

### LINKED ISSUE #395: 'Disabled' attribute on buttons is ignored during testing
### Version
1.0.0-beta.11

### Reproduction link
[https://github.com/jamesoshea/vue-test-utils-button-bug](https://github.com/jamesoshea/vue-test-utils-button-bug)

### Steps to reproduce
Run npm install, then npm test. The test will fail as vue-test-utils trigger a click on a disabled element, and the element's bound method is called anyway.

### What is expected?
The method should not get called when the button is clicked.

When testing a vue component, I will sometimes call .trigger('click') on a wrapper containing just a button. If the button is disabled and a method is bound to its click event, the button is clicked and the method called anyway.

eddyerburgh:
- Fixes https://github.com/vuejs/vue-test-utils/issues/395

### DIFF
--- a/src/wrappers/wrapper.js
+++ b/src/wrappers/wrapper.js
@@ -523,7 +523,7 @@ export default class Wrapper implements BaseWrapper {
     }
 
     // Don't fire event on a disabled element
-    if(this.attributes().disabled) {
+    if (this.attributes().disabled) {
       return
     }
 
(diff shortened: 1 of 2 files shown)


micgro42:
Unfortunately, this solution is too far-reaching. This also disables events, for example, on links with the `disabled` attribute.
While that might be the desired final state, it is something that needs to be asserted by tests. Just plainly disabling the trigger method in that case prevents me from doing so.

PR #428, opened 2018-02-12
Fix: add-attrs and add-listeners correctly restore silent config option

ghost:
add-attrs and add-listeners both incorrectly tried to store the original config, but since they just stored a reference it definitely didn't work.


eddyerburgh (approved):
Good spot, thanks for the PR 😀

Issue #432, opened 2018-02-15
TypeScript: typing error with props and attributes


vwxyutarooo:
### Version
1.0.0-beta.11

### Reproduction link
[https://github.com/vwxyutarooo/vue-test-utils/blob/dev/types/test/wrapper.ts#L19](https://github.com/vwxyutarooo/vue-test-utils/blob/dev/types/test/wrapper.ts#L19)

### Steps to reproduce
Use `wrapper.props().checked` or `wrapper.attributes().foo` in TypeScript.

### What is expected?
`interface BaseWrapper` issue.

### What is actually happening?
```
      TS2339: Property 'checked' does not exist on type 'void | { [name: string]: any; }'.
  Property 'checked' does not exist on type 'void'.
```

---
`void` is unnecessary I suppose. This method always returns any type of data including undefined. Is it my misunderstanding or just bug?

Environments for just in case.
```
- Vue@2.5.13
- @vue/test-utils@1.0.0-beta.11
- jest@22.3.0
- typescript@2.6.2
```

vwxyutarooo:
I'm trying to fix it.

[https://github.com/vwxyutarooo/vue-test-utils/blob/dev/types/index.d.ts#L52](https://github.com/vwxyutarooo/vue-test-utils/blob/dev/types/index.d.ts#L52)

kazupon:
@vwxyutarooo You can send PR :)

vwxyutarooo:
@kazupon Thanks, I've create it!

PR #433, opened 2018-02-15
fix: typing error with props and attributes

vwxyutarooo:
For #432


kazupon (approved):
You're right.
ref: implementation of `props` 
https://github.com/vuejs/vue-test-utils/blob/dev/src/wrappers/wrapper.js#L390

eddyerburgh (approved):
Thanks 😀

PR #434, opened 2018-02-17
[docs]: update how to setup vue-router

ematipico:
I noticed that the instructions to setup the router were not enough and while  I was browsing the issues I found the solution to the problem. So I took the liberty to propose a change to the docs, probably somebody else could bump to the same issue


eddyerburgh (approved):
Thanks 😀

Issue #427, opened 2018-02-12
Testing server side code


wagerfield:
### What problem does this feature solve?

This feature would solve the problem of testing Vue components in a server-side/node environment.

In cases where custom components, directives, mixins etc. need to perform different branches of logic depending on the environment in which they are run (_browser_ or _node_) there is no way of testing the _node_ branches currently.

Vue provides an `$isServer` [instance property](https://vuejs.org/v2/api/#vm-isServer) on each Vue component—this is useful for branching environment dependent logic:

```js
const EnvText = {
  render(createElement) {
    const text = this.$isServer ? 'server' : 'client'
    return createElement('span', text)
  }
}
```

Though the example above is very contrived, there is currently no way of testing the server-side rendering branch of this component (that I know of).

The solution/hack I implemented to work around this problem overrides the `prototype.$isServer` for a `localVue` instance:

```js
const setLocalVueEnv = (localVue, isServer) => {
  Object.defineProperty(localVue.prototype, '$isServer', {
    get: () => isServer
  })
}

test('server env', () => {
  const isServer = true
  const localVue = createLocalVue()

  setLocalVueEnv(localVue, isServer)

  const wrapper = mount(EnvText, { localVue })
  expect(wrapper.vm.$isServer).toBe(isServer)
  expect(wrapper.text()).toBe('server')
})
```

### What does the proposed API look like?
Here are three API proposals I thought might work:

```js
// Option 1: Pass options to createLocalVue
const localVue = createLocalVue({ isServer: true })

// Option 2: Pass a mounting option
const wrapper = mount(EnvText, { isServer: true })

// Option 3: Pass a mock property
const wrapper = mount(EnvText, {
  mocks: {
    $isServer: true
  }
})
```

**NOTE:** Assuming `$isServer` can be mocked/configured on the global Vue instance (I don't think it can) _Option 1 would limit this feature to only working with `localVue` instances_...that might be a good thing though?

eddyerburgh:
There's two problems here. 

First, `mount` and `shallow` create DOM nodes, so they need a DOM environment. Instead, we could create a render function which would be a wrapper around 'vue-server-renderer' `renderToString`. This would return a string of the output, which you could load in cheerio to assert against. 

Second, we should be able to mock `$isServer`. This might require a change to the code in Vue core for us to change the value on each mount/render.

eddyerburgh:
[...]

PR #435, opened 2018-02-19
Add renderToString method

eddyerburgh:
- Add a `renderToString` (https://github.com/vuejs/vue-test-utils/issues/427)


src/renderToString.js
```diff
@@ -0,0 +1,31 @@
...
+import Vue from 'vue'
+import createInstance from './lib/create-instance'
+import './lib/polyfills/object-assign-polyfill'
+import { throwError } from './lib/util'
+
+Vue.config.productionTip = false
+Vue.config.devtools = false
+
+export default function renderToString (component: Component, options: Options = {}): string {
+  const renderer = require('vue-server-renderer').createRenderer()
```
ematipico: Is there a reason why `vue-server-renderer` gets imported using CommonJS and not ESM?

eddyerburgh: ESM imports are hoisted to the top, so any ESM import will always be bundled.

vue-server-renderer relies on node modules, like path, which makes it difficult to compile with webpack. require is only run when the code runs, so putting it inside the renderToString means we only try and import vue-server-renderer when the renderToString function is run.

We should also add a warning that checks if the user is in node, and we can throw an error if they aren't to tell them renderToString needs to be run in node.

ematipico: Clear explanation, thank you very much!

dotnetCarpenter:
I just did a `npm update` and got *npm WARN @vue/test-utils@1.0.0-beta.12 requires a peer of vue-server-renderer@2.x but none is installed. You must install peer dependencies yourself.*.

Could you please update the [release notes](https://github.com/vuejs/vue-test-utils/releases/tag/v1.0.0-beta.12) to explain that we need to install [vue-server-renderer](https://github.com/vuejs/vue/tree/dev/packages/vue-server-renderer#readme) because of the added `renderToString method`.

I did a `git blame` on the `package.json` to see why but I imagine most developers won't go through the trouble.

eddyerburgh:
@dotnetCarpenter thanks for finding this.

I've created an issue to remove it as a peerDependency—https://github.com/vuejs/vue-test-utils/issues/440.

I think most users won't use `renderToString`, so we can handle the dependency ourselves, without throwing a warning for people who won't use the method

dotnetCarpenter:
@eddyerburgh That was not what I intended. If you have an API that has `renderToString` it should work out of the box. Having *vue-server-renderer* as a peer dependency is good as it gives me the choice to install it either as a `devDependecy` or a `dependency`.

I have a *laravel-mix* project that does not use SSR, hence do not need *vue-server-renderer* and I got a nuxt project that already has *vue-server-renderer*. I prefer nuxt to handle versions of *vue-server-renderer* but in my *laravel-mix* project, I have installed it as a `devDependecy` because it is only relevant to my tests.

But if you want to put in the work to handle it in *vue-test-utils*, that is of course fine. But I was not aiming at that. 😄 I was just expecting a better explanation to why I have to install a package, I am not using. I think a small update to the release text will suffice.

PR #439, opened 2018-02-22
Docs: Remove Chai references from examples 

### LINKED ISSUE #437: Documentation: API Chai Examples Incorrect
### Version
1.0.0-beta.12

### Reproduction link
[https://vue-test-utils.vuejs.org/en/api/mount.html](https://vue-test-utils.vuejs.org/en/api/mount.html)

### Steps to reproduce
Example shows

```javascript
import { shallow } from '@vue/test-utils'
import { expect } from 'chai'
import Foo from './Foo.vue'

describe('Foo', () => {
  it('renders a div', () => {
    const wrapper = shallow(Foo)
    expect(wrapper.contains('div')).toBe(true)
  })
})
```

### What is expected?
Based upon Chai API, example would have:

`expect(wrapper.contains('div')).to.be.true`

Example shows

`expect(wrapper.contains('div')).toBe(true)`

This style seems consistent with Jest.

---
This is in a few areas of the API.  Should the documentation show the Chai syntax for expect or the Jest syntax?

dennythecoder:
fixes #437 

Removed most of the references to Chai (sans Karma example page).  Updated the style in a few spots from Chai styling to Jest. For example:

````javascript
// Chai
expect(wrapper.value).to.equal(true)
// Jest
expect(wrapper.value).toBe(true)
````

Excluded the CN docs as there was a PR that looked like it'd conflict if I made the same changes.

### DIFF
--- a/docs/en/api/createLocalVue.md
+++ b/docs/en/api/createLocalVue.md
@@ -11,7 +11,6 @@ Use it with `options.localVue`:
 
 ```js
 import { createLocalVue, shallow } from '@vue/test-utils'
-import { expect } from 'chai'
 import Foo from './Foo.vue'
 
 const localVue = createLocalVue()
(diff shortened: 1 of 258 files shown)


eddyerburgh (approved):
Thanks 😀

Issue #362, opened 2018-01-19
.classes() returns "this.element.className.split is not a function" on svg elements


Epro95Z28:
### Version
1.0.0-beta.2

### Reproduction link
[https://jsfiddle.net/bnLL85ea/1/](https://jsfiddle.net/bnLL85ea/1/)

### Steps to reproduce
-Perform .classes() on an svg image. Example
<svg data-v-35693dba="" version="1.1" role="presentation" width="24" height="24" viewBox="0 0 1792 1792" class="icon fa-icon text-danger" style="font-size: 1.5em;"><path d="M912 1696q0-16-16-16-59 0-101.5-42.5t-42.5-101.5q0-16-16-16t-16 16q0 73 51.5 124.5t124.5 51.5q16 0 16-16zM1728 1408q0 52-38 90t-90 38h-448q0 106-75 181t-181 75-181-75-75-181h-448q-52 0-90-38t-38-90q50-42 91-88t85-119.5 74.5-158.5 50-206 19.5-260q0-152 117-282.5t307-158.5q-8-19-8-39 0-40 28-68t68-28 68 28 28 68q0 20-8 39 190 28 307 158.5t117 282.5q0 139 19.5 260t50 206 74.5 158.5 85 119.5 91 88z"></path> 
    </svg>

### What is expected?
A returned array of strings with all the classes

### What is actually happening?
this.element.className.split is not a function() error

LinusBorg:
* We are at 1.0.0-beta.11 right now, could you verify if that's still an issue in the latest version?
* You write about `classes()` but your code example uses `hasClass()` - which is it?

LinusBorg:
However, I simply think that `<svg>` doesn'T have a className attribute because it doesn't inherit from HTMLElement.

Epro95Z28:
I updated the fiddle. But yes, I am calling .classes
https://jsfiddle.net/bnLL85ea/2/
I looked at -beta.10 and .classes() looks the same as the version I'm running. I will double check though.

Epro95Z28:
Updated to 1.0.0-beta.11 and have the same issue still.

eddyerburgh:
I can't replicate this using the example you posted. How are you running the tests? 

Are you using mocha-webpack? If you're running them in node, are you adding JSDOM properties to the global object?

Epro95Z28:
I will try to recreate this issue in a project. I may not be able to do it until next week.

Epro95Z28:
[...]

PR #445, opened 2018-02-26
fix classes() for svg elements

jeffcunat:
Hello I had a problem using classes function on svg elements (same as #362).
This pull request should fix it as className returns a SVGAnimatedString object for svg elements instead of a simple string.
See notes in [(https://developer.mozilla.org/fr/docs/Web/API/Element/className)]

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@
   "globals": {
     "wrapper": true,
     "expect": true,
-    "Element": true
+    "Element": true,
+    "SVGAnimatedString": true
   },
   "root": true,
   "plugins": [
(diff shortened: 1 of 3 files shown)


eddyerburgh (changes requested):
Thanks for the PR

Can you add a unit test to check this please

jeffcunat:
Hello Edd,
First of all, nice work with vue-test-utils, I use it every day now !
I have modified the unit test for classes to test svg elements. With dev branch it says : 
TypeError: this.element.className.split is not a function
With the PR everything's OK

eddyerburgh (approved):
Thanks for adding the test 😀

PR #447, opened 2018-02-27
Monorepo (WIP)

eddyerburgh:
- Split into test-utils, server-test-utils, create-instance repos

TODO

- Fix broken tests
- Improve build/ test pipeline
- Move scripts to package root

New structure:

- packages
  - server-test-utils (published as `@vue/server-test-utils`)
  - test-utils (published as `@vue/test-utils`)
  - create-instance (private)
  - shared (private)


38elements:
@eddyerburgh 
`test/specs/mounting-options/slots.spec.js`  is not tested.

PR #461, opened 2018-03-09
Refactor: simplify code of removeDuplicateNodes method

lusarz:


packages/test-utils/src/find-vnodes.js
```diff
@@ -24,14 +24,7 @@ function findAllVNodes (vnode: VNode, nodes: Array<VNode> = []): Array<VNode> {
...
 function removeDuplicateNodes (vNodes: Array<VNode>): Array<VNode> {
-  const uniqueNodes = []
-  vNodes.forEach((vNode) => {
-    const exists = uniqueNodes.some(node => vNode.elm === node.elm)
-    if (!exists) {
-      uniqueNodes.push(vNode)
-    }
-  })
-  return uniqueNodes
+  return vNodes.filter((vNode, index, array) => index === array.findIndex(node => vNode.elm === node.elm))
```
eddyerburgh: I think this could be simplified further:
```js
return vNodes.filter((vNode, index) => index === vNodes.findIndex(node => vNode.elm === node.elm))
```

lusarz: Probably you're right, I'm gonna update :)

lusarz: Done!

PR #462, opened 2018-03-09
refactor: small optimization in checking for component with error during mount

lusarz:


packages/test-utils/src/mount.js
```diff
@@ -29,10 +29,10 @@ export default function mount (component: Component, options: Options = {}): Vue
     vm.$mount()
   }
 
-  const componentsWithError = findAllVueComponentsFromVm(vm).filter(c => c._error)
+  const componentWithError = findAllVueComponentsFromVm(vm).find(c => c._error)
```
lusarz, line 32: We care only for first error, so there is no need to search for all.

eddyerburgh (approved):
Thanks 👍

PR #466, opened 2018-03-13
DOCS: replace hasProp() to props() in docs examples

iztsv:
Hey!

I've noticed that `hasProp` is deprecated:

```
console.error node_modules/@vue/test-utils/dist/vue-test-utils.js:15
      [vue-test-utils]: hasProp() has been deprecated and will be removed in version 1.0.0. Use props() instead—https://vue-test-utils.vuejs.org/en/api/wrapper/props
```

So I replaced `hasProps` to `props` in the docs.

### DIFF
--- a/docs/en/api/mount.md
+++ b/docs/en/api/mount.md
@@ -42,7 +42,7 @@ describe('Foo', () => {
         color: 'red'
       }
     })
-    expect(wrapper.hasProp('color', 'red')).toBe(true)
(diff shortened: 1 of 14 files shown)


eddyerburgh (approved):
Great spot. Thanks for the changes 😀

iztsv:
@eddyerburgh Thanks for the reply. No problem 😀

PR #469, opened 2018-03-14
docs: remove deprecated hasStyle method

iztsv:
Hello!

As I can see:

> hasStyle() has been deprecated and will be removed in version 1.0.0. Use wrapper.element.style instead

English documentation contain information about `hasStyle` method on `options.html` page only:

> ` This can be used with [`hasStyle`](wrapper/hasStyle.md) to check multi element CSS selectors.`

and `hasStyle` link locates to `404`.

So, I've removed info about `hasStyle` method from docs.


eddyerburgh (approved):
Thanks 😀

PR #477, opened 2018-03-18
feat(render): add render method

eddyerburgh:
- Add render method in vue/server-test-utils that returns cheerio wrapper


38elements:
@eddyerburgh 
I think it is necessary to add the documentation about this render method.

eddyerburgh:
Yes agreed. We need to add docs for render and rendertostring.

38elements:
I think that the api document for this render method is also necessary.
https://github.com/vuejs/vue-test-utils/tree/dcc3167aa1123e102b910b7d101baf642a09716b/docs/en/api

PR #485, opened 2018-03-23
docs: add render method to docs

eddyerburgh:


docs/en/api/README.md
```diff
@@ -2,6 +2,7 @@
 
 * [mount](./mount.md)
 * [shallow](./shallow.md)
+* [render](api/render.md)
```
38elements: I think it is necessary to change `api/render.md` to `./render.md`.

docs/en/api/render.md
```diff
@@ -0,0 +1,103 @@
...
+
+- **Options:**
+
+See [options](./options.md)
+
+- **Usage:**
+
+Renders an object to a string and returns a [cheerio wrapper](https://github.com/cheeriojs/cheerio).
+
+Cheerio is a node implementation of the jQuery library. It has a similar API to the Vue Test Utils [`Wrapper`](wrapper/README.md).
```
38elements: IMHO, Since I think node is Node.js, I think this is better.
Cheerio is a library like jQuery to handle DOM in Node.js.

eddyerburgh: Thanks, how about:

`Cheerio is an implementation of core jQuery designed for Node.js.` 
or 
`Cheerio is jQuery-like library to traverse the DOM in Node.js`

dennythecoder: > Cheerio is an implementation of core jQuery designed for Node.js.

has my vote.

38elements: Since I think that the words which is `core jQuery` is not easy to understand,
the below sentence is better.
> Cheerio is jQuery-like library to traverse the DOM in Node.js

eddyerburgh: I've gone with:

>Cheerio is jQuery-like library to traverse the DOM in Node.js

because it's easier to understand

docs/en/api/renderToString.md
```diff
@@ -25,10 +25,12 @@ Renders a component to HTML.
 
 `renderToString` uses [`vue-server-renderer`](https://ssr.vuejs.org/en/basic.html) under the hood, to render a component to HTML.
 
+`renderToString` is exported from the `@vue/server-test-utils` package.
```
38elements: IMHO, Since the code is `import { renderToString } from '@vue/server-test-utils'`, I think this is better.
You import `renderToString` from the `@vue/server-test-utils` package.

docs/en/api/render.md
38elements: IMHO, there is the blank lines.
I think it is better that the blank lines removed and 2 space is added to end of line. 

```
Renders an object to a string and returns a [cheerio wrapper](https://github.com/cheeriojs/cheerio).  
Cheerio is a node implementation of the jQuery library. It has a similar API to the Vue Test Utils [`Wrapper`](wrapper/README.md).  
`render` uses [`vue-server-renderer`](https://ssr.vuejs.org/en/basic.html) under the hood, to render a component to static HTML.  
`render` is exported from the `@vue/server-test-utils` package.
```

Renders an object to a string and returns a [cheerio wrapper](https://github.com/cheeriojs/cheerio).  
Cheerio is a node implementation of the jQuery library. It has a similar API to the Vue Test Utils [`Wrapper`](wrapper/README.md).  
`render` uses [`vue-server-renderer`](https://ssr.vuejs.org/en/basic.html) under the hood, to render a component to static HTML.  
`render` is exported from the `@vue/server-test-utils` package.

docs/en/api/renderToString.md
eddyerburgh: That sounds strange to me. How about:

> `render` is part of the `@vue/server-test-utils` package.

or

>`render` is included in the `@vue/server-test-utils` package.

38elements: IMHO,
I think that "`render` is included in the `@vue/server-test-utils` package." is better,
since this is more appropriate, when this is translated into Japanese.

PR #491, opened 2018-03-27
Expose fundamental type to support typecasting

DrSensor:
Example use case:
```ts
const nameInputList = wrapper.findAll({ name: 'my-text-field' }).wrappers.find(cmp => cmp.props().label === 'Name') as Wrapper<VTextField>

expect(nameInputList.exists()).toBe(true)
```
then warning `[ts] Object is possibly 'undefined'.` can be avoided and autocomplete work as expected

### DIFF
--- a/packages/test-utils/types/index.d.ts
+++ b/packages/test-utils/types/index.d.ts
@@ -70,54 +70,52 @@ interface BaseWrapper {
   destroy (): void
 }
 
-declare global {
-  interface Wrapper<V extends Vue> extends BaseWrapper {
-    readonly vm: V
-    readonly element: HTMLElement
-    readonly options: WrapperOptions
-  
-    find<R extends Vue> (selector: VueClass<R>): Wrapper<R>
-    find<R extends Vue> (selector: ComponentOptions<R>): Wrapper<R>
-    find (selector: FunctionalComponentOptions): Wrapper<Vue>
-    find (selector: string): Wrapper<Vue>
-    find (selector: RefSelector): Wrapper<Vue>
-  
-    findAll<R extends Vue> (selector: VueClass<R>): WrapperArray<R>
-    findAll<R extends Vue> (selector: ComponentOptions<R>): WrapperArray<R>
-    findAll (selector: FunctionalComponentOptions): WrapperArray<Vue>
-    findAll (selector: string): WrapperArray<Vue>
-    findAll (selector: RefSelector): WrapperArray<Vue>
-  
-    html (): string
-    text (): string
-    name (): string
-  
-    emitted (event?: string): { [name: string]: Array<Array<any>> }
-    emittedByOrder (): Array<{ name: string, args: Array<any> }>
-  }
-  
-  interface WrapperArray<V extends Vue> extends BaseWrapper {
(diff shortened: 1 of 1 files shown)


ktsn:
I don't think it is a good idea to expose them in global namespace. If you just want to ignore `undefined` type, you should use non-null assertion operator. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

```ts
const nameInputList = wrapper
  .findAll({ name: 'my-text-field' })
  .wrappers
  .find(cmp => cmp.props().label === 'Name')!

expect(nameInputList.exists()).toBe(true)
```

DrSensor:
Thanks, it works for me :raised_hands: . There is something that I need to address before closing/continue this PR

- What if some project apply [no-non-null-assertion](https://palantir.github.io/tslint/rules/no-non-null-assertion/) rule? (related to [strictNullChecks](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html))
- I notice that declaring on the global scope can make conflicts with libs that have the same type (or maybe it just overloads type which is good). How about wrapping it on `namespace` or just `export` it?
- Is there some possibility of a class member in `Wrapper<T>` that return type `T`?

ktsn:
> What if some project apply no-non-null-assertion rule? (related to strictNullChecks)

Then, you may want to add null check or disable the rule in the test case? IMO, the intention of the rule is to ensure the devs always check `null` or `undefined` if it can be. So, casting to remove `undefined` type should be restricted too if your team disallow non null assertion. Of course I understand it is a bit verbose in the test, so I guess disabling the rule is the best choice.

> How about wrapping it on namespace or just export it?

I guess exporting these types is ok.

> Is there some possibility of a class member in Wrapper<T> that return type T?

Sorry, I don't understand about that. Could you describe it more detail?

DrSensor:
> Then, you may want to add null check or disable the rule in the test case?

Well seems like there is no choice, if that happened to me when onboarding project,
I will just use `// @ts-ignore` :smiley: 

> I guess exporting these types is ok.

:ok_hand: 

> Is there some possibility of a class member in Wrapper that return type T?

I usually do typecasting like `this.$refs.upload as MyComponent` to access computed/method.
After reading `index.d.ts` again seems like they are some similar case.

```ts
const fupload = wrapper.findAll({ name: 'file-upload' }).wrappers
// I bet this will produce a type error with `mycomputedproperties` as unknown properties
const a = fupload.vm.mycomputedproperties

// by typecasting it will not produce type error
const button = wrapper.findAll({ name: 'file-upload' }).At(0) as Wrapper<FileUpload>
const a = fupload.vm.mycomputedproperties
```

ktsn:
Actually, you can pass your component type as type parameter in that case.

```ts
const fupload = wrapper.findAll<FileUpload>({ name: 'file-upload' }).wrappers
```

In any cases, I think we can export `Wrapper` and `WrapperArray` for convenience. Can you update the `index.d.ts` to do that? 🙂

DrSensor:
Sure

ktsn (approved):

PR #493, opened 2018-03-28
[docs][en] Replace from mapCoverage to collectCoverage in Jest sample.

potato4d:
The mapCoverage option is already deprecated.
If you move this sample, an error will be displayed and it will not work.
The error is displayed as "Option 'mapCoverage' has been removed, as it no longer necessary."

### DIFF
--- a/docs/en/guides/testing-SFCs-with-jest.md
+++ b/docs/en/guides/testing-SFCs-with-jest.md
@@ -159,7 +159,7 @@ Jest recommends creating a `__tests__` directory right next to the code being te
 
 Jest can be used to generate coverage reports in multiple formats. The following is a simple example to get started with:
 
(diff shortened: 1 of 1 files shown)


eddyerburgh (approved):
Thanks 😀

eddyerburgh (changes requested):
Can you remove `collectCoverage` from the config example. We already have it in the following example

potato4d:
@eddyerburgh 
Which example should I delete?
collectCoverage still seems necessary.

docs/en/guides/testing-SFCs-with-jest.md
```diff
@@ -49,7 +49,7 @@ Next, create a `jest` block in `package.json`:
       // process `*.vue` files with `vue-jest`
       ".*\\.(vue)$": "<rootDir>/node_modules/vue-jest"
     },
-    "mapCoverage": true
+    "collectCoverage": true
```
eddyerburgh: Can you remove collectCoverage from here, since the previous example wasn't collecting coverage

potato4d: @eddyerburgh Oh, OK!

potato4d:
@eddyerburgh fix it.

eddyerburgh (approved):
Thanks :)

PR #506, opened 2018-03-31
docs: make it clear that  mocha is a test runner

### LINKED ISSUE #467: Docs - the note about Jest's limitations is unclear
### Version
1.0.0-beta.12

### Reproduction link
[https://vue-test-utils.vuejs.org/en/guides/testing-SFCs-with-jest.html](https://vue-test-utils.vuejs.org/en/guides/testing-SFCs-with-jest.html)

### Steps to reproduce
Read the section of the docs called "Processing SFCs in Jest", especially the note

### What is expected?
I should not be confused 

I am confused, in particular about what the limitations of Jest actually are.

Especially
"To use them, read the guide on testing SFCs with Mocha   webpack."

---
In that sentence I assume "them" is the unsupported features, but I'm not sure. Also.. these features aren't mentioned on the linked page. I think it should say "you will have to use Mocha instead of Jest", But maybe it should say  "you will have to use Mocha in addition to Jest". The provided link is more confusing than helpful to me.

Having no background in javascript testing and how they interface with code etc I do not understand the scope of these limitations or the real effects. of using Jest over Mocha. Adding a sentence or two about the impact of these features would be really helpful

eddyerburgh:
Closes #467

### DIFF
--- a/docs/en/guides/testing-SFCs-with-jest.md
+++ b/docs/en/guides/testing-SFCs-with-jest.md
@@ -53,7 +53,7 @@ Next, create a `jest` block in `package.json`:
 }
 ```
 
-> **Note:** `vue-jest` currently does not support all the features of `vue-loader`, for example custom block support and style loading. In addition, some webpack-specific features such as code-splitting are not supported either. To use them, read the guide on [testing SFCs with Mocha + webpack](./testing-SFCs-with-mocha-webpack.md).
+> **Note:** `vue-jest` currently does not support all the features of `vue-loader`, for example custom block support and style loading. In addition, some webpack-specific features such as code-splitting are not supported either. To use unsupported features, you can use Mocha to run your tests, and webpack  to compile your components. To get started,  read the guide on [testing SFCs with Mocha + webpack](./testing-SFCs-with-mocha-webpack.md).
 
 ## Handling webpack Aliases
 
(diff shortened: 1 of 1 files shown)


docs/en/guides/testing-SFCs-with-jest.md
```diff
@@ -53,7 +53,7 @@ Next, create a `jest` block in `package.json`:
 }
 ```
 
-> **Note:** `vue-jest` currently does not support all the features of `vue-loader`, for example custom block support and style loading. In addition, some webpack-specific features such as code-splitting are not supported either. To use them, read the guide on [testing SFCs with Mocha + webpack](./testing-SFCs-with-mocha-webpack.md).
+> **Note:** `vue-jest` currently does not support all the features of `vue-loader`, for example custom block support and style loading. In addition, some webpack-specific features such as code-splitting are not supported either. To use unsupported features, you can use Mocha to run your tests, and webpack  to compile your components. To get started,  read the guide on [testing SFCs with Mocha + webpack](./testing-SFCs-with-mocha-webpack.md).
```
38elements: IMHO,
Since it is explicit not to use Jest, I think This is better.

To use unsupported features, you need to use Mocha instead of Jest to run your tests, and webpack to compile your components. To get started,

Issue #156, opened 2017-11-06
Test slot scope


Akryum:
We should add a way to test the scoped slots. For example, `<slot v-for="..." :item="item" :index="index" />` ➡️ We should be able to test the values of  `item` and `index` in the slot.

LinusBorg:
ok so there's two problems:

1. creating a proper scopeSlot Function
2. adding the scoped slots content to the vm.

### 1. creating a proper scopeSlot Function

For normal slots, vue-test-utils uses `vue-template-compiler`, which still have some limitations.

My suggestion would be to use a real Vue instance instead to create the slot content:
```javascript
function createSlotContent ({
  slotValue: string,
  slotName: string = 'default',
  scoped: Boolean = false
}) : Function | Array<VNode> {
  const slotParent = new Vue({
    components: {
      slotter: { template: `<div></div>`},
    }
    template: `<div><slotter ref="slotter">${slotValue}</slotter></div>`,
  })
  return slotParent.[scoped ? '$scopedSlots' : '$slots'][slotName]
}
```

### 2. adding the scoped slots content to the vm.

For normal slots, we simply do `vm.$slots = slotContent`, but that doesn't work for $scopedSlots because of [this line](https://github.com/vuejs/vue/blob/956756b1be7084daf8b6afb92ac0da7c24cde2a5/src/core/instance/render.js#L74) in the `vm._render` function:
```javascript
if (_parentVnode) {
      vm.$scopedSlots = _parentVnode.data.scopedSlots || emptyObject
    }
```
This code will run everytime we (re-)render the vm, and since there isn't an actual parentVnode to read scoped slots from, we would have to mock it.

That must be tested extensively, I guess.

eddyerburgh:
Adding a parent node would solve another problem we're having with `$attrs`. I think this could be a good solution.

I'll look into it.

LinusBorg:
Great! I'll also take a look between the holidays if I have an hour or two.

jorySsense:
Is there anything I can do to help with this?

38elements:
I try to resolve this issue.
I will send pull request within 3 days.

eddyerburgh:
@38elements That would be great! We were thinking about creating a parent instance inside the create-instance method, like  @LinusBorg suggested.

38elements:
@eddyerburgh 
Since I am unsure of whether my implementation method is correct,
I sent a pull request of incomplete code.
It is #507 . 
[...]

PR #507, opened 2018-04-01
feat: add scoped slots option

38elements:
This is related to #156 .
This adds `scopedSlots` option by replacing the [renderSlot](https://github.com/vuejs/vue/blob/5a9da95b8a865416f082952a48416ffc091e4078/src/core/instance/render-helpers/render-slot.js#L8) method.
It is supported at vue@2.5+.
It does not support `template` tag.

### DIFF
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -73,7 +73,8 @@ Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#hea
 
 - type: `{ [name: string]: string }`
 
-Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.  
+Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
+
 There is two limitations.  
 
 * This supports vue@2.5+.
@@ -85,10 +86,10 @@ Example:
 ```js
 const wrapper = shallow(Component, {
   scopedSlots: {
-    bar: '<p slot-scope="props">{{props.index}},{{props.text}}</p>'
+    foo: '<p slot-scope="props">{{props.index}},{{props.text}}</p>'
   }
 })
-expect(wrapper.html()).toBe('<div><p>0,text1</p><p>1,text2</p><p>2,text3</p></div>')
+expect(wrapper.find('#fooWrapper').html()).toBe('<div id="fooWrapper"><p>0,text1</p><p>1,text2</p><p>2,text3</p></div>')
 ```
 
 ### `stubs`
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -6,6 +6,7 @@ Options for `mount` and `shallow`. The options object can contain both Vue Test 
 
 - [`context`](#context)
 - [`slots`](#slots)
+- [`scopedSlots`](#scopedslots)
 - [`stubs`](#stubs)
 - [`mocks`](#mocks)
 - [`localVue`](#localvue)
@@ -67,6 +68,28 @@ There is a limitation to this.
 
 This does not support PhantomJS.  
 Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).
+
+### `scopedSlots`
+
+- type: `{ [name: string]: string }`
+
+Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.  
+There is two limitations.  
+
+* This supports vue@2.5+.
+
+* You can not set a `template` tag to top of `scopedSlots` option.
+
+Example:
+
+```js
+const wrapper = shallow(Component, {
+  scopedSlots: {
+    bar: '<p slot="item" slot-scope="props">{{props.index}},{{props.text}}</p>'
+  }
+})
+expect(wrapper.html()).toBe('<div><p>0,text1</p><p>1,text2</p><p>2,text3</p></div>')
+```
 
 ### `stubs`
 
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -85,7 +85,7 @@ Example:
 ```js
 const wrapper = shallow(Component, {
   scopedSlots: {
-    bar: '<p slot="item" slot-scope="props">{{props.index}},{{props.text}}</p>'
+    bar: '<p slot-scope="props">{{props.index}},{{props.text}}</p>'
   }
 })
 expect(wrapper.html()).toBe('<div><p>0,text1</p><p>1,text2</p><p>2,text3</p></div>')
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -75,11 +75,11 @@ Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#hea
 
 Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
 
-There is two limitations.  
+There are two limitations.  
 
-* This supports vue@2.5+.
+* This option is only supported in vue@2.5+.
 
-* You can not set a `template` tag to top of `scopedSlots` option.
+* You can not use `<template>` tag as the root element in the `scopedSlots` option.
 
 Example:
 
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -75,11 +75,14 @@ Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#hea
 
 Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
 
-There are two limitations.
+There are three limitations.
 
 * This option is only supported in vue@2.5+.
 
 * You can not use `<template>` tag as the root element in the `scopedSlots` option.
+
+* This does not support PhantomJS.  
+Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).
 
 Example:
 
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -75,7 +75,7 @@ Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#hea
 
 Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
 
-There are two limitations.  
+There are two limitations.
 
 * This option is only supported in vue@2.5+.
 
--- a/docs/en/api/options.md
+++ b/docs/en/api/options.md
@@ -67,7 +67,7 @@ You can pass text to `slots`.
 There is a limitation to this.
 
 This does not support PhantomJS.  
-Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).
+You can use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer) as an alternative.
 
 ### `scopedSlots`
 
@@ -82,7 +82,7 @@ There are three limitations.
 * You can not use `<template>` tag as the root element in the `scopedSlots` option.
 
 * This does not support PhantomJS.  
-Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).
+You can use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer) as an alternative.
 
 Example:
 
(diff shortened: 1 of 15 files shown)


test/specs/mounting-options/scopedSlots.js
```diff
@@ -0,0 +1,40 @@
+import { describeWithShallowAndMount, vueVersion } from '~resources/utils'
+import ComponentWithScopedSlots from '~resources/components/component-with-scoped-slots.vue'
+
+describeWithShallowAndMount('scopedSlots', (mountingMethod) => {
+  if (vueVersion >= 2.5) {
+    it('mounts component scoped slots', () => {
+      const wrapper = mountingMethod(ComponentWithScopedSlots, {
+        scopedSlots: {
+          'item': '<p slot="item" slot-scope="props">{{props.index}},{{props.text}}</p>'
```
eddyerburgh: Do we need to include slot=item in the markup? We already identify it as the item slot in the scopedSlots object

packages/test-utils/types/index.d.ts
```diff
@@ -113,6 +113,7 @@ interface MountOptions<V extends Vue> extends ComponentOptions<V> {
   localVue?: typeof Vue
   mocks?: object
   slots?: Slots
+  scopedSlots?: string
   stubs?: Stubs,
```
eddyerburgh: This should be an object. Can you add a test for this in types/test?

test/specs/mounting-options/scopedSlots.js
```diff
@@ -0,0 +1,40 @@
+import { describeWithShallowAndMount, vueVersion } from '~resources/utils'
+import ComponentWithScopedSlots from '~resources/components/component-with-scoped-slots.vue'
+
+describeWithShallowAndMount('scopedSlots', (mountingMethod) => {
+  if (vueVersion >= 2.5) {
+    it('mounts component scoped slots', () => {
```
eddyerburgh: We have an itIf helper function in resouces/utils. Could you use that instead of if else statement

eddyerburgh (changes requested):
Looks good! I don't like monkey patching the _t helper but we can look at changing that in the future. For now we can go ahead with this. I've added some comments

38elements:
Thank you for reviewing.
I changed this.

38elements:
Thank you for reviewing.
I changed this.

packages/test-utils/types/index.d.ts
```diff
@@ -113,6 +113,7 @@ interface MountOptions<V extends Vue> extends ComponentOptions<V> {
   localVue?: typeof Vue
   mocks?: object
   slots?: Slots
+  scopedSlots?: object
```
ktsn: We can specify more detailed type `Record<string, string>`

38elements: Thank you for pointing out.
I changed it.

docs/en/api/options.md
```diff
@@ -68,6 +69,29 @@ There is a limitation to this.
 This does not support PhantomJS.  
 Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).
 
+### `scopedSlots`
+
+- type: `{ [name: string]: string }`
+
+Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
+
+There is two limitations.
```
eddyerburgh: There is two limitations > `There are two limitations. `

docs/en/api/options.md
```diff
@@ -68,6 +69,29 @@ There is a limitation to this.
...
 
+### `scopedSlots`
+
+- type: `{ [name: string]: string }`
+
+Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
+
+There is two limitations.  
+
+* This supports vue@2.5+.
```
eddyerburgh: > This option is only supported in vue@2.5+

docs/en/api/options.md
```diff
@@ -68,6 +69,29 @@ There is a limitation to this.
...
+
+* You can not set a `template` tag to top of `scopedSlots` option.
```
eddyerburgh: > You cannot use `<template>` tag  as  the root element  in the`scopedSlots` option.

packages/create-instance/add-scoped-slots.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import { compileToFunctions } from 'vue-template-compiler'
+import { throwError } from 'shared/util'
+
+export function addScopedSlots (vm: Component, scopedSlots: Object): void {
+  Object.keys(scopedSlots).forEach((key) => {
+    const template = scopedSlots[key].trim()
+    if (template.substr(0, 9) === '<template') {
+      throwError('scopedSlots option does not support template tag.')
```
eddyerburgh: Can the error message read:

> the scopedSlots option does not support a <template> tag as the root element.

packages/create-instance/create-instance.js
```diff
@@ -57,6 +59,28 @@ export default function createInstance (
...
 
+  if (options.scopedSlots) {
+    const vueVersion = Number(`${Vue.version.split('.')[0]}.${Vue.version.split('.')[1]}`)
+    if (vueVersion >= 2.5) {
+      vm.$_vueTestUtils_scopedSlots = {}
+      const renderSlot = vm._renderProxy._t
+      vm._renderProxy._t = function (name, feedback, props, bindObject) {
+        const scopedSlotFn = vm.$_vueTestUtils_scopedSlots[name]
+        if (scopedSlotFn) {
+          props = { ...bindObject, ...props }
```
eddyerburgh: Why do we need to set vm._renderProxy.props?

38elements: `scopedSlotFn` needs `vm._renderProxy.props`.
If `vm._renderProxy.prop` is not setted, the render function can not use `props`.

```
with(this){return _c('p',{},[_v(_s(props.index)+","+_s(props.text))])}
```

38elements:
Thank you for pointing out.
I changed them.

packages/create-instance/create-instance.js
38elements: There is a bug.
When `slot-scope` is other than `props`, this is not work.

38elements:
There is a bug.
When `slot-scope` is other than `props`, this is not work.
I will fix it.

packages/create-instance/create-instance.js
```diff
@@ -60,20 +60,33 @@ export default function createInstance (
...
           props = { ...bindObject, ...props }
-          vm._renderProxy.props = props
-          return scopedSlotFn.call(vm._renderProxy)
+          const proxy = {}
+          Object.keys(vm._renderProxy).concat(Object.keys(Vue.prototype)).forEach((key) => {
+            if (key[0] === '_') {
+              proxy[key] = vm._renderProxy[key]
+            }
+          })
+          proxy[slotScope] = props
```
38elements: Since when `slot name`  and  `slot scope` is same,  `vm._renderProxy.[<slot name>]` is changed in `_t` function, `proxy` is created.

```
<ul>
  <slot name="foo"
    v-for="(item, index) in foo"
    :text="item.text"
    :index="index">
  </slot>
</ul>
```

```
function render() {
  with(this) {
    return _c('ul', [_l((foo), function (item, index) {
      return _t("foo", null, {
        text: item.text,
        index: index
      })
    })], 2)
  }
}
```

scopedSlotFn

```
function anonymous() {
  with(this){return _c('p',{},[_v(_s(foo.index)+","+_s(foo.text))])}
}
```

Mattchewone:
Looking forward to this being merged.

Thanks for this.

packages/create-instance/create-instance.js
```diff
@@ -75,10 +75,9 @@ export default function createInstance (
         if (scopedSlotFn) {
           props = { ...bindObject, ...props }
           const proxy = {}
-          Object.keys(vm._renderProxy).concat(Object.keys(Vue.prototype)).forEach((key) => {
-            if (key[0] === '_') {
-              proxy[key] = vm._renderProxy[key]
-            }
+          const helpers = ['_c', '_o', '_n', '_s', '_l', '_t', '_q', '_i', '_m', '_f', '_k', '_b', '_v', '_e', '_u', '_g']
```
38elements: `Object.keys(vm._renderProxy).concat(Object.keys(Vue.prototype)).forEach((key) => {` may be better,
since it is abstruct.

packages/create-instance/add-scoped-slots.js
```diff
@@ -0,0 +1,17 @@
...
+import { throwError } from 'shared/util'
+
+export function addScopedSlots (vm: Component, scopedSlots: Object): void {
+  Object.keys(scopedSlots).forEach((key) => {
+    const template = scopedSlots[key].trim()
+    if (template.substr(0, 9) === '<template') {
+      throwError('the scopedSlots option does not support a template tag as the root element.')
+    }
+    const domParser = new window.DOMParser()
+    const document = domParser.parseFromString(template, 'text/html')
```
eddyerburgh: Can you rename this variable to avoid collisions with `window.document`?

packages/create-instance/create-instance.js
```diff
@@ -57,6 +59,40 @@ export default function createInstance (
...
+        const scopedSlotFn = vm.$_vueTestUtils_scopedSlots[name]
+        const slotScope = vm.$_vueTestUtils_slotScopes[name]
+        if (scopedSlotFn) {
+          props = { ...bindObject, ...props }
+          const proxy = {}
+          const helpers = ['_c', '_o', '_n', '_s', '_l', '_t', '_q', '_i', '_m', '_f', '_k', '_b', '_v', '_e', '_u', '_g']
+          helpers.forEach((key) => {
+            proxy[key] = vm._renderProxy[key]
+          })
+          proxy[slotScope] = props
```
eddyerburgh, line 82: Does Vue throw an error if there are two conflicting slot scopes? If not, we should throw an error here

eddyerburgh (changes requested):
Looking good, this adds some much needed support 👍 . I've added a couple of comments

packages/create-instance/add-scoped-slots.js
38elements: I changed it.

packages/create-instance/create-instance.js
38elements: It is impossible to set a word that starts with "_" in Vue.

```
[Vue warn]: Property or method "_foo" is not defined on the instance but referenced during render. 
Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. 
See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.
```

eddyerburgh: I meant if you mounted with this option:
```
{
scopedSlots: {
  foo: '<div />',
  foo: '<input />'
}
}
```

The second scopedSlot would overwrite the first scopedSlots in `_vueTestUtils_slotScopes`. Although thinking about it, this is probably fine.

38elements: I will add a exception and document.

38elements: It is not possible to register the same key.

```
{ foo: 1, foo: 2 }
// => { foo: 2 }
```

eddyerburgh: 🤦‍♂️ Of course ;p

docs/en/api/options.md
```diff
@@ -68,6 +69,32 @@ There is a limitation to this.
...
+Provide an object of scoped slots contents to the component. The key corresponds to the slot name. The value can be a template string.
+
+There are three limitations.
+
+* This option is only supported in vue@2.5+.
+
+* You can not use `<template>` tag as the root element in the `scopedSlots` option.
+
+* This does not support PhantomJS.  
+Please use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).
```
eddyerburgh: Can you change to:
> You can use [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer) as an alternative.

38elements:
Thank you for pointing out.
I changed it.

packages/create-instance/create-instance.js
```diff
@@ -61,7 +61,7 @@ export default function createInstance (
 
   if (options.scopedSlots) {
     if (window.navigator.userAgent.match(/PhantomJS/i)) {
-      throwError('the scopedSlots option does not support strings in PhantomJS. Please use Puppeteer, or pass a component.')
+      throwError('the scopedSlots option does not support in PhantomJS. Please use Puppeteer, or pass a component.')
```
eddyerburgh: > the scopedSlots option does not support in PhantomJS. Please use Puppeteer, or pass a component.
to 

> the scopedSlots option does not support PhantomJS. Please use Puppeteer, or pass a component.

38elements: I changed it.

eddyerburgh (approved):
Great job 👍

38elements:
@eddyerburgh 
Thank you for reviewing.

Haroenv:
@38elements, what was the reason `<template>` in scoped slot was not implemented?

Haroenv:
@38elements, what was the reason `<template>` in scoped slot was not implemented?

Issue #473, opened 2018-03-16
Provide an API to get wrapped refs


LinusBorg:
### What problem does this feature solve?
we currently have `wrapper.find()` which can be used to select an element by a CSS selector, or find a component instance by name/constructor. Find returns a Wrapper for the item that was found.

We don't have a similar way of using a ref. We *can* do:

```javascript
wrapper.vm.$refs.myElement
```

But that will give use the plain element, not a wrapped version, so we can't use the Wrapper's helpers on this element.

The same is true for component refs, of course

### What does the proposed API look like?
```javascript
wrapper.ref('refName') // => Returns instance of Wrapper

// alternatively we could stay true to the Vue syntax here, if technically possible:
wrapper.$refs.refName // => Returns instance of Wrapper
```

eddyerburgh:
We actually do support ref selectors:
```js
const buttonWrapper = wrapper.find({ ref: 'myButton' })
buttonWrapper.trigger('click')
```

It was discussed in this thread—https://github.com/vuejs/vue-test-utils/issues/67

We should probably improve the docs to improve visibility.

LinusBorg:
Oh, great!

dennythecoder:
I really like @LinusBorg's suggestion here:

> // alternatively we could stay true to the Vue syntax here, if technically possible:
wrapper.$refs.refName // => Returns instance of Wrapper

Doesn't appear to be difficult to implement, but I think it'd be nice to pull one or more wrappers from the usage of `wrapper.$refs`.  Curious about @matt-oconnell's thoughts

eddyerburgh:
I think we should keep the find and findAll methods for traversing the rendered nodes. Since we already support a ref selector in find, I'm closing this issue.

PR #511, opened 2018-04-02
docs: extend examples in find api (#473)

igeligel:
Regarding #473 i updated the documentation of the `.find(selector)` function. Whether the proposed API will be implemented, I guess the `.find(selector)` function stays the same more or less and documentation about getting a component should include all possibilities.

I included examples of every selector possible so people might see it. I also included some whitespace so the different types of selections are separated.

Thanks for the great project 👍

### DIFF
--- a/docs/en/api/wrapper/find.md
+++ b/docs/en/api/wrapper/find.md
@@ -27,8 +27,8 @@ expect(bar.is(Bar)).toBe(true)
 const barByName = wrapper.find({ name: 'bar' })
 expect(barByName.is(Bar)).toBe(true)
 
-const barRef = wrapper.find({ ref: 'bar' })
-expect(barRef.is(Bar)).toBe(true)
+const fooRef = wrapper.find({ ref: 'foo' })
+expect(fooRef.is(Foo)).toBe(true)
 ```
 
 - **See also:** [Wrapper](README.md)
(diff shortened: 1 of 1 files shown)


38elements:
IMHO,
Since it seems that `name` option and `ref` option are the same,
I think that it is difficult to understand the difference between `name` option  and `ref` option.

docs/en/api/wrapper/find.md
```diff
@@ -17,10 +17,18 @@ import Foo from './Foo.vue'
...
 const div = wrapper.find('div')
 expect(div.is('div')).toBe(true)
+
 const bar = wrapper.find(Bar)
 expect(bar.is(Bar)).toBe(true)
+
+const barByName = wrapper.find({ name: 'bar' })
+expect(barByName.is(Bar)).toBe(true)
+
+const barRef = wrapper.find({ ref: 'bar' })
```
eddyerburgh: Can you  change this to a different value, like  `'foo'`

eddyerburgh (changes requested):

docs/en/api/wrapper/find.md
igeligel: I renamed this now to `foo` but I guess it is somehow confusing.

eddyerburgh (approved):
Thanks 😀

PR #513, opened 2018-04-06
Reproduction: find stubbed children

lusarz:
Good morning,

I'm wondering how should I test such a scenario. I want to stub some component, but also want to check if this component exists.

### DIFF
--- a/test/specs/mounting-options/stubs.spec.js
+++ b/test/specs/mounting-options/stubs.spec.js
@@ -2,13 +2,18 @@ import ComponentWithChild from '~resources/components/component-with-child.vue'
 import ComponentWithNestedChildren from '~resources/components/component-with-nested-children.vue'
 import Component from '~resources/components/component.vue'
 import ComponentAsAClass from '~resources/components/component-as-a-class.vue'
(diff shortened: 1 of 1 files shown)


eddyerburgh:
This should be a StackOverflow question.

You can pass an object with a custom component stub:

```js
const SomeStub = {
  template: '<div />'
}
const wrapper = mount(ComponentWithChild, { stubs: {
    'child-component' : SomeStub
  }})
expect(wrapper.contains(Component)).to.be.true
```

lusarz:
Thanks @eddyerburgh  for response, but it doesn't solve my issue. In that way Component is not stub, but is rendered.

In similar scenario, when I use `shallow` instead of `mount`, I'm able to use `expect(wrapper.contains(Component)).to.be.true`. In my case I don't want to use shallow, because I want other children to be rendered, but wanted to not render one specific child, because it contains some logic I don't want to be calculated during tests.

This pr was more proposition to implement that question, sorry I didn't create issue for that, but thought that I'll get quicker response by pr ;)

eddyerburgh:
I've updated my previous response, that should do what you want

PR #517, opened 2018-04-07
refactor: small refactoring of validators

lusarz:


packages/shared/validators.js
```diff
@@ -53,51 +45,22 @@ export function componentNeedsCompiling (component: Component) {
     !component.functional
 }
 
-export function isValidSelector (selector: any) {
```
lusarz, line 56: This method was not used anywhere

eddyerburgh (approved):
Thanks 👍

PR #520, opened 2018-04-08
Refactor: create isString and isFunction methods

lusarz:


packages/shared/stub-components.js
```diff
@@ -5,17 +5,14 @@ import { compileToFunctions } from 'vue-template-compiler'
...
 function isVueComponent (comp) {
   return comp && (comp.render || comp.template || comp.options)
 }
 
 function isValidStub (stub: any) {
-  return !!stub &&
-      typeof stub === 'string' ||
-      (stub === true) ||
-      (isVueComponent(stub))
+  return isString(stub) || (stub === true) || isVueComponent(stub)
```
lusarz, line 15: There is no need to check for `!!stub`, becase `isString(null)` returns false, because `typeof null !== 'string'`

lusarz:
Hmm, seems like it's not possible because of flow checks.

PR #522, opened 2018-04-10
fix(types): update packages/test-utils/types/index.d.ts

38elements:
This adds the types for [sync](https://github.com/vuejs/vue-test-utils/blob/74a27ae0eb7be0d9705a0ae9106e1942e758e9ea/docs/en/api/options.md#sync) and [name option](https://github.com/vuejs/vue-test-utils/blob/74a27ae0eb7be0d9705a0ae9106e1942e758e9ea/docs/en/api/selectors.md#name).


eddyerburgh (approved):
Thanks @38elements  :)

PR #529, opened 2018-04-14
chore: update version

38elements:


eddyerburgh (approved):
We need a script to release while we're still in beta. Once we've released we can just run lerna publish, and it will manage the versioning for us. Unfortunately lerna doesn't support non semantic versioning (so we can't publish 1.0.0-beta.x with lerna). That means I need to manually release, and I keep forgetting to update these versions 😕

38elements:
You can check the version this page.
https://github.com/vuejs/vue-test-utils/search?utf8=%E2%9C%93&q=1.0.0-beta&type=

Issue #325, opened 2018-01-03
Pass mocks object to global config


lmiller1990:
Hi,

I have a feature request/idea.

I work in a non English speaking country, so we are using `vue-i18n` to create a bilingual app. It adds a global `$t` function. I have to do:

```js
mocks: {
  $t: jest.fn()
}
```

in every test, or the test fails with `$t is undefined`, since almost all components have some form of text.

Should this be something handled by a test running? Not sure - if not, maybe we can have some way to set some default options for all tests, or all `shallow` renders or something.

eddyerburgh:
We already have a global config to add stubs. 

I think it would be useful for people to have a global config for mocks too. The local mounting options would overwrite the mocks set in the config if they clash.

Would you like to make a PR to implement this?

lmiller1990:
I would like to make it! Do you have a good place I can look to get started?

Sorry the slow reply.

Edit: got something working using a similar way to config stubs. But it would be nice to have a config file (.js) we could load options from?

eddyerburgh:
Do you want to make a PR with the original way, using the config object.

We should start another discussion about whether to support a config file or not.

lmiller1990:
It seems like you can do the same thing at `stubs` in the current `master`, simply `mocks['myMock'] = jest.fn()` for example.

Maybe we should discuss a config file first before going any further.

dobromir-hristov:
Exact same issue as @lmiller1990. Also things like ElementUI's $alert,$confirm or own global helpers we have can be easily mocked everywhere :)

abhipanda:
Yes please - big problem :(

lmiller1990:
A simple work around you can use in the meantime, before we solve the global config problem, is to make a helper with all your commonly used mocks, for example:

```js
// Let's call it mocks.js
import { shallow } from '@vue/test-utils'

const mocks = {
  $t: () => {} // I always mock vuei18n
}

const shallowWithMocks = (component, otherData, otherMocks) => shallow(component, { 
  mocks: {
    ..mocks, ...otherMocks
  },
  ...otherData 
})

export shallowWithMocks
```

```js
import { shallowWithMocks } from './mocks.js'

const wrapper = shallowWithMocks(component, { 
  propsData: { /* */ }
})
```

Note: I didn't test this, but I have something similar in a project somewhere. It probably has some typos. 

______

[...]

PR #531, opened 2018-04-15
Allow mocks to be set in config object

lmiller1990:
First try at #325 .

I am sure this is not the best way to write the test, though. 

Initially I was using `$store` instead of `$t`, which caused a test related to installing vuex on `createLocalVue` to fail. I wonder why?

### DIFF
--- a/docs/en/api/config.md
+++ b/docs/en/api/config.md
@@ -58,5 +58,5 @@ import VueTestUtils from '@vue/test-utils'
 
 VueTestUtils.config.methods['errors'] = () => {
   any: () => false
-}  
+}
 ```
--- a/docs/en/api/config.md
+++ b/docs/en/api/config.md
@@ -43,3 +43,20 @@ VueTestUtils.config.mocks['$store'] = {
   }
 }
 ```
+
+### `methods`
+
+- type: `Object`
+- default: `{}`
+
+You can configure default methods using the `config` object. This can be useful for plugins that inject methods to components, like [VeeValidate](https://vee-validate.logaretm.com/). You can override methods set in `config` by passing `methods` in the mounting options.
+
+Example:
+
+```js
+import VueTestUtils from '@vue/test-utils'
+
+VueTestUtils.config.methods['errors'] = () => {
+  any: () => false
+}  
+```
--- a/docs/en/api/config.md
+++ b/docs/en/api/config.md
@@ -37,7 +37,7 @@ Example:
 ```js
 import VueTestUtils from '@vue/test-utils'
 
-VueTestUtils.config.stubs['$store'] = {
+VueTestUtils.config.mocks['$store'] = {
   state: {
     id: 1
   }
--- a/docs/en/api/config.md
+++ b/docs/en/api/config.md
@@ -24,3 +24,22 @@ import VueTestUtils from '@vue/test-utils'
 
 VueTestUtils.config.stubs['my-component'] = '<div />'
 ```
+
+### `mocks`
+
+- type: `Object`
+- default: `{}`
+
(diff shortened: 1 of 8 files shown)


packages/shared/merge-options.js
```diff
@@ -16,12 +16,29 @@ function getStubs (optionStubs, config) {
   }
 }
 
+function getMocks (optionMocks, config) {
```
eddyerburgh: This is nearly the same as the getStubs function, can you split it into a generic getOptions function?

lmiller1990: Sure. Maybe we can use it to provide an interface to all options.

lmiller1990: Ok! done.

- `methods` also works in the same way, now that we have a generic `getOptions`. Just need to add it to `mergeOptions`. Should we add that in this commit? It's only a few lines.
- `computed` might need a bit more work, since it needs a `getter` and `setter` to be set up. It can't just be added in the same way as `methods` and `mocks`.
- I did not try `props` and `data` yet.

eddyerburgh: Great work! Yes, please add methods :)

docs/en/api/config.md
```diff
@@ -24,3 +24,22 @@ import VueTestUtils from '@vue/test-utils'
 
 VueTestUtils.config.stubs['my-component'] = '<div />'
 ```
+
+### `mocks`
+
```
38elements, line 29: I think that it is necessary to add links for this section to following pages.
https://github.com/vuejs/vue-test-utils/blob/dev/docs/en/api/README.md
https://github.com/vuejs/vue-test-utils/blob/dev/docs/en/README.md
https://github.com/vuejs/vue-test-utils/blob/dev/docs/en/SUMMARY.md

lmiller1990: Yep, for sure. Will do. Thanks.

lmiller1990:
I added `methods` as well. I initially thought we should allow for all options, but some don't really make that much sense (maybe)?

The ones I can think of use cases for:

- [x] `methods`
- [x] `mocks`
- [x] `stubs`

May or may not make sense:

- [ ] `data` (can accomplish the same thing with `config.mocks`, even if it's not exactly the same)
- [ ] `propsData` (would you want a default prop..?)

Can't think of a use case for:
- [ ] `computed`
- [ ] `slots` 
- [ ] `localVue`

What does everyone think? I'm pretty happy wit the current state, it seems there was a shared interest for `mocks` mainly, but I'm sure other people have different use cases that might call for the others (not sure what, though).

docs/en/api/config.md
lmiller1990: Edit: actually, I don't think we need to. This doesn't add a new API, just build on the existing `config` page.

eddyerburgh:
I think methods, mocks, and stubs is good for now 👍

eddyerburgh (changes requested):
Can you add some tests in tests/specs/mounting-options/methods.spec.js and mocks.spec.js to make sure the options take priority over config version, similar to this test: https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/mounting-options/stubs.spec.js#L226

lmiller1990:
Sure. 
- [x] mocks https://github.com/vuejs/vue-test-utils/pull/531/commits/7cd229ce0bdbd97956d1e9f088578d4806286b74
- [x] methods https://github.com/vuejs/vue-test-utils/pull/531/commits/31d1f7a6817474ace65bf915a72db630f78e3922

Done!

eddyerburgh (approved):
Great work. I'm going to open a PR to fix config for server-test-utils

PR #535, opened 2018-04-17
Refactor: create isString and isFunction methods

lusarz:


38elements:
IMHO, I think replacing `typeof foo === string` to `isString(foo)` and `typeof foo === function` to `isFunction(foo)` does not have benefit.
`vuejs/vue` uses `typeof`.
https://github.com/vuejs/vue/search?p=2&q=typeof&type=&utf8=%E2%9C%93

lusarz:
For me it looks a bit more aesthetically. On the other hand `vuejs/vue` have similar methods, for example [isTrue](https://github.com/lusarz/vue/blob/f7b5c86ef2620c1d7ec275f86ae0f380ccaea4ef/src/shared/util.js#L15-L17).

38elements:
It is subjective whether it is stylish or not.
I did not mentioned `isTrue`.
I mentioned `isString` and `isFunction`.

lusarz:
Yeah, I agree it's subjective, so it's only proposition to merge. 

I know you didn't mentioned `isTrue`, but if replacing `foo === true` to `isTrue(foo)` is beneficial, why don't replace `typeof foo === string` to `isString(foo)` ?

38elements:
IMHO, I think replacing `typeof foo === string` to `isString(foo)` and `typeof foo === function` to `isFunction(foo)` does not have benefit.

eddyerburgh:
I agree, I don't think we should refactor simple typeof checks into their own functions. It's a common way for people to check the type and is easy to understand

PR #536, opened 2018-04-17
Refactor: refactor shared/util

lusarz:


packages/shared/util.js
```diff
@@ -8,16 +8,14 @@ export function warn (msg: string) {
...
+export function camelize (str: string): string {
+  return str.replace(/-(\w)/g, (_, c) => c ? c.toUpperCase() : '')
+}
 
-/**
- * Capitalize a string.
- */
-export const capitalize = (str: string) => str.charAt(0).toUpperCase() + str.slice(1)
+export function capitalize (str: string): string {
+  return str.charAt(0).toUpperCase() + str.slice(1)
```
lmiller1990, line 16: I think it's easier to read if the regex are declared outside (before refactor style).

I like the new `camelize` - but it feels like it 's more consistent with the other two helpers to do it the same way.

Is there some benefits to the refactor, or just for style?

eddyerburgh:
Thanks @lusarz for the PR.

These methods are copied directly from Vue source code, and I don't think there's a benefit to removing the regex. I'm going to close this and the previous PR. 

If you would like to do some refactoring the [add-slots file](https://github.com/vuejs/vue-test-utils/blob/dev/packages/create-instance/add-slots.js), and [stub-components file](https://github.com/vuejs/vue-test-utils/blob/dev/packages/shared/stub-components.js) need to be refactored!

lusarz:
@eddyerburgh ok, I understand. In accordance with your suggestion I did small refactoring of stub-components file :)

PR #538, opened 2018-04-18
refactor: remove unnecessary processing

38elements:
This removes unnecessary creating array and a loop.


eddyerburgh (approved):
Thanks 👍

PR #541, opened 2018-04-19
refactor: refactor stub-components.js

lusarz:
Pull request contains changes like
- move public (exported) methods to top of `stub-components.js` file
- separate stub options validation from creation of stubs
- clean a few repeated code
- overall cleaning

### DIFF
--- a/packages/shared/stub-components.js
+++ b/packages/shared/stub-components.js
@@ -2,27 +2,62 @@
 
 import Vue from 'vue'
 import { compileToFunctions } from 'vue-template-compiler'
-import { throwError } from './util'
-import { componentNeedsCompiling } from './validators'
-import { compileTemplate } from './compile-template'
-import { capitalize, camelize, hyphenate } from './util'
+import { validateStubOptions } from 'shared/stub-components-validation'
+import { componentNeedsCompiling } from 'shared/validators'
+import { compileTemplate } from 'shared/compile-template'
 
-function isVueComponent (comp) {
-  return comp && (comp.render || comp.template || comp.options)
+export function createComponentStubsForGlobals (vueInstance: Component) {
+  return createBlankStubs(vueInstance.options.components)
 }
 
-function isValidStub (stub: any) {
-  return !!stub &&
-      typeof stub === 'string' ||
-      (stub === true) ||
-      (isVueComponent(stub))
+export function createComponentStubsForAll (component: Component) {
+  return createComponentStubsRecursively(component)
 }
 
-function isRequiredComponent (name) {
-  return name === 'KeepAlive' || name === 'Transition' || name === 'TransitionGroup'
+export function createComponentStubs (components: Object = {}, stubOptions: Object = {}): Object {
+  validateStubOptions(stubOptions, components)
+  return Array.isArray(stubOptions)
+    ? createBlankStubs(stubOptions)
+    : createStubsFromObject(stubOptions, components)
 }
 
-function getCoreProperties (component: Component): Object {
+function createComponentStubsRecursively (component: Component, stubs: Object = {}): Object {
+  if (!component) return stubs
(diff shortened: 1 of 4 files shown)


packages/test-utils/src/shallow.js
```diff
@@ -26,15 +26,11 @@ export default function shallow (
...
-  const stubbedComponents = createComponentStubsForAll(component)
-  const stubbedGlobalComponents = createComponentStubsForGlobals(vue)
-
   return mount(component, {
     ...options,
     components: {
-      // stubbed components are used instead of original components components
-      ...stubbedGlobalComponents,
-      ...stubbedComponents
+      ...createComponentStubsForGlobals(vue),
```
lusarz, line 32: `stubbedComponents` and `stubbedGlobalComponents` was unnecessary variables, I invoked functions directly

packages/shared/stub-components.js
```diff
@@ -37,150 +72,15 @@ function getCoreProperties (component: Component): Object {
...
-
-  // Loop through extended component chains to stub all child components
-  while (extended) {
-    if (extended.components) {
-      stubComponents(extended.components, stubbedComponents)
-    }
-    extended = extended.extends
+function addToIgnoredElements (name) {
+  // ignoreElements does not exist in Vue 2.0.x
+  if (Vue.config.ignoredElements) {
```
lusarz, line 93: These code was duplicated, so I moved it into function

packages/shared/stub-components.js
```diff
@@ -37,150 +72,15 @@ function getCoreProperties (component: Component): Object {
...
-  const stubbedComponents = {}
-
-  if (component.components) {
-    stubComponents(component.components, stubbedComponents)
-  }
-
-  let extended = component.extends
-
-  // Loop through extended component chains to stub all child components
-  while (extended) {
```
lusarz, line 160: Moved this logic into `createComponentStubsRecursively`

38elements:
This looks like only changing style.
Would you tell me the benefits of this?

lusarz:
Yes, main purpose of refactoring is changing style and improving the design of an existing code.

I think after these changes code would be easier to maintain and it would be easier to find bugs and introduce extensions.

38elements:
Thank you for replying.

There is no objective basis for its benefit.

IMHO,
This is cosmetic change.
I think this change makes difficult to `git blame`. 

This makes difficult to understand code for me.
For example, the `createBlankStub` function helps to understanding code for me,
this function is removed.

lusarz:
True, I agree with `git blame` issue. Main reason of that is that I moved public (exported) functions to top of file, what I think is good practice.

I replaced `createBlankStub` with `createStub` method, which returns 'blank' stub by default. But yeah, I agree it was no perfect solution. I updated it in [second commit](https://github.com/vuejs/vue-test-utils/pull/541/commits/25c1adc7e5cdfd800cb0cb3df14ae8d1443be94c). What do you think about that?

38elements:
I am sorry.
My expression may have been ambiguous.
It is important whether there is benefit or not.
IMHO,
Since this change has no benefit, I think this change is not necessary.
Other people may not think so.

lusarz:
Reducing number of lines of code is beneficial itself

lusarz:
I'm not yet happy with everything, so I'm gonna close this pull request for now

eddyerburgh:
@38elements this part of the codebase is in need of refactoring, so I think a cosmetic change is fine. If we reduce the lines significantly and keep the code readable, that's a good benefit

38elements:
@eddyerburgh 
~~I think that a change which is consistent with the style of existing code is fine.
I think that it is fine if the readability is improved explicitly.~~

38elements:
@eddyerburgh 
Since you are a maintainer, 
I think that the code you are easy to maintain is correct.
Would you please do not mind my opinion.

PR #542, opened 2018-04-19
refactor: run renderToString tests in node environment

eddyerburgh:


test/specs/mounting-options/slots.spec.js
```diff
@@ -9,15 +9,17 @@ import {
   itDoNotRunIf
 } from '~resources/utils'
 
-describeWithMountingMethods('options.slots', (mountingMethod) => {
+describeWithMountingMethods.only('options.slots', (mountingMethod) => {
```
38elements: I think it is necessary to remove `.only`.

PR #544, opened 2018-04-20
Refactor: refactor stub-components.js

lusarz:
Pull request contains changes like
- separate stub options validation from creation of stubs
- overall cleaning

### DIFF
--- a/packages/shared/stub-components-validate.js
+++ b/packages/shared/stub-components-validate.js
@@ -1,0 +1,44 @@
+// @flow
+
+import { throwError } from './util'
+import { compileToFunctions } from 'vue-template-compiler'
+
+export function validateStubOptions (stubOptions: Array<string> | Object) {
+  if (Array.isArray(stubOptions)) {
+    if (containsNonStringItem(stubOptions)) {
+      throwError('each item in an options.stubs array must be a string')
+    }
+  } else {
+    if (containsInvalidOptions(stubOptions)) {
+      throwError('options.stub values must be passed a string or component')
+    }
+
+    if (necessaryCompileToFunctionsMissed(stubOptions)) {
+      throwError('vueTemplateCompiler is undefined, you must pass components explicitly if vue-template-compiler is undefined')
+    }
+  }
+}
+
+function containsNonStringItem (array: Array<string>): boolean {
+  return array.some(name => typeof name !== 'string')
+}
+
+function necessaryCompileToFunctionsMissed (stubOptions: Object): boolean {
+  return Object.keys(stubOptions)
+    .map(key => stubOptions[key])
+    .some(stub => typeof stub === 'string') && !compileToFunctions
+}
+
+function containsInvalidOptions (stubOptions: Object): boolean {
+  return Object.keys(stubOptions)
+    .map(key => stubOptions[key])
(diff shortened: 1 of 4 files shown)


packages/shared/stub-components.js
```diff
@@ -40,147 +25,72 @@ function getCoreProperties (component: Component): Object {
...
+      return object
+    }, {})
   }
+  return stubOptions
+}
 
-  if (component.extendOptions && component.extendOptions.components) {
-    stubComponents(component.extendOptions.components, stubbedComponents)
-  }
+export function createComponentStubs (components: Object = {}, stubOptions: Object): Object {
```
lusarz, line 70: Personally I think public methods should be moved to top of file. Didn't do it to satisfy `git blame` as much as possible

packages/shared/stub-components.js
```diff
@@ -40,147 +25,72 @@ function getCoreProperties (component: Component): Object {
...
+    .filter(name => options[name] !== false)
+    .reduce((stubs, name) => {
+      stubs[name] = createStub(components[name], options[name])
 
-    components[c] = createBlankStub(instance.options.components[c])
-    delete instance.options.components[c]._Ctor // eslint-disable-line no-param-reassign
-    delete components[c]._Ctor // eslint-disable-line no-param-reassign
-  })
-  return components
+      if (Vue.config.ignoredElements) {
```
lusarz: Do we need it ? I copied it from original source.

eddyerburgh: I don't think we do. you can remove it :)

packages/shared/stub-components.js
eddyerburgh: I don't think we should move the position of functions. Most files in this project have public functions at the bottom of the files

packages/shared/stub-components-validate.js
```diff
@@ -0,0 +1,58 @@
...
+  } else {
+    if (containsInvalidOptions(stubOptions)) {
+      throwError('options.stub values must be passed a string or component')
+    }
+
+    if (necessaryCompileToFunctionsMissed(stubOptions)) {
+      throwError('vueTemplateCompiler is undefined, you must pass components explicitly if vue-template-compiler is undefined')
+    }
+
+    if (containsTemplateWithCircularReference(stubOptions, components)) {
```
eddyerburgh: This is actually causing bugs for some users. The check is not strict enough, so I've removed the check in the source code for the moment. Could you either remove this code, or improve the code to handle edge cases like: `UnreadState: '<div data-stub-unread-state v-bind="$props"><slot/></div>'`

packages/shared/stub-components.js
lusarz: Ok, let it stay like that

packages/shared/stub-components.js
lusarz: Removed :)

packages/shared/stub-components-validate.js
lusarz: Removed :)

eddyerburgh (approved):

eddyerburgh:
Thanks :)

Issue #546, opened 2018-04-20
slot-scope doesn't accept all valid expressions


smiller171:
### Version
dev branch

### Reproduction link
[https://codesandbox.io/s/mow3jyo89j?module=%2Ftest%2Funit%2Fspecs%2FScoped-broken.spec.js&view=preview](https://codesandbox.io/s/mow3jyo89j?module=%2Ftest%2Funit%2Fspecs%2FScoped-broken.spec.js&view=preview)

### Steps to reproduce
Access a scope with a JS expression like `"{ item }"` instead of `"item"`

### What is expected?
Using `"{ item }"` to identify a scope should allow for accessing that variable directly instead of having to use `something.item`

### What is actually happening?
```
item is not defined

ReferenceError: item is not defined
    at Object.eval (eval at createFunction (https://mow3jyo89j.codesandbox.io/node_modules/vue-template-compiler/browser.js), :3:39)
    at Proxy.vm._renderProxy._t (https://mow3jyo89j.codesandbox.io/node_modules/@millergeek/test-utils/dist/vue-test-utils.js:1931:31)
    at eval (https://mow3jyo89j.codesandbox.io/src/components/Scoped.vue?f8cd7dd6:1:334)
    at Proxy.renderList (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:3707:16)
    at Proxy.render (https://mow3jyo89j.codesandbox.io/src/components/Scoped.vue?f8cd7dd6:1:273)
    at VueComponent.Vue._render (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:4546:22)
    at VueComponent.updateComponent (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:2790:21)
    at Watcher.get (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:3144:25)
    at new Watcher (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:3133:12)
    at mountComponent (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:2797:3)
    at VueComponent.Vue.$mount (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:8542:10)
    at VueComponent.Vue.$mount (https://mow3jyo89j.codesandbox.io/node_modules/vue/dist/vue.common.js:10941:16)
[...]

PR #547, opened 2018-04-21
feat: destructuring slot scope

38elements:
This adds the function of [destructuring slot scope](https://vuejs.org/v2/guide/components-slots.html#Destructuring-slot-scope).
This is related to #546 .


eddyerburgh (approved):
Thanks :)

PR #548, opened 2018-04-22
docs: add Peer Dependency documentation

FeliciousX:
Currently, you must have `vue-template-compiler` installed separately because it's a peer dependency but it is not documented.


eddyerburgh (approved):
Thanks :)

PR #553, opened 2018-04-23
fix: remove circular dependency checks

eddyerburgh:
The checks currently fail for lots of edge cases:

```
stubs: {
  UnreadState: '<div data-stub-unread-state v-bind="$props"><slot/></div>',
  UnreadState: '<div>unread-state</div>',
}
```

This removes checks for circular dependencies when using the stubs option. This was simple error handling, but I believe the logic to implement it effectively is substantial. For the moment, we should remove the check

### DIFF
--- a/test/specs/mounting-options/stubs.spec.js
+++ b/test/specs/mounting-options/stubs.spec.js
@@ -299,9 +299,9 @@ describeWithMountingMethods('options.stub', (mountingMethod) => {
     expect(HTML).contains('No render function')
   })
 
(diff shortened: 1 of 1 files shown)


Akryum:
@eddyerburgh I'm not sure the PR removes the checks in the code itself (you seem to only have changed the tests).

eddyerburgh:
Ah sorry, I was in a rush. 

This has been removed in another PR, but it won't be released for another week or so (until beta.16).

PR #555, opened 2018-04-24
feat: Adds `provide` to the global config.

jeremyzahner:
While working on our current project, we noticed that mocking global provides (usually provided by the root Vue instance) is quite painful because it needs to be added to every single mount()/shallow() in the tests. 

This implements the config abilities that the mocks, stubs and methods also have.

The only feature that we can not easily support is the merging of functions (obviously). We did implement an error to signal this very clearly.

### DIFF
--- a/test/specs/mounting-options/provide.spec.js
+++ b/test/specs/mounting-options/provide.spec.js
@@ -1,3 +1,4 @@
+import { config } from '~vue/test-utils'
 import ComponentWithInject from '~resources/components/component-with-inject.vue'
 import { injectSupported } from '~resources/utils'
 import {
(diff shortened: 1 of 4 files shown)


test/specs/mounting-options/provide.spec.js
```diff
@@ -44,4 +56,38 @@ describeWithMountingMethods('options.provide', (mountingMethod) => {
 
       expect(wrapper.vm.setInBeforeCreate).to.equal('created')
     })
+
+  it('injects the provide from the config', () => {
```
eddyerburgh: You can use the `itDoNotRunIf` and `injectSupported` functions to only run for Vue versions which support inject/provide. Example—https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/mounting-options/provide.spec.js#L9

eddyerburgh (changes requested):
Looks great, all we need is to skip the tests in Vue < 2.2 and I'm happy to merge

test/specs/mounting-options/provide.spec.js
jeremyzahner: @eddyerburgh Done in 0ee6ce8f29c0af45ccc8475b994e28b6dedaf84a. Did amend the existing commit.

jeremyzahner:
@eddyerburgh Ready to merge.

eddyerburgh (approved):
Great job :)

PR #556, opened 2018-04-24
Refactor: refactor add-slots.js

lusarz:


packages/create-instance/add-slots.js
```diff
@@ -1,60 +1,39 @@
...
-      throwError('the slots option does not support strings in PhantomJS. Please use Puppeteer, or pass a component.')
-    }
-    const domParser = new window.DOMParser()
-    const _document = domParser.parseFromString(slotValue, 'text/html')
-    const _slotValue = slotValue.trim()
-    if (_slotValue[0] === '<' && _slotValue[_slotValue.length - 1] === '>' && _document.body.childElementCount === 1) {
-      elem = vm.$createElement(compileToFunctions(slotValue))
-    } else {
-      const compiledResult = compileToFunctions(`<div>${slotValue}{{ }}</div>`)
-      const _staticRenderFns = vm._renderProxy.$options.staticRenderFns
```
lusarz, line 26: I removed this 'replacing' of `staticRenderFns ` and didn't experienced unit-test failures - is it ok to remove, or I miss something ?

packages/shared/util.js
```diff
@@ -21,3 +21,11 @@ export const capitalize = (str: string) => str.charAt(0).toUpperCase() + str.sli
  */
 const hyphenateRE = /\B([A-Z])/g
 export const hyphenate = (str: string) => str.replace(hyphenateRE, '-$1').toLowerCase()
+
+export function toArray (value: any) {
+  return Array.isArray(value) ? value : [value]
+}
+
+export function isObject (obj: mixed): boolean %checks {
```
lusarz, line 29: This method is [copied from Vue](https://github.com/vuejs/vue/blob/dev/src/shared/util.js#L41-L43)

packages/create-instance/add-slots.js
eddyerburgh: I don't think it's OK to remove, but we should add some tests to cover it. Can you write a test to check that a static slot is added correctly.

lusarz: Thanks @eddyerburgh for information - I'll add appropriate tests for that

lusarz: I moved back `staticRenderFns` related code in pull request. To be honest I don't know what is static slot - I'm not yet familiar with everything related with Vue.

eddyerburgh (approved):

Issue #530, opened 2018-04-14
Add setChecked and setValue methods


eddyerburgh:
### What problem does this feature solve?
Changing the value so that the v-model is updated can be confusing:

```js
const input = wrapper.find('input[type="text"]')
input.element.value = 'some value' // input element value is changed, v-model is not
input.trigger('input') // v-model updated
```

```js
const radioInput = wrapper.find('input[type="radio"]')
radioInput.element.checked = true // input element value is changed, v-model is not
radioInput.trigger('input') // v-model not updated
radioInput.trigger('change') // v-model updated
```

### What does the proposed API look like?
```js
const input = wrapper.find('input[type="text"]')
input.setValue('some value')
const radioInput = wrapper.find('input[type="radio"]')
radioInput.setChecked()
const option = wrapper.find('option')
option.setSelected()
```

styro0:
Hi, you should take a look to [mwangaben-vthelpers](https://github.com/mwangaben/mwangaben-vthelpers) which provides some helpers for vue-test-utils like b.type(text, selector) :

```
type(text, input) {
  let node = this.find(input)
  node.element.value = text
  node.trigger('input')
}
```

14nrv:
Hi, you should take a look to [mwangaben-vthelpers](https://github.com/mwangaben/mwangaben-vthelpers) which provides some helpers for vue-test-utils like b.type(text, selector) :

```
type(text, input) {
  let node = this.find(input)
  node.element.value = text
  node.trigger('input')
}
```

38elements:
IMHO,
Since `type` reminds `keydown-event`,  I think that `type` is not appropriate.

beyer-martin:
@eddyerburgh are you working on this? If not I believe I can do it

eddyerburgh:
@beyersito I'm not currently working on it, it would be great if you could :)

beyer-martin:
@eddyerburgh Will do then, I'll start with setValue

beyer-martin:
I've started with setValue, I would be great if you could check if I started correctly.

I have a few doubts: 

1. setValue could work on the following inputs: text like inputs (text, email, ...), select, checkbox. The triggered event will follow v-model source.

2. setChecked could work on radios and in checkboxes receiving a boolean parameter.

3. Should it validate that the wrapper is compatible with the method?

4. What to do if the wrapper is a component

Sorry to bother this much

Thanks!

eddyerburgh:
[...]

PR #557, opened 2018-04-24
feat: Add setValue method

beyer-martin:
#530 
First contribution, wanted to know if I am following the correct path.
Its not finished yet.

Thanks

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -546,11 +546,11 @@ export default class Wrapper implements BaseWrapper {
     const event = 'input'
 
     if (tag === 'SELECT') {
-      throwError('wrapper.setValue() cannot be called on select')
+      throwError('wrapper.setValue() cannot be called on a <select> element. Use wrapper.setSelected() instead')
     } else if (tag === 'INPUT' && type === 'checkbox') {
-      throwError('wrapper.setValue() cannot be called on checkbox. Use wrapper.setChecked() instead')
+      throwError('wrapper.setValue() cannot be called on a <input type="checkbox" /> element. Use wrapper.setChecked() instead')
     } else if (tag === 'INPUT' && type === 'radio') {
-      throwError('wrapper.setValue() cannot be called on radio. Use wrapper.setChecked() instead')
+      throwError('wrapper.setValue() cannot be called on a <input type="radio" /> element. Use wrapper.setChecked() instead')
     } else if (tag === 'INPUT' || tag === 'textarea') {
       // $FlowIgnore
       el.value = value
@@ -582,10 +582,8 @@ export default class Wrapper implements BaseWrapper {
     const type = this.attributes().type
     const event = 'change'
 
-    if (this.isVueComponent) {
-      throwError('wrapper.setChecked() cannot be called on component. Use wrapper.setValue() instead')
-    } else if (tag === 'SELECT') {
-      throwError('wrapper.setChecked() cannot be called on select')
+    if (tag === 'SELECT') {
+      throwError('wrapper.setChecked() cannot be called on a <select> element. Use wrapper.setSelected() instead')
     } else if (tag === 'INPUT' && type === 'checkbox') {
       // $FlowIgnore
       if (el.checked !== checked) {
@@ -594,7 +592,7 @@ export default class Wrapper implements BaseWrapper {
       }
     } else if (tag === 'INPUT' && type === 'radio') {
       if (!checked) {
-        throwError('wrapper.setChecked() cannot be called with parameter false on radio')
+        throwError('wrapper.setChecked() cannot be called with parameter false on a <input type="radio" /> element.')
       } else {
         // $FlowIgnore
         if (!el.checked) {
@@ -631,9 +629,9 @@ export default class Wrapper implements BaseWrapper {
     } else if (tag === 'SELECT') {
       throwError('wrapper.setSelected() cannot be called on select. Call it on one of its options')
     } else if (tag === 'INPUT' && type === 'checkbox') {
-      throwError('wrapper.setSelected() cannot be called on checkbox. Use wrapper.setChecked() instead')
+      throwError('wrapper.setSelected() cannot be called on a <input type="checkbox" /> element. Use wrapper.setChecked() instead')
     } else if (tag === 'INPUT' && type === 'radio') {
-      throwError('wrapper.setSelected() cannot be called on radio. Use wrapper.setChecked() instead')
+      throwError('wrapper.setSelected() cannot be called on a <input type="radio" /> element. Use wrapper.setChecked() instead')
     } else if (tag === 'INPUT' || tag === 'textarea') {
       throwError('wrapper.setSelected() cannot be called on "text" inputs. Use wrapper.setValue() instead')
     } else {
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -559,6 +559,7 @@ export default class Wrapper implements BaseWrapper {
       throwError('wrapper.setValue() cannot be called on this element')
     }
 
+    // $FlowIgnore
     el.value = value
     this.trigger(event)
   }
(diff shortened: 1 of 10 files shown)


eddyerburgh (changes requested):
Looks great so far :)

Things we need:

- setValue added to flow/wrapper.flow.js
- A setValue type added to spackages/test-utils/types/index.d.ts and a test in packages/test-utils/types/wrapper.ts

I also think we should throw an error if setValue is called on an element that it won't affect. I think it affects most of the text content input types, like `text`, `email`. You can see a full list here—https://html.spec.whatwg.org/multipage/input.html#attr-input-type-keywords.

Could you add tests for each of the text control elements and confirm that `setValue` does set the v-model value for each type. Then can you write a test and throw an error if `setValue` is called on an element that isn't a text control.

Thanks 😀

beyer-martin:
I checked that v-model works on every input and confirmed it looking in v-model source code (https://github.com/vuejs/vue/blob/550c3c0d14af5485bb7e507c504664a7136e9bf9/src/platforms/web/compiler/directives/model.js)

I used the same rules that v-model uses to determine the input type, in order to check if the method was valid. Errors and suggestions are thrown if invalid.

Im having difficulties understanding what to test in packages/test-utils/types/wrapper.ts

Will continue with setSelected.

Thanks for reviewing

packages/test-utils/src/wrapper.js
```diff
@@ -531,6 +531,81 @@ export default class Wrapper implements BaseWrapper {
...
+
+    if (!el) {
+      throwError('cannot call wrapper.setValue() on a wrapper without an element')
+    }
+
+    const tag = el.tagName
+    const type = this.attributes().type
+    const event = 'input'
+
+    if (this.isVueComponent) {
```
eddyerburgh: Why is this check here? I don't think the function needs to behave differently if the wrapper contains a vue instance

packages/test-utils/src/wrapper.js
```diff
@@ -531,6 +531,81 @@ export default class Wrapper implements BaseWrapper {
...
+    if (!el) {
+      throwError('cannot call wrapper.setChecked() on a wrapper without an element')
+    }
+
+    const tag = el.tagName
+    const type = this.attributes().type
+    const event = 'change'
+
+    if (this.isVueComponent) {
+      throwError('wrapper.setChecked() cannot be called on component. Use wrapper.setValue() instead')
```
eddyerburgh: A component instance still has a root element, which can have its value set, so I think we should let setChecked work for vue instance wrappers

eddyerburgh:
> Im having difficulties understanding what to test in packages/test-utils/types/wrapper.ts

The tests are to check the typings are correct. You can just add two lines that call ```wrapper.setChecked(true)```
and 
```wrapper.setValue('some string')```

You can see an example for `findAll`: https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/types/test/wrapper.ts#L60

This file is compiled by typescript, so if the typings are incorrect the file won't compile and the type tests will fail.

File: packages/test-utils/src/wrapper.js (hunk 1 above)
Comment (beyer-martin) [reply packages/test-utils/src/wrapper.js]: Its unnecesary, will remove

File: packages/test-utils/src/wrapper.js (hunk 2 above)
Comment (beyer-martin) [reply packages/test-utils/src/wrapper.js]: My concern is related to how v-model behaves on components, but Im not sure. Let me know your final decision

**Comment by beyer-martin** [comment]
I believe its done, except for that mysterious failing compatibility test on checkbox and radio v-model @2.1.10. Any idea on that? Thanks

**Comment by beyer-martin** [comment]
@eddyerburgh any idea?

File: packages/test-utils/src/wrapper.js (hunk 2 above)
Comment (eddyerburgh) [reply packages/test-utils/src/wrapper.js]: The v-model would be set on the root element, so if you have a component like this:

```html
<template>
  <input type="text" />
</template>
```

And you called `setValue`, you would set the value of the input:

```
const wrapper = mount(TestComponent)
wrapper.setValue('some val')
```

All component instance wrappers have a root element (this.element) so you can keep the current code and it would work correctly on instance wrappers.

eddyerburgh:
Sorry for the delay, I must have missed the notifications from this PR.

I'm not sure why the tests fail on 2.1.10 without investigating. Have you debugged locally? There might be a different implementation of v-model in 2.1.x

38elements:
If you  change vue and vue-template-compiler version to 2.1.10 in `package.json`  and reinstall node modules,
you can execute test in Vue@2.1.10 .

beyer-martin:
Its solved! Thanks for the help!

> There might be a different implementation of v-model in 2.1.x

That was in fact the problem, on that versions v-model on checkboxes and radios is listening for click events. With some changes now it should work on all versions

Anything missing?

packages/test-utils/src/wrapper.js
```diff
@@ -531,6 +531,116 @@ export default class Wrapper implements BaseWrapper {
...
+    if (!el) {
+      throwError('cannot call wrapper.setValue() on a wrapper without an element')
+    }
+
+    const tag = el.tagName
+    const type = this.attributes().type
+    const event = 'input'
+
+    if (tag === 'SELECT') {
+      throwError('wrapper.setValue() cannot be called on select')
```
eddyerburgh: You can tell the user to call setSelected(). Also, can you change the errors to this format:

`wrapper.setValue() cannot be called on a <select> element. Use wrapper.setSelected() instead`

packages/test-utils/src/wrapper.js
```diff
@@ -531,6 +531,116 @@ export default class Wrapper implements BaseWrapper {
...
+    } else {
+      throwError('wrapper.setValue() cannot be called on this element')
+    }
+  }
+
+  /**
+   * Checks radio button or checkbox element
+   */
+  setChecked (checked: boolean) {
+    if (typeof checked !== 'undefined') {
```
eddyerburgh, line 567: I think we should default to true, so you can call setChecked() and it will set the element to checked. What do you think?

packages/test-utils/src/wrapper.js
eddyerburgh: Can you remove this check for isVueComponent

packages/test-utils/types/index.d.ts
```diff
@@ -72,6 +72,10 @@ interface BaseWrapper {
   setData (data: object): void
   setMethods (data: object): void
   setProps (props: object): void
+
+  setValue (value: any): void
+  setChecked (checked: boolean): void
```
eddyerburgh, line 77: setSelected is missing

test/specs/wrapper/setValue.spec.js
```diff
@@ -0,0 +1,69 @@
...
+describeWithShallowAndMount('setValue', (mountingMethod) => {
+  it('sets element value', () => {
+    const wrapper = mountingMethod(ComponentWithInput)
+    const input = wrapper.find('input[type="text"]')
+    input.setValue('foo')
+
+    expect(input.element.value).to.equal('foo')
+  })
+
+  it('calls trigger with input event', () => {
```
eddyerburgh: We can remove this test, as long as the dom updates with the v-model, it doesn't matter whether we call trigger or dispatch an event directly.

eddyerburgh:
Good job finding that 2.1.x listens to the click event 👏

I've just realized there aren't any docs pages. Can you add them in this PR?

packages/test-utils/src/wrapper.js
beyer-martin: It is set to true if no argument is passed. It isn't set on the parameters as a default but it behaves the way you suggest, there is a test for this specific case.

packages/test-utils/types/index.d.ts
beyer-martin: Ups, sorry. Done

beyer-martin:
> I've just realized there aren't any docs pages. Can you add them in this PR?

Yeah of course, will do!

eddyerburgh:
Hey @beyersito do you have time to add docs this week? It would be great to get this merged 😄

38elements:
Since there is no reply, I think it is better to close this pull request and someone send new one, or merge this one and someone add document.

beyer-martin:
Sorry for the delay, will add docs this week or if someone prefers to add docs great

lmiller1990:
Let's get this merged, I can add docs tomorrow unless you were about to do so @beyersito ?

eddyerburgh:
Great, I was thinking about this PR earlier 👍

beyer-martin:
Please do @lmiller1990 
Been very busy.
Please accept my apologies 😢

eddyerburgh (approved):
Thanks @beyersito and @lmiller1990

PR #562, opened 2018-04-25
docs: switch 'Action' component to 'Getters' to match other Vuex examples

chrisdiana:
Other Vuex examples are using the the type of store method as the name of the component:

Actions:
```
import Actions from '../../../src/components/Actions'
```

Modules:
```
import Modules from '../../../src/components/Modules'
```

Set Getters example component to match these to minimize confusion. Looks like Portuguese was already fixed.


eddyerburgh:
Thanks for the PR.

This PR is against the incorrect branch, gh-pages is the branch that includes the built docs. Can you make a new PR with the dev branch?

chrisdiana:
whoops sorry about that @eddyerburgh. I'll resubmit against dev.

Issue #563, opened 2018-04-25
Adding method to set data object 


shortdiv:
### What problem does this feature solve?
Currently, `setData` only works with setting single data values, but it doesn't allow for setting objects. For example, given `{ text : { message: 'null' } }`, I can't set the data with `setData` without mutating the entire object; i.e. `setData({ text: { message: 'hello' } })`. 

This is problematic because I might want to append a k/v pair to an object while still preserving existing k/v pairs. While this functionality is available via "vanilla vue" through `wrapper.vm.$set(obj, key, value)`, it feels like an inconsistency given that you can set data using `setData` otherwise.

### What does the proposed API look like?
The proposed API adds a `setDataObj` method to vue-test-utils thereby allowing you to set data if updating an object rather than a value. It would be very similar to Vue.set or vm.$set but be more consistent with how data is currently set in tests. 
```
setDataObj(obj, key, value)
```

shortdiv:
I implemented a proof of concept, to make sure this could work -> https://github.com/shortdiv/vue-test-utils/commit/ca4413699037652b8bfd1e3d2fc0e8f4297ae177

eddyerburgh:
This is a good idea.

Instead of adding a new method, I think we should change the behavior of `setData` to not overwrite objects.

shortdiv:
Sweet! I'll update my code and submit a PR for this

johnyluyte:
Hello, 

Using `1.0.0-beta.16`, I'm having similar problem with `setData()`, not with Object, but with Arrays.

My Component:
````javascript
data() {
  return {
    items: [
      { id: 1, name: "Apple" },
      { id: 2, name: "Banana" }
    ]
  }
}
````

Testing:
````javascript

wrapper.setData({
  items: []
})
// What's wrong: Does not clear the items array
// wrapper.vm.items
// { id: 1, name: "Apple" },
// { id: 2, name: "Banana" }

wrapper.setData({
  items: [
    { id: 1, name: "SuperWatermelon" }
  ]
})
// What's wrong: Modified the 1st item, but didn't remove the 2nd item.
// wrapper.vm.items
// { id: 1, name: "SuperWatermelon" },
// { id: 2, name: "Banana" }

wrapper.setData({
  items: [
    { id: 1, name: "Dog" }
    { id: 2, name: "Cat" }
  ]
})
// What's wrong: Nothing, this one is successful
// wrapper.vm.itees
// { id: 1, name: "Dog" },
[...]

PR #565, opened 2018-04-26
refactor: Update setData method (address #563)

shortdiv:
This is a refactor of the setData method to allow for setting data values and objects. The current implementation of setData only allows for setting values and not objects. 

Addresses #563

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -420,13 +420,14 @@ export default class Wrapper implements BaseWrapper {
     }
 
     Object.keys(data).forEach((key) => {
-      // $FlowIgnore : Problem with possibly null this.vm
-      if (typeof data[key] === typeof {} && data[key] !== null) {
+      if (typeof data[key] === 'object' && data[key] !== null) {
         Object.keys(data[key]).forEach((key2) => {
-          var newObj = Object.assign({}, data[key], data[key][key2])
+          const newObj = { ...data[key], ...data[key][key2] }
+          // $FlowIgnore : Problem with possibly null this.vm
           this.vm.$set(this.vm, [key], newObj)
         })
       } else {
+        // $FlowIgnore : Problem with possibly null this.vm
         this.vm.$set(this.vm, [key], data[key])
       }
     })
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -421,7 +421,14 @@ export default class Wrapper implements BaseWrapper {
 
     Object.keys(data).forEach((key) => {
       // $FlowIgnore : Problem with possibly null this.vm
-      this.vm.$set(this.vm, [key], data[key])
+      if (typeof data[key] === typeof {} && data[key] !== null) {
+        Object.keys(data[key]).forEach((key2) => {
+          var newObj = Object.assign({}, data[key], data[key][key2])
+          this.vm.$set(this.vm, [key], newObj)
+        })
+      } else {
+        this.vm.$set(this.vm, [key], data[key])
+      }
     })
   }
 
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -1,6 +1,7 @@
 // @flow
 
 import Vue from 'vue'
+import merge from 'lodash/merge'
 import getSelectorTypeOrThrow from './get-selector-type'
 import {
   REF_SELECTOR,
@@ -421,11 +422,10 @@ export default class Wrapper implements BaseWrapper {
 
     Object.keys(data).forEach((key) => {
       if (typeof data[key] === 'object' && data[key] !== null) {
-        Object.keys(data[key]).forEach((key2) => {
-          const newObj = { ...data[key], ...data[key][key2] }
-          // $FlowIgnore : Problem with possibly null this.vm
-          this.vm.$set(this.vm, [key], newObj)
-        })
+        // $FlowIgnore : Problem with possibly null this.vm
+        const newObj = merge(this.vm[key], data[key])
+        // $FlowIgnore : Problem with possibly null this.vm
+        this.vm.$set(this.vm, [key], newObj)
       } else {
         // $FlowIgnore : Problem with possibly null this.vm
         this.vm.$set(this.vm, [key], data[key])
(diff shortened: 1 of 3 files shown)


eddyerburgh:
Ah I'm so sorry, I wasn't clear in my comment!
 
I think we should keep the signature the same as it is currently. Instead, we could update the function to do a deep merge:

```js
const Cmp = {
  data: () => ({
     anObject: {
       propA: 'a',
       propB: 'b'
     }
  }) 
}
const wrapper = mount(Cmp)
wrapper.setData({
  anObject: {
    propB: 'c'
  }
})
wrapper.vm.anObject.propA // 'a'
wrapper.vm.anObject.propB // 'c'
```

shortdiv:
Ah gotcha, will update accordingly!

packages/test-utils/src/wrapper.js
```diff
@@ -421,7 +421,14 @@ export default class Wrapper implements BaseWrapper {
 
     Object.keys(data).forEach((key) => {
       // $FlowIgnore : Problem with possibly null this.vm
```
38elements: I think that this line should be removed.

packages/test-utils/src/wrapper.js
```diff
@@ -421,7 +421,14 @@ export default class Wrapper implements BaseWrapper {
...
-      this.vm.$set(this.vm, [key], data[key])
+      if (typeof data[key] === typeof {} && data[key] !== null) {
+        Object.keys(data[key]).forEach((key2) => {
+          var newObj = Object.assign({}, data[key], data[key][key2])
+          this.vm.$set(this.vm, [key], newObj)
```
38elements: I think it is necessary to add the following line before this line.
`// $FlowIgnore : Problem with possibly null this.vm`

packages/test-utils/src/wrapper.js
```diff
@@ -421,7 +421,14 @@ export default class Wrapper implements BaseWrapper {
...
+        })
+      } else {
+        this.vm.$set(this.vm, [key], data[key])
```
38elements: I think it is necessary to add the following line before this line.
`// $FlowIgnore : Problem with possibly null this.vm`

packages/test-utils/src/wrapper.js
```diff
@@ -421,7 +421,14 @@ export default class Wrapper implements BaseWrapper {
 
     Object.keys(data).forEach((key) => {
       // $FlowIgnore : Problem with possibly null this.vm
-      this.vm.$set(this.vm, [key], data[key])
+      if (typeof data[key] === typeof {} && data[key] !== null) {
+        Object.keys(data[key]).forEach((key2) => {
+          var newObj = Object.assign({}, data[key], data[key][key2])
```
38elements: I think `spread syntax` is frequently used in this repository.
I think it is better to change `var` to `const`.

```js
const newObj = { ...data[key], ...data[key][key2]}
```

packages/test-utils/src/wrapper.js
```diff
@@ -421,7 +421,14 @@ export default class Wrapper implements BaseWrapper {
 
     Object.keys(data).forEach((key) => {
       // $FlowIgnore : Problem with possibly null this.vm
-      this.vm.$set(this.vm, [key], data[key])
+      if (typeof data[key] === typeof {} && data[key] !== null) {
```
38elements: Would you tell me why are you using `typeof data[key] === typeof {}`, not `typeof data[key] === 'object'` ?
I think  `typeof data[key] === 'object'` is better.

38elements:
It is necessary to execute `yarn flow`.

packages/test-utils/src/wrapper.js
```diff
@@ -420,13 +420,14 @@ export default class Wrapper implements BaseWrapper {
     }
 
     Object.keys(data).forEach((key) => {
-      // $FlowIgnore : Problem with possibly null this.vm
-      if (typeof data[key] === typeof {} && data[key] !== null) {
+      if (typeof data[key] === 'object' && data[key] !== null) {
         Object.keys(data[key]).forEach((key2) => {
-          var newObj = Object.assign({}, data[key], data[key][key2])
+          const newObj = { ...data[key], ...data[key][key2]}
```
38elements: It is necessary to add space.

```js
const newObj = { ...data[key], ...data[key][key2] }
```

38elements:
It is necessary to execute `yarn lint`.

packages/test-utils/src/wrapper.js
```diff
@@ -420,8 +420,16 @@ export default class Wrapper implements BaseWrapper {
     }
 
     Object.keys(data).forEach((key) => {
-      // $FlowIgnore : Problem with possibly null this.vm
-      this.vm.$set(this.vm, [key], data[key])
+      if (typeof data[key] === 'object' && data[key] !== null) {
```
38elements: If `data[key]` is an object, `data[key]` is not null.
I think `if (typeof data[key] === 'object') {` is better.

test/specs/wrapper/setData.spec.js
```diff
@@ -132,4 +132,25 @@ describeWithShallowAndMount('setData', (mountingMethod) => {
     wrapper.setData({ message: null })
     expect(wrapper.text()).to.equal('There is no message yet')
   })
+  it('should update a data object', () => {
```
38elements: I think it is better to add a blank line before this line.

test/specs/wrapper/setData.spec.js
```diff
@@ -132,4 +132,25 @@ describeWithShallowAndMount('setData', (mountingMethod) => {
...
+    }
+    const wrapper = mountingMethod(Cmp)
+    wrapper.setData({
+      anObject: {
+        propA: {
+          prop1: 'c'
+        }
+      }
+    })
+    expect(wrapper.vm.anObject.propA.prop1).to.equal('c')
```
eddyerburgh: Can you give propA another property, and add an assertion that the value is unchanged after calling setData.

test/specs/wrapper/setData.spec.js
```diff
@@ -132,4 +132,25 @@ describeWithShallowAndMount('setData', (mountingMethod) => {
     wrapper.setData({ message: null })
     expect(wrapper.text()).to.equal('There is no message yet')
   })
+  it('should update a data object', () => {
+    const Cmp = {
```
eddyerburgh: The convention we follow in this repo is to name these TestComponent

packages/test-utils/src/wrapper.js
shortdiv: The null check was necessary because `null` is of type object. If you're updating an object with `{message: null}`, it will pass this conditional, which it technically shouldn't since it's not a nested object.

shortdiv:
Updated! ✨

eddyerburgh (approved):
Great work, thanks :)

PR #568, opened 2018-04-27
docs: switch 'Action' component to 'Getters' to match other Vuex examples

chrisdiana:
Resubmitting on the correct branch this time.

Other Vuex examples are using the the type of store method as the name of the component:

Actions:
```
import Actions from '../../../src/components/Actions'
```

Modules:
```
import Modules from '../../../src/components/Modules'
```

Set Getters example component to match these to minimize confusion. Looks like Portuguese was already fixed.


eddyerburgh (approved):
Thanks :)

PR #578, opened 2018-05-04
Fixed typo

symmetriq:


eddyerburgh (changes requested):
Thanks for the PR. Can you remove the dist files from the PR, they're only updated when a version is released

symmetriq:
Ahh, sorry about that. Done.

eddyerburgh (approved):
Thanks :)

Issue #550, opened 2018-04-23
Cannot read property 'deps' of null


Mattchewone:
### Version
1.0.0-beta.14

### Reproduction link
[https://codesandbox.io/s/1pm8pzxm4](https://codesandbox.io/s/1pm8pzxm4)

### Steps to reproduce
Open the test tab to see the failing test.

### What is expected?
Using `findAll` with `.at()` to trigger a click

### What is actually happening?
Test is failing with "Cannot read property 'deps' of null" message

---
I am storing open/close states in parent component and testing by triggering clicks in the scopedSlot component.

Mattchewone:
Also not sure why in the `codesandbox` example these aren't printed in the console, but on my local version my terminal is filled with these:
```
[Vue warn]: Unknown custom element: <v-card-title>
```

I can't see what is wrong with the way I have added Vuetify to the localVue and Vue instance. So any insights would be greatly appreciated.

Mattchewone:
@eddyerburgh Could you point me in the right direction as to where I could look into this so I could perhaps create a PR for a fix. Assuming this is a bug and not an issue with my code..

Thanks

eddyerburgh:
The issue will be in either this file—https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/set-watchers-to-sync.js, or this file—https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/order-watchers.js

Mattchewone:
Thanks @eddyerburgh Is there a guide on getting the test utils running locally. I finally managed to get it to installed but getting lots of flow check errors when running the test suite. Wanted to get it to run locally before I looked into those files. 
Maybe I am missing something from the setup, but I cloned and then installed, and it kept failing on the postinstall with a `lerna` failure. So I installed some deps to get passed that, but then running `npm test` just fails on `flow check` - `Module not found: ...`

eddyerburgh:
You need to run `lerna bootstrap` in the root of the project, then `yarn install` in the root. I'll add a contributing guide soon.

PR #583, opened 2018-05-05
fix: only order deps if watcher exists

eddyerburgh:
Fixes https://github.com/vuejs/vue-test-utils/issues/550


Mattchewone:
Thanks sorry I hadn’t managed to look into it to help out.

PR #586, opened 2018-05-05
WIP: Add parent in create-instance

### LINKED ISSUE #472: wrapper.vm.$listeners not equal the actual vueInstance.$listeners
### Version
1.0.0-beta.12

### Reproduction link
[https://jsbin.com/nopulunamo/3/edit?html,console,output](https://jsbin.com/nopulunamo/3/edit?html,console,output)

### Steps to reproduce
1. create a wrapper by method mount , and pass  listeners param

       const wrapper = mount(myComponent, {
          listeners: {
            change: someFunction,
          },
        });

### What is expected?
     wrapper.vm.$listerners  = {
        change: function invoker(){}
     }

     invoker.fns  ===  someFunction

    wrapper.vm.$listerners.change === someFunction

eddyerburgh:
This will solve a few issues. It should make scopedSlots and slots easier to implement

Problems:
- Must call `$mount` to create a nested component instance, which is incompatible with server-test-utils

Fixes #472


eddyerburgh:
@38elements One of the issues with this PR is that it's impossible to add this functionality—https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/mounting-options/slots.spec.js#L107, where the slot is rendered using the context of the component that is mounted. 

On reflection, I don't think we should have this behavior, since slots do not behave like this in Vue. What do you think?

38elements:
Since the behavior is a rare use case and rarely used to write specifications,
I think it is unnecessary to support the behavior.

PR #587, opened 2018-05-06
fix(types): shallow -> shallowMount

38elements:
This changes `shallow` to `shallowMount` in types.


eddyerburgh (approved):
Thanks. I'm planning on moving the whole repository to use typescript instead of flow, that way we could use the types in the unit tests and stop them going out of sync with the source code.

38elements:
I think it is good.

PR #595, opened 2018-05-10
fix: automatically extend extended child components

### LINKED ISSUE #382: Child component using Vue.extends not receiving global $state property from Vuex
### Version
1.0.0-beta.10

### Reproduction link
[https://jsfiddle.net/7t28dy1u/](https://jsfiddle.net/7t28dy1u/)

### Steps to reproduce
Run the fiddle - console error because `this.$store` is undefined on the child component.

### What is expected?
All child components should receive the global `$store` mixin from Vuex.

Only the root element is receiving the `$store` property.

---
I can fix the jsbin by removing the `Vue.extend` notation and just passing the raw javascript objects into mount. I also confirmed that this works as expected when using Vue to construct the component normally (not using vue-test-utils).

I ran into this because I was using ES6 imports to import my store into all my app components. I switched to using `this.$store` which works fine in the app, but broke a lot of my unit tests that were using vue-test-utils to mount the components.

eddyerburgh:
Closes #382


flow/options.flow.js
```diff
@@ -9,5 +9,6 @@ declare type Options = { // eslint-disable-line no-undef
     stubs?: Object,
     context?: Object,
     attrs?: Object,
-    listeners?: Object
+    listeners?: Object,
+    logModifiedComponents?: Boolean
```
38elements, line 13: I think it is necessary to add  the types to the below files.

https://github.com/vuejs/vue-test-utils/blob/14c40e6af2517b5a058f84d8cea51a711d929924/packages/test-utils/types/index.d.ts#L142

https://github.com/vuejs/vue-test-utils/blob/14c40e6af2517b5a058f84d8cea51a711d929924/packages/server-test-utils/types/index.d.ts#L52

docs/en/api/config.md
```diff
@@ -79,3 +79,18 @@ VueTestUtils.config.provide['$logger'] = {
   }
 }
 ```
+
+### `logModifiedComponents`
+
+- type: `Boolean`
+- default: `true`
+
+Logs warning when extended child components are automatically stubbed. Hides warnings when set to `false`.
```
38elements: I think that it is better to describe that `logModifiedComponents` can not be overwritten by mounting options.

PR #603, opened 2018-05-13
docs: Fix typos

vjoao:
There was a typo on "silumar" -> "simular" which mean "to simulate". The method name is also outdated.


eddyerburgh (approved):
Thanks

PR #604, opened 2018-05-14
fix: Do not deep merge array data in setData

lchanmann:
## What went wrong?

In https://github.com/vuejs/vue-test-utils/pull/565, we do a deep merge on Object data however if the data is an array deep merge would mutate the data so `vm.$set` would have no effect.

The `typeof` operator is kind of inconsistent (i.e. `typeof [] === 'object'`) and we need also array checking to avoid deep merge array data.


lchanmann:
ping @shortdiv, @eddyerburgh

eddyerburgh (approved):
Thanks for the fix

lchanmann:
TY

PR #606, opened 2018-05-14
feat: render component name in stub

### LINKED ISSUE #28: Render component names/ids when shallow
Is it possible/a good idea for `shallow()` output to included the component names?

Right now `shallow(component).html()` gives me:

```html
<div data-v-1d62245a="" id="app"><!----> <!----></div>
```

It would be great if included the name of the component, or similar. For example:

```html
<div data-v-1d62245a="" id="app"><!--Login--> <!--SignUp--></div>
```

This would me a I could make basic assertions about what components _would_ have been included, based on conditional logic of the component under test.

Let me know if the current behaviour is a deliberate design decision, or if you're interested in a PR.

### LINKED ISSUE #410: Automatically stub nested components with their names instead of `<!---->`
In my workflow it is a pretty common use case to have a unit test that checks if some nested component is being rendered conditionally.

For example i have a component with a template like this (I use pug for templating, just so there won't be any confusion):

```pug
.container
  MyCustomComponent(v-if="condition" :title="cog.title")
```

And I want to have a unit test that makes sure that `MyCustomComponent` is only rendered when that condition is satisfied.

I'm coming from React background, where `enzyme`'s shallow render passes component's display name by default. So if I would shallow render the setup above - the `wrapper.html()` would look something like this (if the `v-if` condition was met):

```html
<div class="container">
  <MyCustomComponent title="something from cog.title" />
</div>
```

But in `vue-test-utils` the result would actually look like this:

```html
<div class="container">
  <!---->
</div>
```

There is currently an option to supply a stubs object to `shallow` render method, but that requires you to manually write a list of components and their respective stubs, which is not time-efficient.

Currently I wrote a simple stub-generator, which acts somewhat like identity-object-proxy, and basically looks at the list of `components` in options of the component we want to render, and generates an object of stubs.

That allows me to achieve a result like this:

```html
<div class="container">
  <mycustomcomponent title="something from cog.title" />
</div>
```
Which is basically what I need.

I can see more people that might want this kind of behaviour to speed up the unit testing, as this is a pretty common approach in other systems.

My stubs generator code for context:
```js
// Using custom proxy to automatically stub with proper component names
const idObj = new Proxy({}, {
  get: function getter(target, key) {
    return `<${key} />`;
  },
});

// Generate the object of stubs with proxy
const genStubs = component => Object.keys(component.options.components).reduce((acc, comp) => {
  acc[comp] = idObj[comp];
  return acc;
}, {});
```

Pretty basic, but it allows me to simply use `wrapper.find('<component name>');` to conditionally check if something was rendered, without manually supplying the stubs object every time, since I have a custom wrapper around `shallow` that passes my generator to the `stubs` option.

I'm not sure if this should be the default behaviour, since people might rely on current way of stubbing in the existing code, but I can see something like a flag "useNamedStubs" for `shallow` options, or something similar.

eddyerburgh:
This change renders stubbed components with the component name and a stub suffix:

```js
const wrapper = shallow({
  template: `<clock-component />`,
  components: {
    ClockComponent
  }
})

wrapper.html() // '<clock-component-stub></clock-component>'
```

Closes #410 
Closes #28

### DIFF
--- a/packages/shared/stub-components.js
+++ b/packages/shared/stub-components.js
@@ -60,7 +60,9 @@ function createStubFromString (templateString: string, originalComponent: Compon
 function createBlankStub (originalComponent: Component) {
   return {
     ...getCoreProperties(originalComponent),
-    render: h => h('')
+    render (h) {
+      return h(`${originalComponent.name}-stub`)
+    }
   }
 }
 
@@ -78,7 +80,7 @@ export function createComponentStubs (originalComponents: Object = {}, stubs: Ob
       if (typeof stub !== 'string') {
         throwError('each item in an options.stubs array must be a string')
       }
-      components[stub] = createBlankStub({})
+      components[stub] = createBlankStub({ name: stub })
     })
   } else {
     Object.keys(stubs).forEach(stub => {
@@ -89,7 +91,7 @@ export function createComponentStubs (originalComponents: Object = {}, stubs: Ob
         throwError('options.stub values must be passed a string or component')
       }
       if (stubs[stub] === true) {
-        components[stub] = createBlankStub({})
+        components[stub] = createBlankStub({ name: stub })
         return
       }
 
@@ -124,7 +126,7 @@ export function createComponentStubs (originalComponents: Object = {}, stubs: Ob
       }
       // ignoreElements does not exist in Vue 2.0.x
       if (Vue.config.ignoredElements) {
-        Vue.config.ignoredElements.push(stub)
+        Vue.config.ignoredElements.push(`${stub}-stub`)
       }
     })
   }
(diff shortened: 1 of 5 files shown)


iztsv:
@eddyerburgh thanks for the feature!

If `ClockComponent` uses a child component in template like `<child-component :some-prop="someValue"></child-component>`, will stubbed html be like:

```
'<clock-component-stub>
  <child-component some-prop="VALUE"></child-component>
</clock-component>'
```
?

eddyerburgh:
shallow does not render slots by defualt.

You could acheive that effect by manually stubbing both components:

```js
const wrapper = mount(TestComponent, {
  stubs: [ 'ChildComponent', 'ClockComponent']
})
```

trollepierre:
I don't want to have to stub any SubChildComponent in my test.

eddyerburgh:
If you want this feature then create an issue and we can discuss it there

stephane:
I'm a bit lost in the various related github issues/PR about stubs and I don't understand why I need to add a `stubs:  { 'VueComponent': '<div />' }` to my tests which use `shallowMount` to avoid warnings.
From v3.0.0-beta.16 to beta.18

38elements:
If you are concerned about it, I think it is better that you create an issue.

PR #607, opened 2018-05-15
fix: inconsistent message log

briwa:
Fix inconsistent log message on `create-instance.js`

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -74,7 +74,7 @@ export default function createInstance (
     if (component.components[c].extendOptions &&
       !instanceOptions.components[c]) {
       if (options.logModifiedComponents) {
-        warn(`an extended child component ${c} has been modified to ensure it has the correct instance properties. This means it is not possible to find the component with a component selector. To find the component, you must stub it manually using the stubs mounting option.`)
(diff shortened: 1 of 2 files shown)


38elements:
I think this change is unnecessary.
I executed git clone this repository and `yarn install` and `npm run test:unit`.
It worked normally.
It is necessary to execute `npm run build:test` before `npm run test:unit:only`.
The explanation is below.
https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md#commonly-used-npm-scripts

https://github.com/vuejs/vue-test-utils/blob/934745b7bb89c60d926622250d450bee0fdbf598/package.json#L23-L26

briwa:
@38elements my bad, you're right. I might have a setup problem. After doing a fresh install there's no dependency problems. I've updated my PR description.

However, there's one thing I noticed where the log messages was somewhat incorrect. Was wondering if it's still needed to change it? Please find 32bb238. Let me know what you think.

briwa:
@38elements ah sorry, I missed it. seems like it's an intended change from you https://github.com/vuejs/vue-test-utils/commit/dea5dba7. My bad! Closing this PR now

PR #613, opened 2018-05-15
fix: clone propsData to avoid mutation

### LINKED ISSUE #393: Default props modified between tests suites
### Version
1.0.0-beta.11

### Reproduction link
[https://jsfiddle.net/bqtfqwpw/](https://jsfiddle.net/bqtfqwpw/)

### Steps to reproduce
Tested component: https://jsfiddle.net/g7s289sh/
Tests suites: https://jsfiddle.net/bqtfqwpw/

1. Install vue/test-utils 1.0.0-beta.10
2. Run tests => All ok
3. Update vue/test-utils to 1.0.0-beta.11
4. Run tests => Test fails

### What is expected?
Tests passed

Tests failed
image

---
Looks like vue/test-utils shared default props between test suites

briwa:
Potentially fix #393.

In the issue it was mentioned that cloning might cause other problems with stubbing. Was wondering if doing this might break it? Seems like all other tests are passing, maybe I'm missing something 🤔 @eddyerburgh 

Also, I've added a test to make sure it won't happen again, but I am not sure if this is the best way to do it, whether it belongs to `mounting-options` or not, whether the test structure adheres the repo standard, etc. I'm open for suggestions.

Thanks.

### DIFF
--- a/flow/options.flow.js
+++ b/flow/options.flow.js
@@ -1,5 +1,6 @@
 declare type Options = { // eslint-disable-line no-undef
     attachToDocument?: boolean,
+    propsData?: Object,
     mocks?: Object,
     methods?: Object,
     slots?: Object,
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -57,7 +57,7 @@ export default function createInstance (
 
   const Constructor = vue.extend(component)
 
-  const instanceOptions = { ...options }
+  const instanceOptions = { ...options, propsData : { ...options.propsData } }
   deleteoptions(instanceOptions)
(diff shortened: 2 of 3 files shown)


eddyerburgh:
Cloning will be fine with stubs. I thought cloning a function would break the reference, but I was wrong.

You need to fix a flow error. You can add a propsData property to the options object in flow/options.flow.js.

Then I'm happy to merge 👍

briwa:
Alright, fixed @eddyerburgh. Thanks.

eddyerburgh:
Thanks for the quick fix

eddyerburgh (approved):

PR #618, opened 2018-05-16
fix(types): add render() to packages/server-test-utils/types/index.d.ts

38elements:
This adds the type of [render()](https://vue-test-utils.vuejs.org/en/api/render.html).


eddyerburgh (approved):
Thanks

PR #628, opened 2018-05-19
fix: run vm.$forceUpdate in setData and setProps method

lusarz:
After update from vue-test-utils from `1.0.0-beta.11` to current version, I noticed that children are not always updated:

    it('download button is disabled when content is empty', () => {    
        wrapper.setData({ content: [] });
        expect(wrapper.find(vButton).vm.disabled).to.be.true;
    });

Using `wrapper.vm.$forceUpdate()` fixed this issue. To be honest I'm not sure why :)

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -432,6 +432,9 @@ export default class Wrapper implements BaseWrapper {
         this.vm.$set(this.vm, [key], data[key])
       }
     })
+    if (this.vm) {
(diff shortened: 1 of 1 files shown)


eddyerburgh:
Thanks for the PR. Can you add a test case that was broken before the bug fix?

lusarz:
Sure, I'll try to add test that covers my situation

lusarz:
After investigation I noticed that this bug occurred in very very uncommon scenario which is not worth attention here.

PR #630, opened 2018-05-19
fix: remove error-wrapper (fix #629)

### LINKED ISSUE #629: Remove ErrorWrapper
`ErrorWrapper` is returned only in `wrapper.find`. I think there is no benefit of using it over returning null.

`wrapper.find` returns `null` (or `undefined`) when no item is found, similarly to #`Array.prototype.find`

lusarz:


eddyerburgh:
Thanks for the work, but it is better to work on PRs for changes that have been approved. You can find 
a list of changes that have been approved in the issues by searching the intend to implement tag.

I'm closing this PR, as there is no plan to remove ErrorWrapper.

PR #643, opened 2018-05-23
docs: video link for common tips

briwa:
I thoroughly enjoyed the deck provided under the Common Tips section on the docs. I thought it might help others if they can see the deck while watching the actual Youtube video of Matt O'Connell himself speaking at VueNYC on the actual deck.

Not sure if I'm doing it right on this one, I'm open for ideas or concerns.

Thanks.


eddyerburgh:
Good idea, I really enjoyed the talk too :)

How about we link to the video by default, and include a link to the deck, rather than the other way round. 

Also we should make sure the parentheses aren't in the link, otherwise it looks like one long link, rather than two separate links. Something like this:
```
This topic is discussed with more details in a [great presentation by Matt O'Connell](http://slides.com/mattoconnell/deck#/) ([video](https://www.youtube.com/watch?v=OIpfWTThrK8)).
```

briwa:
@eddyerburgh Good idea. I was even thinking to simply replace the deck with the video link, since we might no longer need the deck if we have the video link. What do you think? I can also update the `ja` and `zh` section.

eddyerburgh:
Ok yes, could you replace the deck link with the video link and update the japanese and Chinese translations too please?

eddyerburgh (approved):
Thanks 😁

PR #652, opened 2018-05-24
feat: overwrite arrays

### LINKED ISSUE #605: Using `setData` with array does not trigger watchers or update the view
### Version
1.0.0-beta.16

### Reproduction link
[https://github.com/tbuteler/vue-test-utils-issue](https://github.com/tbuteler/vue-test-utils-issue)

### Steps to reproduce
Clone the repository, then run:

```bash
yarn install
yarn run tests
```

### What is expected?
Component:

```Vue
<template>
  <div id="app">
    <ul v-if="items.length > 0">
      <li v-for="item in items">
        {{ item }}
      </li>
    </ul>
  </div>
</template>

<script>
export default {
  name: 'app',
  data () {
    return {
      loaded: false,
      items: []
    }
  },
  watch: {
    items () {
      this.loaded = true
    }
  }
}
</script>
```

Test:
```javascript
import Vue from 'vue'
import App from '../src/App.vue'

describe('App.vue', () => {
  it('array data triggers watchers', () => {
    const wrapper = mount(App)
    wrapper.setData({ items: ['Foo', 'Bar'] })
    expect(wrapper.vm.loaded).toBe(true)
    expect(wrapper.findAll('li').at(0).text()).toBe('Foo')
    expect(wrapper.findAll('li').at(1).text()).toBe('Bar')
  })
})
```
After doing `setData` I expected the HTML to be updated and the watcher (which sets `loaded` to `true`) to be run.

Although the `wrapper.vm.items` property yields what I just set, no updates are triggered on the wrapper and the tests fail.

---
Test passes in 1.0.0-beta.15.

eddyerburgh:
breaking change

Closes #605

### DIFF
--- a/docs/api/wrapper/setData.md
+++ b/docs/api/wrapper/setData.md
@@ -1,6 +1,8 @@
 ## setData(data)
 
-Sets `Wrapper` `vm` data and forces update.
+Sets `Wrapper` `vm` data.
+
+setData works by merging existing properties, except for arrays which are overwritten.
 
 **Note the Wrapper must contain a Vue instance.**
 
--- a/flow/modules.flow.js
+++ b/flow/modules.flow.js
@@ -12,7 +12,7 @@ declare module 'lodash/cloneDeep' {
   declare module.exports: any;
 }
 
-declare module 'lodash/merge' {
+declare module 'lodash/mergeWith' {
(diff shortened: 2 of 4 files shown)


lmiller1990 (approved):
Looks good to me

PR #654, opened 2018-05-25
fix: make component construct correctly which class extends from Vue

kuitos:
from `mount` typings we can pass a VueClass type component:
https://github.com/vuejs/vue-test-utils/blob/c0b2101f75b80e1dd5691656d0929085d5adb886/packages/test-utils/types/index.d.ts#L149

consider such a scenario
```js
class Component extends Vue {
    $mount() {
        // do something
    }
}
mount(Component)
```

or a more simple one:
```js
const Component = Vue.extend()
Component.prototype.$mount = function() { /* do something */}
mount(Component)
```

The Component prototype will be replaced if we extend the passed component directly.
https://github.com/vuejs/vue-test-utils/blob/c0b2101f75b80e1dd5691656d0929085d5adb886/packages/create-instance/create-instance.js#L58

### DIFF
--- a/flow/vue.flow.js
+++ b/flow/vue.flow.js
@@ -2,6 +2,6 @@
 
 // Importing these types declares them, so they are available globally
 
-declare type Component = Object // eslint-disable-line no-undef
+declare type Component = Object | Function // eslint-disable-line no-undef
 declare type VNode = Object // eslint-disable-line no-undef
 declare type SlotValue = Component | string | Array<Component> | Array<string>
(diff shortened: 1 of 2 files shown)


eddyerburgh (approved):
Thanks

kuitos:
wow so fast!👍
btw, any plans about the next releasing? I am working on a repo which depends on this feature... sorry to rush you😂

eddyerburgh:
Around a week 👍

Issue #582, opened 2018-05-05
The content of the template passed to slot option is not updated


38elements:
### Version
1.0.0-beta.15

### Reproduction link
[https://github.com/vuejs/vue-test-utils/blob/1a4d12fddf7adb5015986ed01209021ccd4ec281/test/specs/mounting-options/slots.spec.js#L111](https://github.com/vuejs/vue-test-utils/blob/1a4d12fddf7adb5015986ed01209021ccd4ec281/test/specs/mounting-options/slots.spec.js#L111)

### Steps to reproduce
Changing as below.
https://github.com/vuejs/vue-test-utils/blob/1a4d12fddf7adb5015986ed01209021ccd4ec281/test/specs/mounting-options/slots.spec.js#L109-L112

↓

```
      const wrapper5 = mountingMethod(ComponentWithSlots, { slots: { default: '1{{ foo }}2' }})
      expect(wrapper5.find('main').html()).to.equal('<main>1bar2</main>')
      wrapper5.trigger('keydown')
      expect(wrapper5.find('main').html()).to.equal('<main>1BAR2</main>') // add this line
      const wrapper6 = mountingMethod(ComponentWithSlots, { slots: { default: '<p>1</p><p>2</p>' }})
```

### What is expected?
The content of the template passed to slot option is  updated.

### What is actually happening?
The content of the template passed to slot option is not updated.

---
Would you please check below links?

* https://github.com/vuejs/vue-test-utils/pull/274#issuecomment-352262141

* https://github.com/vuejs/vue-test-utils/pull/274/commits/3b99090a6917b7861096b2903fc9ec7811607464#diff-dbfd1670fb2dd50be051adb6643c8c24

Issue #455, opened 2018-03-06
Cannot trigger method updated Vue lifecycle


ghost:
### Version
1.0.0-beta.11

### Reproduction link
[https://github.com/dungla4/vue-test-utils-vuex-router](https://github.com/dungla4/vue-test-utils-vuex-router)

### Steps to reproduce
- Clone repo then run: `npm install` or `yarn install`
- Run command: `npm test` or `yarn test`

### What is expected?
Should call `updated` Vue lifecycle method and print in log

### What is actually happening?
Not trigger `updated` method, just trigger `mounted` method

---
Normally, `updated` is triggered when template is changed. I tried to change data by `wrapper.setData`, or set data directly via `wrapper.vm.msg = ..` but no luck. This function works well if you run project by command `yarn start` and click some link at the bottom of page

38elements:
@eddyerburgh 
In real Vue, When `vm._watcher.sync` is `true`, `updated` hook is not called.
Do you think when `vm._watcher.sync` is `true`, `updated` hook should be called in Vue Test Utils?

eddyerburgh:
No, I think the solution is to either make a change to Vue core, or reimplement `update`. Changing the render watcher to sync has caused bugs that can't be fixed without changes to Vue core, or hacky code.

38elements:
Thank you for reply.
I think in order to solve this problem which `updated` hook is not executed when `sync` mounting option is `true`,
 it is necessary to let Vue instance to do different behavior from the real Vue instance.

If you are not resolving this issue currently,
I think I try to resolve this issue.

38elements:
Even if the behavior is different,
I think it is better to call `updated` hook synchronously when `vm._watcher.sync` is `true`.

eddyerburgh:
Ok, that would be great if you could investigate. I'd be interested to see a solution. I'm going to investigate changing Vue core to add a sync option this weekend so we can compare solutions 👍

38elements:
I am sorry.
Since I have work to do on this Saturday and Sunday,
I may not be able to make deliverable.
I will make deliverable within 8 days.

eddyerburgh:
That's fine, this issue might take a while to solve. 

[...]

PR #661, opened 2018-05-26
fix: #455

38elements:
This is related to #455 and #582.

updated hook is called in `vm._update()` at 2.0.8.
https://raw.githubusercontent.com/vuejs/vue/v2.0.8/dist/vue.js

I read below lines.

https://github.com/vuejs/vue/blob/f43ce3a5d8f73e273f2d03c9d86ea5662cda481a/src/core/instance/lifecycle.js#L182-L190

https://github.com/vuejs/vue/blob/653aac2c57d15f0e93a2c1cc7e6fad156658df19/src/core/observer/watcher.js#L187-L202

https://github.com/vuejs/vue/blob/b7445a2b945dcded287601ace8e711ab5cf35ab5/src/core/observer/scheduler.js#L130-L151

https://github.com/vuejs/vue/blob/653aac2c57d15f0e93a2c1cc7e6fad156658df19/src/core/observer/scheduler.js#L35-L94

https://github.com/vuejs/vue/blob/b7445a2b945dcded287601ace8e711ab5cf35ab5/src/core/observer/scheduler.js#L96-L105

https://github.com/vuejs/vue/blob/f43ce3a5d8f73e273f2d03c9d86ea5662cda481a/src/core/instance/lifecycle.js#L315

### DIFF
--- a/test/specs/mounting-options/sync.spec.js
+++ b/test/specs/mounting-options/sync.spec.js
@@ -1,3 +1,4 @@
+import sinon from 'sinon'
 import { describeWithShallowAndMount } from '~resources/utils'
 
 describeWithShallowAndMount('options.sync', (mountingMethod) => {
@@ -46,7 +47,7 @@ describeWithShallowAndMount('options.sync', (mountingMethod) => {
             <pre>computed.text: <em>{{ computedText }}</em></pre>
           </div>
         </div>
-        </div>
+      </div>
       `,
       data () {
         return {
@@ -109,5 +110,26 @@ describeWithShallowAndMount('options.sync', (mountingMethod) => {
       expect(wrapper.text()).to.equal('world')
       done()
     })
+  })
+
+  it('call updated when sync is not false', () => {
+    const spy = sinon.stub()
+    const TestComponent = {
+      template: '<div>{{ foo }}</div>',
+      data () {
+        return {
+          foo: 'foo'
+        }
+      },
+      updated () {
+        spy()
+      }
+    }
+    const wrapper = mountingMethod(TestComponent, {
+      sync: true
+    })
+    expect(spy.notCalled).to.equal(true)
+    wrapper.vm.foo = 'bar'
+    expect(spy.calledOnce).to.equal(true)
   })
 })
--- a/test/specs/mounting-options/sync.spec.js
+++ b/test/specs/mounting-options/sync.spec.js
@@ -113,9 +113,17 @@ describeWithShallowAndMount('options.sync', (mountingMethod) => {
   })
 
   it('call updated when sync is not false', () => {
+    const fooSpy = sinon.stub()
+    const Foo = {
+      template: '<div>{{ foo }}</div>',
+      props: ['foo'],
+      updated () {
+        fooSpy()
+      }
+    }
     const spy = sinon.stub()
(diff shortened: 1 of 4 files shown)


eddyerburgh:
This is a good solution, but I'm not sure we should keep it long term.

The recent changes to run watchers synchronously has caused lots of bugs.

I have a PR open that will add an async option to Vue, so we can set that to make rendering work synchronously—https://github.com/vuejs/vue/pull/8240

I think the solution when we have the async option is to remove the code that sets watchers to sync, because there are too many bugs. Instead we should warn that users with Vue < 2.6 must either update Vue or use Vue.nextTick for updates to the DOM. 

One question with this PR—I don't think child components will call `update` with this solution?

38elements:
Thank you for reply.
`sync` option is useful.
Since `sync` option is different from the original behavior of Vue,
I think that it is better to remove `sync` option.
Since if `sync` option is removed, I think it is not necessary to resolve #455,
I will remove code which resolve #455 and keep code which resolve #582.

> One question with this PR—I don't think child components will call update with this solution?

I think `update` method of child component is replaced at below line.
I will add test.

https://github.com/vuejs/vue-test-utils/pull/661/files#diff-81317be9320c4ebacbb2460e0f504e58R28

test/specs/mounting-options/sync.spec.js
```diff
@@ -126,10 +134,14 @@ describeWithShallowAndMount('options.sync', (mountingMethod) => {
...
+      stubs: { foo: Foo },
       sync: true
     })
     expect(spy.notCalled).to.equal(true)
+    expect(fooSpy.notCalled).to.equal(true)
     wrapper.vm.foo = 'bar'
     expect(spy.calledOnce).to.equal(true)
+    expect(fooSpy.calledOnce).to.equal(true)
+    expect(wrapper.html()).to.equal('<div>bar<div>bar</div></div>')
   })
```
38elements: @eddyerburgh 
> One question with this PR—I don't think child components will call update with this solution?

Since this test passed,  child components are updated.

38elements:
I will remove code which resolve #455 and keep code which resolve #582 within 3 days.

eddyerburgh:
I think this issue will be fixed in #586

38elements:
Since this change will be unnecessary, I close this pull request.

38elements:
@eddyerburgh 

I am sorry.
I change my opinion.
I think `sync` mounting option is necessary and should not be removed.
`sync` mounting option is useful.

It is not problem  that `sync` mounting option is different from the original behavior of Vue.
It should be explicitly described in document  that `sync` mounting option is different from the original behavior of Vue.

If the `async` option is added to Vue@2.6, it is not added to Vue < 2.6.
In this case, It is inconvenient to run the same test by different Vue versions.
At Vue@2.6+, it is possible to use async option is false.
At Vue < 2.6, it is impossible to synchronously test .
Or The expression in the test will become different depending on the Vue version.
I think it is better to abstract the difference between `async` option and existing `sync` mounting option by `sync` mounting option.

eddyerburgh:
Ok, can you reimplement your original code for update?

38elements:
@eddyerburgh 
Thank you for reply.
I send pull request #675.

PR #674, opened 2018-06-01
docs: keep update

Jinjiang:
* Removed some refs from `wrapper/*.md`to `../wrapper/` because they are on the same page actually. @eddyerburgh @38elements   
* Chinese update @fnlctrl 

Thanks.


38elements:
Hi, @Jinjiang.

IMHO,
Since the links in `find.md` and `findAll.md` transfer top of page,
I think it is better not to remove them.

38elements:
Transitioning from the right side index, it displays separate page.

https://vue-test-utils.vuejs.org/

https://vue-test-utils.vuejs.org/api/wrapper/find.html

Jinjiang:
Sorry for late. 😅

> Transitioning from the right side index, it displays separate page.
> https://vue-test-utils.vuejs.org/
> https://vue-test-utils.vuejs.org/api/wrapper/find.html

Yes, I also find it. I am a little confused about offering https://vue-test-utils.vuejs.org/api/wrapper/#find-selector and https://vue-test-utils.vuejs.org/api/wrapper/find.html at the same time. I guess the second one is not on purpose since no other link can go there and the relative URLs of links there goes wrong @eddyerburgh 🤔

Jinjiang:
> IMHO,
> Since the links in find.md and findAll.md transfer top of page,
> I think it is better not to remove them.

And that's another thing confused for me because not every API headings has a link to top. Only `find()` and `findAll()` have them. So I removed them.
Or we add link to top for every API headings but it's not necessary imo.

Thanks :-)

eddyerburgh:
Yeah I think we can still work on the layout. I wanted to include the name of the arguments, but the URL is autogenerated. If either of you have suggestions for improving the docs I'm open to suggestions!

38elements:
>IMHO,
Since the links in find.md and findAll.md transfer top of page,
I think it is better not to remove them.

Thank you for reply.
The top of page is Wrapper section.
I think it is appropriate to transfer to Wrapper section by clicking Wrapper link.

https://vue-test-utils.vuejs.org/api/wrapper/#properties

38elements:
@Jinjiang 
I am sorry.
I changed my opinion.
I think that it is not important that removing  or not links in `find.md` and `findAll.md`.
I think that you will maintain this document longer than me.
I think that your opinion should be respected.

38elements (approved):

eddyerburgh (approved):
Thanks 😄

fnlctrl (approved):

Issue #651, opened 2018-05-24
Improving `find` and `findAll` API for asserting presence of stubs 


lmiller1990:
### What problem does this feature solve?
If I want to assert a stub was rendered, I have to do the following:

Demo component:

```html
<!-- component.vue -->
<template>
  <div>
    <Hello />
  </div>
</div>
```

I want to assert `Hello` is rendered one time.

```js
// test
const HelloStub = {
  'name': "Hello"
  render: h => h('div')
}

const wrapper = shallow(Component}

expect(wrapper.find(HelloStub).exists()).toBe(true)
```

I always need to create a minimal:

```js
const HelloStub = { name: "Hello", render: h => h("div") }
```
That is an extra (correctly formatted) four lines - over half the test. 

My current workflow simply defines a helper:

```js
const createStub = name => ({ name, render: h => h("div") })
```

Which adds an extra line for the import, and another for create the stub in every test. 1/3 of the test is, in my opinion, unnecessary boilerplate setup.

### What does the proposed API look like?
I find myself defining and importing the above helper function in every project, and importing it to many tests. One idea would be to add this method to `vue-test-utils`. This goes against one design goal, "keeping the API surface small and straightforward". However, it supports one other goal, "The problem should also be common enough to justify the addition". I think asserting components are rendered is done is many tests.

I had two proposed APIs:

1. include helper in `vue-test-utils`

```js
import { shallowMount, createStub } from "@vue/test-utils" 

const Hello = createStub("Hello")
const wrapper = shallowMount(Component)

expect(Component.find(Hello).exists()).toBe(true)
```

2. passing a string to `find`, and `findAll` checks for stubs

```js
import { shallowMount } from "@vue/test-utils" 

const wrapper = shallowMount(Component)

expect(Component.find("Hello").exists()).toBe(true)
```

Since Vue components cannot clash with existing HTML elements, it is easy to know when `find` receives a HTML element like `<div>`, and when it's looking for an actual component. In the case of an a actual component, `find` should look for stubs instead of html elements.

I like the second solution more because it leads to more compact code, and doesn't change overall API very much.

I am happy to implement either.

eddyerburgh:
I think a `createStub` utility would probably be useful.

The second option is a good idea. I think users would find this more intuitive than using a component as a selector.

You could match a vNode using the `tag` property in the componentOptions object. 

Although in your example you can assert that Hello is rendered using the component as a selector:

```js
const wrapper = mount(Component)
[...]

Issue #658, opened 2018-05-25
Component stubs should render default slots


trollepierre:
### Version
1.0.0-beta.16

### Reproduction link
[http://google.com](http://google.com)

### Steps to reproduce
```
<template>
  <v-layout>
    My text
  </v-layout>
</template>

<script>

  export default {
    name: 'Component',
  }
</script>
```

```
import { createLocalVue, shallowMount } from '@vue/test-utils'
import Vuetify from 'vuetify'
import Vuex from 'vuex'
import Component from './Component.vue'

describe('components | Component', () => {
  let localVue

  beforeEach(() => {
    localVue = createLocalVue()
    localVue.use(Vuetify)
    localVue.use(Vuex)
  })

  describe('template', () => {
    it('should match snapshot', () => {
      // When
      const wrapper = shallowMount(Component, { localVue })

      // Then
      expect(wrapper.element).toMatchSnapshot()
    })
  })
})
```
___
When I comment  ```   localVue.use(Vuex)``` , it shouldn't affect my snapshot, because I am not using any store in my component

### What is expected?
Green test

### What is actually happening?
Jest Snapshot result:
```

- <div
  class="layout"
>
  
    My text

</div>
  <!---->
```

---
could you provide a template of CodePen, CodeSandbox if you want a link, please?

eddyerburgh:
Can you add a reproduction on CodeSandBox—https://codesandbox.io/s/m4qk02vjoy?

trollepierre:
Yes :  https://codesandbox.io/s/vvk1pw1w3l
in TestComponent.spec.js, I commented 
localVue.use(Vuex)
and the snapshot test fails

But I am not using Vuex inside this component

lmiller1990:
I don't think this is a bug, @trollepierre . I was able to get your test passing using this code:

```js
import { createLocalVue, mount } from '@vue/test-utils'
import TestComponent from "./TestComponent";
import Vuetify from 'vuetify'

describe('components | Component', () => {
  // let localVue

  beforeEach(() => {
    localVue = createLocalVue()
    localVue.use(Vuetify)
    //localVue.use(Vuex)
  })

  describe('template', () => {
    it('should match snapshot', () => {
      // When
      const wrapper = mount(TestComponent)

      // Then
      expect(wrapper.vm.$el).toMatchSnapshot()
    })
  })
})
```

What I did 

[...]

PR #678, opened 2018-06-03
add `createStub` function

lmiller1990:
Regarding #651 

This PR adds a `createStub` method, which creates a component as such:

```js
const MyComp = createStub("MyComp")

MyComp //=>{ name: "MyComp", render: h => h("div") }
```
Useful for making it clear what is going on when you call `find` and `findAll`.

### DIFF
--- a/docs/api/createStub.md
+++ b/docs/api/createStub.md
@@ -1,0 +1,31 @@
+## createStub()
+
+- **Arguments:**
+
+  - `{String} name`
+
+- **Returns:**
+  - `{Component}`
+
+- **Usage:**
+
+Creates a stub component. Useful when combined with [`find`](wrapper/find)  and [`findAll`](wrapper/findAll).
+
+```js
+import { shallowMount, createStub } from '@vue/test-utils'
+import ComponentWithChild from './ComponentWithChild.vue'
+
+describe('ComponentWithChild', () => {
+  it('renders a child component', () => {
+    const Child = createStub('Child')
+    const wrapper = shallowMount(ComponentWithChild, {
+      stubs: {
+        Child: Child
+      }
+    })
+    expect(wrapper.findAll(Child).length).toBe(1)
+  })
+})
+```
+
+Where `Child` is a component with `name: 'Child'`.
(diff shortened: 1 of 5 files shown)


docs/api/createStub.md
```diff
@@ -0,0 +1,31 @@
...
+      stubs: {
+        Child: Child
+      }
+    })
+    expect(wrapper.findAll(Child).length).toBe(1)
+  })
+})
+```
+
+Where `Child` is a component with `name: 'Child'`.
```
eddyerburgh, line 31: I don't think the component needs to have the name child. It should use the constructor. Can you check if the example works for a component without a name?

packages/test-utils/src/create-stub.js
```diff
@@ -0,0 +1,7 @@
+export default function createStub (name) {
```
38elements, line 1: I think that it is better to add flow type as below.

```js
export default function createStub (name: string): Component {
```

eddyerburgh: Yes, and add `// @flow` to the top of the file

lmiller1990:
Got caught up, hopefully by tomorrow can get this done. For TODO:

- [ ] add test to use constructor, not name
- [ ] add flow to `create-stub`
- [ ] render $slot.default ? for #658

eddyerburgh:
Yes the slot default is a good idea. If you'd like you could also add this feature to the default stub rendered in the add-stubs file, so that all default stubs render $slots.

38elements:
Would you please tell me  difference between this feature and  a string value in `stubs` mounting options when the default stub become supporting `$slot.default`?

eddyerburgh:
@38elements You can use these stubs for unnamed components without importing the component:

```js
const SomeStub = createStub()
const wrapper = mount(TestComponent, {
  SomeComponent: SomeStub
})

wrapper.find(SomeStub)
```

Although this will become irrelevant with #651 🤔

lmiller1990:
Should we just close this since it is more or less super-seeded by #651 ?

PR #685, opened 2018-06-06
fix type definition of classes method (fix #684)

### LINKED ISSUE #684: Typescript error when use wrapper.classes().includes
### Version
1.0.0-beta.16

### Reproduction link
[https://github.com/Yawenina/vue-test-utils-classes](https://github.com/Yawenina/vue-test-utils-classes)

### Steps to reproduce
```
npm install

npm run serve
```

### What is expected?
no error

In my [test file](https://github.com/Yawenina/vue-test-utils-classes/blob/master/tests/unit/HelloWorld.spec.ts#L10), I write a test like this:
```
expect(wrapper.classes().includes('a')).toBe(false);
```
but when I ran `npm run serve`, typescript yield an error:
```
ERROR in vue-test-utils-classes/tests/unit/HelloWorld.spec.ts
10:30 Property 'includes' does not exist on type 'void | string[]'.
  Property 'includes' does not exist on type 'void'.
     8 |       propsData: { msg },
     9 |     });
  > 10 |     expect(wrapper.classes().includes('a')).toBe(false);
       |                              ^
    11 |   });
    12 | });
    13 |
Version: typescript 2.9.1
```

---
I read the [source code](https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/types/index.d.ts#L59) and find the type definition of the `classes()` method is this:

```
classes(): Array<string> | void
```

Maybe because the return value could be void so it does not have a `includes` method.

When I change it to :
```
classes(): Array<string>
```
the error disappears and everything works ok.

Yawenina:
This is related to [#684](https://github.com/vuejs/vue-test-utils/issues/684)

### DIFF
--- a/packages/test-utils/types/index.d.ts
+++ b/packages/test-utils/types/index.d.ts
@@ -56,7 +56,7 @@ interface BaseWrapper {
   visible (): boolean
 
   attributes(): { [name: string]: string }
-  classes(): Array<string> | void
+  classes(): Array<string>
   props(): { [name: string]: any }
 
   hasAttribute (attribute: string, value: string): boolean


eddyerburgh (approved):
Thanks 👍

Issue #631, opened 2018-05-20
"Avoid mutating a prop directly" warning when updating prop


aphofstede:
### Version
1.0.0-beta.16

### Reproduction link
[https://codesandbox.io/s/vnv9wzp8rl](https://codesandbox.io/s/vnv9wzp8rl)

### Steps to reproduce
See repro CodeSandbox.

- Create a minimal component (HelloWorld)
- Create another component (base component MobileCollapseRow) that takes two props, uses one of them and includes the minimal component
- Create the final component (ContractRow) that uses the base component and passes it the prop used above, with a computed prop as value, based on its own prop
- Now create a test that passes in a new value for the prop in the last component and observe the console.

### What is expected?
Prop is updated and value changes without warnings

### What is actually happening?
An warning is logged, complaining about changing a prop that isn't actually touched at all:

[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "actions"

found in

---> <MobileCollapseRow> at src/components/MobileCollapseRow.vue
       <Root>

---
Note that the test in de CodeSandbox example does succeed, but keep an eye on the console tab.

Note that the warning complains about *another* prop! Something fishy is going on here..

Note: Removing the child component (HelloWorld) from the base component (MobileCollapseRow) makes the warning disappear.

Note: Removing the `<a>` where the prop is used from the base component makes the warning disappear.

eddyerburgh:
Thanks for the bug report.

This is caused by a recent change to handling asynchronous updates. This should be fixed in the coming weeks.

aphofstede:
@eddyerburgh Thanks

cdbkr:
Just wondering if it makes sense to extend the configuration of test-utils and enable/disable the warnings. It might control the Vue.config.silent when creating the instance. Would it be a correct solution?

eddyerburgh:
Yes I think that's the only solution at the moment.

We would need to make sure no important warnings are hidden. 

Also we would need to set the config to silent in all methods that interact with the instance, like `setProps` and `trigger`.

aphofstede:
Does that mean Vue is complaining about a non-issue here, or just that you want the option to get rid of pesky logging? :)

eddyerburgh:
Sort of. 

The problem is how we implemented synchronous updates, you can read an explanation here—https://github.com/vuejs/vue-test-utils/issues/676.

[...]

PR #688, opened 2018-06-07
#631 - Silent "Avoid mutating a prop directly" warning when updating prop

cdbkr:
This PR aims to solve #631 

By default, Vue is throwing warnings when setProps is used/invoked, due to changes to observers and "Vue.config.silent" is false by default https://vuejs.org/v2/api/#silent

Since "silent" suppress all warnings and logs, it would be great to reduce the scope just to the desired functions/blocks.

Please, take this as an initial proposal and discuss if it makes sense to adopt this strategy and feel free to point missing implementation.

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -94,3 +94,17 @@ import VueTestUtils from '@vue/test-utils'
 
 VueTestUtils.config.logModifiedComponents = false
 ```
+
+### `silentWarnings`
+
+- type: `Boolean`
+- default: `true`
+
+It suppresses warnings triggered by Vue while mutating component's observables (e.g. props). When set to `false`, all warnings are visible in the console. This is a configurable way which relies on `Vue.config.silent`.
+Example:
+
+```js
+import VueTestUtils from '@vue/test-utils'
+
+VueTestUtils.config.silentWarnings = false
+```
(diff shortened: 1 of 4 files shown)


eddyerburgh:
Thanks for the PR!

I think this is fine. 

The only negative I can see is that it would silence any true errors, so when somebody is trying to change a prop that doesn't exist. But at the moment, there are so many errors that you can't tell the difference anyway.

cdbkr:
Thanks to you!

Your point makes total sense to me and we don't want to suppress valid errors (which can prevent bugs).

Anyway, when changing props which don't exist, aren't we expecting an exception instead of a console error?
https://github.com/vuejs/vue-test-utils/blob/dbf63bb3c6cb06f45713c4d2a98d45b52149a14d/test/specs/wrapper/setProps.spec.js#L30
I added a test case locally and the exception is thrown when changing the not existing prop (when silentWarnings is set to true).

Also, is there any missing part which would use the same logic? (if yes, maybe it would be great to discuss/consider decorators instead)

eddyerburgh:
Ah great, I forgot we had that functionality!

I can't see any other methods currently where we need to set the config to silent.

cdbkr:
Great, if you agree, I'm going to include that test case covering the scenario you described (in case of logic changes) and write some documentation regarding the introduced property "silentWarnings".

eddyerburgh:
Yep sounds good

eddyerburgh (approved):
Thanks 👍 

I'll release this evening

PR #691, opened 2018-06-08
docs: update Vuex guide for namespaced option

cpxPratik:


docs/guides/using-with-vuex.md
```diff
@@ -214,10 +214,10 @@ Simple component that includes one action and one getter.
 And the test:
 
 ``` js
-import { shallowMount, createLocalVue } from '@vue/test-utils'
+import { createLocalVue, shallowMount } from '@vue/test-utils'
 import Vuex from 'vuex'
 import MyComponent from '../../../src/components/MyComponent'
-import mymodule from '../../../src/store/mymodule'
+import myModule from '../../../src/store/myModule'
```
lmiller1990: I like this, fits the common convention of camel casing modules

docs/guides/using-with-vuex.md
```diff
@@ -241,10 +239,11 @@ describe('MyComponent.vue', () => {
 
     store = new Vuex.Store({
       modules: {
-        mymodule: {
+        myModule: {
+          namespaced: true, // If store/myModule is namespaced https://vuex.vuejs.org/guide/modules.html#namespacing
```
lmiller1990: I don't think this is necessary. I think the examples should have as little noise as possible. Since the user is testing their modules, we can assume they know if they namespaced it or not. Since the focus of this guide and repo is the test setup/assertions, not Vuex and the Vuex API, I don't think this should be included.

lmiller1990 (commented):
Some feedback, thanks for the contribution!

eddyerburgh:
Thanks for the PR, I agree with @lmiller1990 

Could you also update the example repo to match your changes—https://github.com/eddyerburgh/vue-test-utils-vuex-example

cpxPratik:
Updated based on comment, also there is a PR on the mentioned example repo to reflect changes.

lmiller1990 (approved):

eddyerburgh (approved):
Thanks

PR #717, opened 2018-06-14
feat: render default slot in stubs

eddyerburgh:
Render slots in default stubs.

Breaking change

### DIFF
--- a/packages/shared/stub-components.js
+++ b/packages/shared/stub-components.js
@@ -65,7 +65,9 @@ function createBlankStub (originalComponent: Component) {
   return {
     ...getCoreProperties(originalComponent),
     render (h) {
-      return h(`${originalComponent.name}-stub`)
+      return h(`${originalComponent.name}-stub`,
+        !originalComponent.functional && this.$slots.default
+      )
     }
   }
 }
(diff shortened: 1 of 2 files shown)


lmiller1990 (commented):
Looks good to me - question, what impact will this have on the public api?

What is the procedure for a breaking change like this? Adding a notice to people using the version prior to the breaking changes?

eddyerburgh:
It's a breaking change because anyone who relies on shallowMount not including slot children will have to change their code.

Because VTU is still in beta, we can release a breaking change in the next version. Once VTU is 1.x we will only add breaking changes in major versions.

packages/shared/stub-components.js
```diff
@@ -79,7 +79,9 @@ function createBlankStub (originalComponent: Component) {
   return {
     ...getCoreProperties(originalComponent),
     render (h) {
-      return h(name)
+      return h(`${originalComponent.name}-stub`,
```
38elements, line 82: I think this line does not need to change.

```js
const name = `${originalComponent.name}-stub`
```

eddyerburgh:
Closed in favor of https://github.com/vuejs/vue-test-utils/pull/782

PR #720, opened 2018-06-15
refactor: add max-len rule

eddyerburgh:


38elements:
IMHO, I think it is better that `code` option is 120.
https://eslint.org/docs/2.0.0/rules/max-len#options

eddyerburgh:
I think we should use the default, prettier also uses 80 as a default. I find it easier to read 80 than 120.

Note there isn't a max-len in the test directory.

eddyerburgh:
We could add a limit of 120 in the test directory, which I think needs a longer limit for the spec names

PR #724, opened 2018-06-16
fix: set wrapper.vm if the element binds Vue instance

38elements:
This is related to #687.

### DIFF
--- a/packages/test-utils/src/create-wrapper.js
+++ b/packages/test-utils/src/create-wrapper.js
@@ -7,7 +7,10 @@ import VueWrapper from './vue-wrapper'
 export default function createWrapper (
   node: VNode | Component,
   options: WrapperOptions
-) {
+): VueWrapper | Wrapper {
+  if (node.componentInstance) {
+    return new VueWrapper(node.componentInstance, options)
+  }
   return node instanceof Vue
     ? new VueWrapper(node, options)
     : new Wrapper(node, options)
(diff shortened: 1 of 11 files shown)


eddyerburgh:
Can you check in createWrapper if there is a componentInstance, and if there is return a wrapper using componentInstance?

```
export default function createWrapper (
  node: VNode | Component,
  options: WrapperOptions
) {
  if (node.componentInstance) {
    return new VueWrapper(node.componentInstance, options)
  }
  return node instanceof Vue
    ? new VueWrapper(node, options)
    : new Wrapper(node, options)
}
```

38elements:
Thank you for reviewing.
I changed it.

eddyerburgh:
If that works then I think you can remove the extra code in this PR?

38elements:
No, it is not.
I think that it is necessary to get a component instance from `vm.$children`.

38elements:
I think that the following code is not necessary if this change code exists. 

```js
  if (node.componentInstance) {
    return new VueWrapper(node.componentInstance, options)
  }
```

eddyerburgh:
Why is it necessary to get the instance from vm.$children?

I thought we could keep the same logic for getting vnodes. Then we could check that if the vnode is for a component instance in the create-wrapper function.

38elements:
I misunderstood.
This passed at 2.5.
I will fix it.

38elements:
I fixed it.
`VNode.child` was changed to `VNode.componentInstance` at `2.1.9`.
https://github.com/vuejs/vue/commit/ced2dfc8f704b07cd59467eca38c58c590d0c012

eddyerburgh (approved):

PR #726, opened 2018-06-16
fix: added quotes in warn message

jacekkarczmarczyk:
When the component has a name "test" or something similar the generated error message `an extended child component test has been modified` is not very clear at the first glance


jacekkarczmarczyk:
Will make a proper PR after cloning repo

PR #727, opened 2018-06-16
refactor: silentWarnings -> silent

38elements:
~~This add `silentWarnings` type.~~
 This renames `silentWarnings` to `silent`.


eddyerburgh:
I know this isn't part of the PR, but I actually think we should rename `silentWarnings` to `silent` to keep the same API as Vue.

Can you make that change in this PR?

38elements:
Thank you for reply.
I changed them.

eddyerburgh (approved):
Thanks

PR #731, opened 2018-06-17
fix: more readable log modified components warning

jacekkarczmarczyk:
When the component has a name "test" or something similar the generated error message 

`an extended child component test has been modified`

 is not very clear at the first glance. I've changed it to

`an extended child component <test> has been modified`

(this is also the way the component names are displayed for example here https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/components/TransitionStub.spec.js#L60)


eddyerburgh (approved):
Thanks 👍

PR #733, opened 2018-06-18
fix(slots): functional component text slots

### LINKED ISSUE #732: Slots with text contents are not properly resolved in functional components.
### Version
1.0.0-beta.19

### Reproduction link
[https://codesandbox.io/s/ollqlyxv2y](https://codesandbox.io/s/ollqlyxv2y)

### Steps to reproduce
In beta.19, template component slots can be filled with strings easily. But functional components still cannot.

* Create a functional component with a slot (can be default or named)
* Create a test suite which fills the slot with a simple string
* See the `<div></div>` problem again.

### What is expected?
The slot should be properly filled with the string

The `<div></div>` problem is occuring again

Related: #718, #707, #711

TheAlexLichter:
Resolves #732

### DIFF
--- a/packages/create-instance/add-slots.js
+++ b/packages/create-instance/add-slots.js
@@ -30,12 +30,10 @@ export function createSlotVNodes (
   return Object.keys(slots).reduce((acc, key) => {
     const content = slots[key]
     if (Array.isArray(content)) {
-      const nodes = content.reduce((accInner, slotDef) => {
-        return accInner.concat(createVNodesForSlot(h, slotDef, key))
-      }, [])
+      const nodes = content.map(slotDef => createVNodesForSlot(h, slotDef, key))
       return acc.concat(nodes)
-    } else {
-      return acc.concat(createVNodesForSlot(h, content, key))
     }
+
+    return acc.concat(createVNodesForSlot(h, content, key))
   }, [])
 }
--- a/packages/create-instance/create-functional-component.js
+++ b/packages/create-instance/create-functional-component.js
@@ -1,39 +1,8 @@
 // @flow
 
-import { compileToFunctions } from 'vue-template-compiler'
 import { throwError } from 'shared/util'
 import { validateSlots } from './validate-slots'
-
-function createFunctionalSlots (slots = {}, h) {
-  if (Array.isArray(slots.default)) {
-    return slots.default.map(h)
-  }
-
-  if (typeof slots.default === 'string') {
-    return [h(compileToFunctions(slots.default))]
-  }
-  const children = []
-  Object.keys(slots).forEach(slotType => {
-    if (Array.isArray(slots[slotType])) {
-      slots[slotType].forEach(slot => {
-        const component =
-          typeof slot === 'string' ? compileToFunctions(slot) : slot
-        const newSlot = h(component)
-        newSlot.data.slot = slotType
-        children.push(newSlot)
-      })
-    } else {
-      const component =
-        typeof slots[slotType] === 'string'
-          ? compileToFunctions(slots[slotType])
-          : slots[slotType]
-      const slot = h(component)
-      slot.data.slot = slotType
-      children.push(slot)
-    }
-  })
-  return children
-}
+import { createSlotVNodes } from './add-slots'
 
 export default function createFunctionalComponent (
   component: Component,
@@ -56,7 +25,7 @@ export default function createFunctionalComponent (
           mountingOptions.context.children.map(
             x => (typeof x === 'function' ? x(h) : x)
           )) ||
-          createFunctionalSlots(mountingOptions.slots, h)
+          createSlotVNodes(h, mountingOptions.slots || {})
       )
     },
     name: component.name,
--- a/flow/options.flow.js
+++ b/flow/options.flow.js
@@ -4,7 +4,7 @@ declare type Options = {
   propsData?: Object,
   mocks?: Object,
   methods?: Object,
-  slots?: Object,
+  slots?: SlotsObject,
   scopedSlots?: Object,
   localVue?: Component,
   provide?: Object,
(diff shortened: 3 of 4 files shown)


packages/server-test-utils/dist/vue-server-test-utils.js
```diff
@@ -19,14 +19,12 @@ function createVNodesForSlot (
   slotValue,
   name
 ) {
-  if (typeof slotValue === 'string' &&
-  !startsWithTag(slotValue)) {
+  if (typeof slotValue === 'string' && !startsWithTag(slotValue)) {
```
jacekkarczmarczyk: These changes are not related to the issue, files are automatically updated on `yarn test` or something, but they shouldn't be included in the PR

packages/create-instance/create-functional-component.js
```diff
@@ -4,35 +4,35 @@ import { compileToFunctions } from 'vue-template-compiler'
 import { throwError } from 'shared/util'
 import { validateSlots } from './validate-slots'
 
-function createFunctionalSlots (slots = {}, h) {
-  if (Array.isArray(slots.default)) {
-    return slots.default.map(h)
-  }
+function startsWithTag (str) {
```
jacekkarczmarczyk: `startsWithTag` is also defined in add-slots.js, I guess it could be extracted to some helper

packages/test-utils/dist/vue-test-utils.js
```diff
@@ -17,7 +17,10 @@ function warn (msg) {
 
 var camelizeRE = /-(\w)/g;
 var camelize = function (str) {
-  var camelizedStr = str.replace(camelizeRE, function (_, c) { return c ? c.toUpperCase() : ''; });
```
jacekkarczmarczyk: These changes are not related to the issue, files are automatically updated on yarn test or something, but they shouldn't be included in the PR

packages/server-test-utils/dist/vue-server-test-utils.js
TheAlexLichter: Whoops! Thanks :relaxed:

packages/create-instance/create-functional-component.js
TheAlexLichter: Right! Actually, we could use the `add-slots` files as well.

Let me refactor it :+1:

TheAlexLichter:
@jacekkarczmarczyk I've refactored it. But I've encountered a strange behavior:

Currently, the implementation uses the `createFunctionalSlotVNodes` function. This is just a "redirect" ([see here](https://github.com/vuejs/vue-test-utils/pull/733/commits/31a3d0bf110f6d52e803a7b57627f87fa1a8d475#diff-2aef106d4e1902310b282815a009490aR41)):

```
export function createFunctionalSlotVNodes (h, slots = {}) {
  return createSlotVNodes(h, slots)
}
```

However, if I replace all calls with the `createSlotVNodes` function, karma throws several errors. 

Can't explain the behavior to me:man_shrugging: 
Do you have any idea?

EDIT: And flow is failing because of types

jacekkarczmarczyk:
You've overwritten https://github.com/vuejs/vue-test-utils/commit/bc736fb6431e54ba5a9cf7908044e5c71234facc ...

TheAlexLichter:
@jacekkarczmarczyk Dang! :roll_eyes:  Still too early for me :see_no_evil: 

Should look better now...

TheAlexLichter:
Okay, one more try. I think I got it now. Test running locally and flow doesn't complain anymore :yum:

packages/create-instance/add-slots.js
```diff
@@ -30,12 +30,17 @@ export function createSlotVNodes (
...
       return acc.concat(nodes)
-    } else {
-      return acc.concat(createVNodesForSlot(h, content, key))
     }
+
+    return acc.concat(createVNodesForSlot(h, content, key))
   }, [])
 }
+
+export function createFunctionalSlotVNodes (
```
eddyerburgh: No need to add this function, we can just use `createSlotVNodes`

packages/create-instance/add-slots.js
```diff
@@ -6,7 +6,7 @@ function startsWithTag (str) {
   return str && str.trim()[0] === '<'
 }
 
-function createVNodesForSlot (
+export function createVNodesForSlot (
```
eddyerburgh: I don't think we need to export

packages/create-instance/create-functional-component.js
```diff
@@ -1,39 +1,8 @@
...
-          ? compileToFunctions(slots[slotType])
-          : slots[slotType]
-      const slot = h(component)
-      slot.data.slot = slotType
-      children.push(slot)
-    }
-  })
-  return children
-}
+import { createFunctionalSlotVNodes } from './add-slots'
```
eddyerburgh: Can you use createSlotVNodes instead?

eddyerburgh (changes requested):
Thanks for the PR, looks good although I think we can use the existing `createSlotVnodes` function instead of the new function

edit: I've just seen your comments about karma. Can you post the errors?

packages/create-instance/create-functional-component.js
TheAlexLichter: I tried this, but it gave me some strange errors.

See https://github.com/vuejs/vue-test-utils/pull/733#issuecomment-397980897

TheAlexLichter: Update: Looks like this had sth. to do with wrong types. It works fine right now :man_shrugging:

TheAlexLichter:
@eddyerburgh I think those were related to flow issues as everything works fine now :relaxed: 

I've updated the PR. Hope the tests will pass like they do locally :+1:

eddyerburgh (approved):
Great, thanks 👍 

A PR that removes more code than it adds, AND fixes a bug is always good!

TheAlexLichter:
@eddyerburgh  You are welcome :relaxed: 

Cannot agree more on that! :speak_no_evil:

TheAlexLichter:
@eddyerburgh Do you know when the change will be released? :)

eddyerburgh:
I'll release within a week

PR #734, opened 2018-06-18
[WIP] Fix named functional text slots

TheAlexLichter:
After #733, default text-slots work quite well. But named slots are awkwardly merged in the default slot variable. I'm working on a fix.

### DIFF
--- a/test/specs/mounting-options/slots.spec.js
+++ b/test/specs/mounting-options/slots.spec.js
@@ -221,7 +221,7 @@ describeWithMountingMethods('options.slots', mountingMethod => {
     }
   })
 
-  it('mounts component with text slot', () => {
+  it('mounts component with named and default text slot', () => {
     const wrapper = mountingMethod(ComponentWithSlots, {
       slots: {
         default: 'hello,',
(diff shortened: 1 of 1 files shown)


TheAlexLichter:
I've looked into template compilation with `Vue.compile`.

If you want to achieve the same in template mode (without a render function), you'd do:
```
<template>
  <template slot='hey'>
    Hi
  </template>
<template>
```

This will be compiled to

 `with(this){return [_c('template',{slot:"hey"},[_v("Hi")])]}`

So we could leverage the `_c` and `_v` method to achieve it, but I'm unsure if we should do this :thinking:

eddyerburgh:
Yes we could use _v, but it would be better if we could construct with h.

eddyerburgh:
We could compile the slot using that template:

```compileToFunctions(`<template slot="${name}">${slotValue}</template>`)``` (https://github.com/vuejs/vue-test-utils/blob/dev/packages/create-instance/add-slots.js#L15)

**Comment by TheAlexLichter** [comment]
Nice idea but I've realized that we can use neither `template` nor `slot` as root element:

> Cannot use template as component root element because it may contain multiple nodes.

I've tried it with text only but this also fails:

>  Component template requires a root element, rather than just text.

**Comment by eddyerburgh** [comment]
Ok, we could render it in a parent div, then use the child vNode when we call `h`
```
const node  = compileToFunctions(`<template slot="${name}">${slotValue}</template>`)
h(node).children[0]
```

**Comment by eddyerburgh** [comment]
@manniL did you try the suggested approach ^^

**Comment by TheAlexLichter** [comment]
@eddyerburgh  Not yet. Will look into it in the next days!

**Comment by eddyerburgh** [comment]
After thinking about this, I don't think we should add this feature.

Since named slots cannot take root text nodes, we shouldn't support them. Instead you can wrap them in a template:

```js
mount(TestComponent, {
  slots: {
    footer: '<template>hey</template>'
  }  
})
```

PR #737, opened 2018-06-19
docs: tweaks

38elements:


docs/api/wrapper-array/setData.md
```diff
@@ -1,6 +1,6 @@
 ## setData(data)
 
-Sets `Wrapper` `vm` data and forces update on each `Wrapper` in `WrapperArray`.
+Sets `Wrapper` `vm` data by each `Wrapper` in `WrapperArray`.
```
eddyerburgh: `by` should be `on`

```data on each `Wrapper````

eddyerburgh (approved):
Thanks 👍

docs/api/wrapper-array/setData.md
38elements: Thank you for pointing out.

PR #742, opened 2018-06-20
Update warning doc links

andreasvirkus:
Noticed these broken links in the warnings and thought I'd help you out in fixing them @eddyerburgh.
Not sure if the `.html` extension is preferred or if that should be prettified in the Netlify's redirects file?

### DIFF
--- a/docs/api/wrapper/attributes.md
+++ b/docs/api/wrapper/attributes.md
@@ -1,4 +1,4 @@
-### attributes()
+## attributes()
 
 Returns `Wrapper` DOM node attribute object.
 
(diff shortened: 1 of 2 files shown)


eddyerburgh:
Thanks, the # is preferred:

`https://vue-test-utils.vuejs.org/api/wrapper/#methods`

I just realized that attributes doesn't have a #

Would you also be able to update https://github.com/vuejs/vue-test-utils/blob/dev/docs/api/wrapper/attributes.md to use a h2 instead of h3. Then the URL https://vue-test-utils.vuejs.org/api/wrapper/#attributes will work.

eddyerburgh (approved):
Thanks. The trigger link is also incorrect, but that's an issue with the docs that we'll fix separately.

eddyerburgh:
You need to update the tests to use the new URLs

eddyerburgh:
This has been fixed in https://github.com/vuejs/vue-test-utils/pull/786/files. Thanks for your work identifying the issue :)

Issue #725, opened 2018-06-16
Make wrapper.vnode and wrapper.element to read-only


38elements:
### What problem does this feature solve?
`wrapper.vnode` and `wrapper.element` are not read-only.
`vueWrapper.vnode` and `vueWrapper.element` are read-only.
This is not consistent.

### What does the proposed API look like?
`wrapper.vnode` and `wrapper.element` become read-only.

38elements:
I will send a pull request widthin 7 days.

PR #748, opened 2018-06-23
feat: element, vnode, vm, and options are read-only

38elements:
Since vm, element and options are read-only in the type definition, this changes them to read-only.
Since the element property is read-only, this removes unnecessary  checks.
This is related to #725.

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -23,32 +23,50 @@ import createWrapper from './create-wrapper'
 import { orderWatchers } from './order-watchers'
 
 export default class Wrapper implements BaseWrapper {
-  vnode: VNode | null;
-  vm: Component | null;
+  +vnode: VNode | null;
+  +vm: Component | null;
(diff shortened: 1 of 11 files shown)


packages/test-utils/src/wrapper.js
```diff
@@ -23,32 +23,50 @@ import createWrapper from './create-wrapper'
...
+        get: () => element,
+        set: () => {}
+      })
+      // $FlowIgnore
+      Object.defineProperty(this, 'vm', {
+        get: () => undefined,
+        set: () => {}
+      })
     }
+    const freezedOptions = Object.freeze(options)
```
eddyerburgh: Please changed freezedOptions to `frozenOptions`

eddyerburgh (changes requested):
Thanks, looks fine. One minor change of a variable named (freezed is non-standard English). Otherwise I'm happy to merge

38elements:
Thank you for pointing out.
I changed it.

eddyerburgh (approved):

Issue #747, opened 2018-06-23
Change WrapperArray.wrappers and WrapperArray.length to read-only


38elements:
### What problem does this feature solve?
`WrapperArray.wrappers` and `WrapperArray.length` are read-only in the type definition.
`WrapperArray.wrappers` and `WrapperArray.length` are not read-only.

### What does the proposed API look like?
`WrapperArray.wrappers` and `WrapperArray.length` are read-only.

38elements:
I will send a pull request widthin 7 days.

PR #749, opened 2018-06-23
feat: throw error if the read-only property is tried to change

38elements:
This makes the read-only property to throw error if the read-only property is tried to change.
This is related to #747.

### DIFF
--- a/packages/test-utils/src/vue-wrapper.js
+++ b/packages/test-utils/src/vue-wrapper.js
@@ -1,6 +1,7 @@
 // @flow
 
 import Wrapper from './wrapper'
+import { throwError } from 'shared/util'
 import { setWatchersToSync } from './set-watchers-to-sync'
 import { orderWatchers } from './order-watchers'
 
@@ -11,17 +12,17 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
     // $FlowIgnore : issue with defineProperty
     Object.defineProperty(this, 'vnode', {
       get: () => vm._vnode,
-      set: () => {}
+      set: () => throwError(`VueWrapper.vnode is read-only`)
     })
     // $FlowIgnore
     Object.defineProperty(this, 'element', {
       get: () => vm.$el,
-      set: () => {}
+      set: () => throwError(`VueWrapper.element is read-only`)
     })
     // $FlowIgnore
     Object.defineProperty(this, 'vm', {
       get: () => vm,
-      set: () => {}
+      set: () => throwError(`VueWrapper.vm is read-only`)
     })
     if (options.sync) {
       setWatchersToSync(vm)
--- a/packages/test-utils/src/vue-wrapper.js
+++ b/packages/test-utils/src/vue-wrapper.js
@@ -12,17 +12,17 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
     // $FlowIgnore : issue with defineProperty
     Object.defineProperty(this, 'vnode', {
       get: () => vm._vnode,
-      set: () => throwError(`VueWrapper.vnode is read-only`)
+      set: () => throwError('wrapper.vnode is read-only')
     })
     // $FlowIgnore
     Object.defineProperty(this, 'element', {
       get: () => vm.$el,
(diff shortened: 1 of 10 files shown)


packages/test-utils/src/wrapper-array.js
```diff
@@ -5,12 +5,21 @@ import type VueWrapper from './vue-wrapper'
 import { throwError, warn } from 'shared/util'
 
 export default class WrapperArray implements BaseWrapper {
-  wrappers: Array<Wrapper | VueWrapper>;
-  length: number;
+  +wrappers: Array<Wrapper | VueWrapper>;
+  +length: number;
 
   constructor (wrappers: Array<Wrapper | VueWrapper>) {
-    this.wrappers = wrappers || []
```
38elements, line 12: `wrappers` is always an Array object.

packages/test-utils/src/wrapper-array.js
```diff
@@ -5,12 +5,21 @@ import type VueWrapper from './vue-wrapper'
...
-    this.length = this.wrappers.length
+    const length = wrappers.length
+    // $FlowIgnore
+    Object.defineProperty(this, 'wrappers', {
+      get: () => wrappers,
+      set: () => {}
```
eddyerburgh: We should throw an error to tell the users it's read only.

I realize this isn't part of this PR, but I think we should do the same for vnode, and element.

38elements: Thank you for pointing out.
I will change them.

38elements: `vnode` is undocmumented.
Should `vnode` is added to document?

eddyerburgh: Thanks. No I don't think it needs to be documented

38elements: Thank you for reply.
I think so too.

packages/test-utils/src/vue-wrapper.js
```diff
@@ -11,17 +12,17 @@ export default class VueWrapper extends Wrapper implements BaseWrapper {
     // $FlowIgnore : issue with defineProperty
     Object.defineProperty(this, 'vnode', {
       get: () => vm._vnode,
-      set: () => {}
+      set: () => throwError(`VueWrapper.vnode is read-only`)
```
eddyerburgh: ~~This can be a Wrapper or a VueWrapper, I think just go with:~~

I don't think we need to distinguish between a wrapper and a VueWrapper, I think just go with wrapper:

`wrapper.vnode is read-only`

packages/test-utils/src/wrapper.js
```diff
@@ -667,7 +668,7 @@ export default class Wrapper implements BaseWrapper {
     })
 
     // $FlowIgnore : Problem with possibly null this.vm
-    this.vnode = this.vm._vnode
+    this._vnode = this.vm._vnode
```
38elements: @eddyerburgh 
There is a problem.
This line did not make error at #748.
Is this line necessary?

eddyerburgh: You can flowignore it

38elements: I removed this line.

38elements:
I changed them.

packages/test-utils/src/wrapper.js
38elements: Since vueWrapper.vnode is read-only, I removed this line.

packages/test-utils/src/wrapper-array.js
```diff
@@ -5,12 +5,21 @@ import type VueWrapper from './vue-wrapper'
 import { throwError, warn } from 'shared/util'
 
 export default class WrapperArray implements BaseWrapper {
-  wrappers: Array<Wrapper | VueWrapper>;
-  length: number;
+  +wrappers: Array<Wrapper | VueWrapper>;
```
eddyerburgh, line 8: Cool I didn't know you could specify read only with flow 👍

eddyerburgh (approved):
Thanks

PR #764, opened 2018-06-27
fix: SetValue can not be used with textarea.

mya-ake:
It's actually happening.

```
const wrapper = shallowMount(TextareaComponent)
const textarea = wrapper.find('textarea')
textarea.setValue('foo')  // [vue-test-utils]: wrapper.setValue() cannot be called on this element
```

Because tagName was in lowercase.
https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/wrapper.js#L702

I added a test and fixed that tagName was lowercase.


eddyerburgh (approved):
Thanks 👍

PR #771, opened 2018-06-28
fix: fix tagName in setValue() and setChecked()

38elements:
This fixes `tagName` in `setSelected()` and `setChecked()`.


eddyerburgh (changes requested):
Can you please add tests as well

38elements:
Thank you for pointing out.
I added tests.

eddyerburgh (approved):
Thanks 👍

PR #778, opened 2018-06-29
refactor: remove Wrapper.isVm

38elements:
Since `wrapper.vm` is read-only, `wrapper.isVm` is unnecessary.


eddyerburgh (approved):
Thanks 👍

PR #782, opened 2018-06-29
feat: render slots by default

eddyerburgh:


chriswa:
To be clear, this PR renders the default slot by default, not named slots.

Issue #761, opened 2018-06-26
SetProps on object prop child changes does not trigger computed or watcher


drewtownchi:
### Version
1.0.0-beta.20

### Reproduction link
[https://github.com/drewtownchi/VueRouterTranspile](https://github.com/drewtownchi/VueRouterTranspile)

### Steps to reproduce
Create some data to be passed down as a prop...

```
prop1: {
    sub1: "hello",
    sub2: "data2"
}

```
Create a component that accepts an object prop.
Initialize the test with propsData and verify the child key sub1 is set correctly via the computed property to "hello".
Change the sub1 data in the test to "test2" and update the props using setProps
Check that the sub1 is set correctly via computed property or data (if using a watcher, both don't work)

```
    expect(wrapper.vm.someData).toBe("hello");
    propsDataForNew.sub1 = "test2";
    wrapper.setProps({
      prop1: propsDataForNew
    });
    expect(wrapper.vm.someData).toBe("test2");
```

### What is expected?
the computed property would be updated to the new data passed down via props

### What is actually happening?
The computed property is giving the old data.

  Expected: "test2"
    Received: "hello"

---
This is working correctly in the browser.  The data outputted to the screen reflects the change in sub1 when sub1 is changed in the parent.

Edit: This was working before in previous versions.  So maybe I am just  doing it the wrong way now?  Now should I be just passing in the updated key?

```
    wrapper.setProps({
      prop1: {
        sub1: "test2"
      }
    });
```
This seems to work.  Not sure if I should no longer be passing in an updated object?

drewtownchi:
The above "working" example doesn't actually work because other properties on the object are unset by calling setProps with the partial object.

So, it seems like I should be passing in the full object to setProps still but it isn't forcing an update on computed/watchers.

It does seem the reactivity does happen if I copy the the initial object with Object.assign or the spread operator and then change the copied object.

eddyerburgh:
Yes the problem is you're calling setProps with the same object.

When setProps reassigns the same object, the property getter returns early because the object reference hasn't changed.

I'll look into a way to fix this

eddyerburgh:
The solution has been to throw an error if you're doing this.

Instead, you'll need to create a new object when you call setProps

CindyConway:
The above comment says you will need to ' create a new when you call setProps'. A new what. It looks like a word is missing.

eddyerburgh:
@CindyConway a new object

PR #787, opened 2018-07-01
fix: Support setProps runs computed and watcher when prop is object (#761)

mya-ake:
Related: #761 

### Changes

- Changed to pass clones to props of createInstance in case of Object
- As a result of this change, test of `mount.spec.js` has been updated


packages/create-instance/add-props.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+
+// SetProps on object prop child changes trigger computed or watcher
+// https://github.com/vuejs/vue-test-utils/issues/761
+export function createProps (propsData: Object): Object {
+  return Object.keys(propsData).reduce((props, key) => {
```
eddyerburgh: Can you just use either the spread opertor (not sure if it will trigger reactivity on nested property objects), or lodash.merge, which we already use in setData.

test/specs/mount.spec.js
```diff
@@ -52,9 +52,9 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'mount', () => {
     const wrapper = mount(ComponentWithProps, { propsData: { prop1 }})
     expect(wrapper.vm).to.be.an('object')
     if (wrapper.vm.$props) {
-      expect(wrapper.vm.$props.prop1).to.equal(prop1)
+      expect(wrapper.vm.$props.prop1).to.deep.equal(prop1)
```
eddyerburgh: I was looking at this fix and I'm not sure we should do it because it changes this behavior. Another solution I could think of was to throw an error and tell the user to call setProps with a new object

packages/create-instance/add-props.js
mya-ake: OK, use lodash.merge.
(Using spread operators will no longer be instances of classes.)

test/specs/mount.spec.js
mya-ake: I confirmed the processing of Vue.js, but it was the same value. So, I thought it would be better not to use `deep`.

> Another solution I could think of was to throw an error and tell the user to call setProps with a new object.

I think that is a good way. But I want to use the same object.
I'll look into a way to fix again.

mya-ake:
@eddyerburgh 

### Changes

- To clone data when it is an object.
- It does not affect other tests

eddyerburgh:
This is a good solution, but it still suffers from this problem:
```
it('setProps removes reference', () => {
  const TestComponent = {
    template: `<div val="value"></div>`,
    props: ['value']
  }
  const wrapper = mountingMethod(TestComponent)
  const value = {}
  wrapper.setProps({value})
  expect(wrapper.props().value).to.equal(value)
})
```

I think a better solution would be to throw an error if the object in setProps has the same reference, and make the user create a new object. That way we situations that might be confusing, like the above.

mya-ake:
Hmmm, I can't think of a way to pass the test.
I will withdraw this Pull request and the next Pull request will throw an error when the same reference.

PR #791, opened 2018-07-03
fix: setProps() throws an error if the property is the same reference

mya-ake:
Related: #761, #787  

Changes

- setProps() throws an error if the property is the same reference
- Added test that props are the same reference

Please tell me if there is a better error message.

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -657,6 +657,17 @@ export default class Wrapper implements BaseWrapper {
           `is not defined on the component`
         )
       }
+      if (
+        typeof data[key] === 'object' &&
+        data[key] !== null &&
+        // $FlowIgnore : Problem with possibly null this.vm
+        data[key] === this.vm[key]
+      ) {
+        throwError(
+          `wrapper.setProps() called with ${key} property ` +
+          `needs to create a new Object`
+        )
+      }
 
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
Tests look good to me. Shame there isn't a nice way to merge newly set props with existing ones (maybe in a future update or something!)

Will be good to have this error message at least, to help guide people having the same problem you did

packages/test-utils/src/wrapper.js
```diff
@@ -657,6 +657,17 @@ export default class Wrapper implements BaseWrapper {
...
         )
       }
+      if (
+        typeof data[key] === 'object' &&
+        data[key] !== null &&
+        // $FlowIgnore : Problem with possibly null this.vm
+        data[key] === this.vm[key]
+      ) {
+        throwError(
+          `wrapper.setProps() called with ${key} property ` +
```
eddyerburgh: wrapper.setProps called with the same object

eddyerburgh: Please reword:
`wrapper.setProps() called with the same object of the existing ${key} property. You must call `wrapper.setProps()` with a new object to trigger reactivity`

mya-ake:
@eddyerburgh 
Thanks! 
Updated the message.

eddyerburgh (approved):
Thanks 👍

PR #800, opened 2018-07-03
chore(types): remove unnecessary type definition

38elements:
`sizzle` is not used in this repository.


eddyerburgh (approved):
Thanks @38elements , but be aware that the entire repo is going to be rewritten with TypeScript in a few months, so flow will be deleted then.

PR #804, opened 2018-07-04
docs: fix broken links in guides/choosing-a-test-runner (fix #805)

### LINKED ISSUE #805: docs: Broken links
### Version
1.0.0-beta.20

### Reproduction link
[https://vue-test-utils.vuejs.org/guides/#testing-single-file-components](https://vue-test-utils.vuejs.org/guides/#testing-single-file-components)

### Steps to reproduce
Open https://vue-test-utils.vuejs.org/guides/#testing-single-file-components

Try to open  in new tab this links:
- Testing SFCs with Jest
- Testing SFCs with Mocha   webpack

### What is expected?
Open docs section in new tab

404 error

moraveyo:


38elements:
It seems that the links work normally now.
Would you tell me why this change is necessary?
https://vue-test-utils.vuejs.org/guides/choosing-a-test-runner.html#testing-single-file-components

moraveyo:
Doesn't open in new tab - got 404. I filed an issue #805

moraveyo:
These links open not really as I expect In the same window too. In fact, these are the sections of the same document: look at the navigation, and after all you can simply scroll down to them.

But when you click at the link - it opens as separate document.
A little confusing.

moraveyo:
Some research.

When you just click the link, it opens with url:https://vue-test-utils.vuejs.org/guides/testing-SFCs-with-jest.html (see the upper case!) - the Vue works it out fine.

When you open it in new tab/window - the page first blinks like something is there, and then it's 404. And the url is all lowercase.

So maybe it's some deeper problem in Vue itself, I'm not so deep in it.

moraveyo:
Should have written it in the issue... I'll copy there.

eddyerburgh:
Thanks for this @moraveyo . Does the URL change solve your issue?

moraveyo:
Yes, it does.

eddyerburgh (approved):

Issue #657, opened 2018-05-25
JSX $scopedSlots support


panganibanpj:
### What problem does this feature solve?
Being able to use $scopedSlots with render function (specifically for JSX usage)
```javascript
shallowMount({
    render() {
        return this.$scopedSlots.foo({ bar: 'baz' });
    },
}, {
    scopedSlots: {
        foo: '<div>{{ bar }}</div>',
    },
})
```
It will complain that `$scopedSlots.foo is not a function`, but if you make it a function, it will say `$scopedSlots.foo.trim() is not a function`

### What does the proposed API look like?
I prefer if `scopedSlots` can be passed function values like:
```javascript
scopedSlots: {
  foo: (createElement, props) => { return createElement('div', props); },
},
```

Otherwise, at least make it so the above syntax would still work (with string `scopedSlot.foo` values but the `this.$scopedSlots.foo()` in the render function still work)

38elements:
What is benefit for supporting JSX $scopedSlots?

davestaab:
This feature would be useful for making renderless components such as outlined by Adam here: https://adamwathan.me/renderless-components-in-vuejs/

It would be helpful to test the component api.
Currently I get the above error in jest in my unit test.

kayandra:
Hi, please can someone help me with how I can test scoped slots?

Currently, my render function looks like this

```
  render() {
    return this.$scopedSlots.default({ ...this.computedStateAndHelpers })
  },
```

In my test (with Jest), I'm trying to mount like this:

```
    const wrapper = shallowMount(Component, {
      scopedSlots: {
        default: '<p></p>'
      }
    })
    expect(wrapper).toBeAViewComponent()
```

But no luck.

I've also tried making default property a method and no luck there either. I believe I followed the docs properly. How do I do this? Thanks!

panganibanpj:
I've worked around it by wrapping the component like
```javascript
shallowMount({
  render() {
    return (
      <Component
        scopedSlots={{
          default: <p></p>
        }}
      />
    );
  }
})
```

kayandra:
hi @panganibanpj I still get 

```
TypeError: this.$scopedSlots.default is not a function
```

Also, won't I lose direct access to the components vm?

davestaab:
@kayandrae07 You should be able to call `find` to get the nested component (may have to mount instead of shallowMount)


Something like this: 
```
const wrapper = mount({
  render() {
    return (
      <Component
        scopedSlots={{
[...]

PR #808, opened 2018-07-05
fix: improve scopedSlots option

### LINKED ISSUE #706: Tests for default scopedSlots not mounting
### Version
1.0.0-beta.18

### Reproduction link
[https://codesandbox.io/s/0mrvjmvk6l](https://codesandbox.io/s/0mrvjmvk6l)

### Steps to reproduce
Hi there, so I’m building a library for work and I’m using the render function and only accepting default scoped slots. I have the render function currently defined as

```
  render() {
      return this.$scopedSlots.default({ ...this.computedStateAndHelpers })
  },
```

When I try to write tests for it (using jest) I always get an error when trying to mount the component

```
const wrapper = shallowMount(Component, {
    scopedSlots: {
        default: () => '<p></p>'
    }
})
expect(wrapper.isVueInstance()).toBe(true)
expect(wrapper.name()).toBe('component-name')
```

I also tried

```
const wrapper = shallowMount(Component, {
    scopedSlots: {
        default: '<p></p>'
    }
})
expect(wrapper.isVueInstance()).toBe(true)
expect(wrapper.name()).toBe('component-name')
```

The error is

```
scopedSlots[key].trim is not a function
```

I don't know what I'm doing wrong, can someone please help me figure this out?

### What is expected?
Expecting the test to pass

The error is

```
scopedSlots[key].trim is not a function
```

38elements:
This resolves #706.


eddyerburgh (changes requested):
Looks great! One small change. Can you rename the function to `createScopedSlots`, to be in line with createComponentStubs, createSlotVNodes, etc?

38elements:
Thank you for reviewing.
I renamed `getScopedSlots` to `createScopedSlots`.

eddyerburgh:
Thanks for making the changes.

Can you also add a test case for the example in the issue:

```js
scopedSlots: {
  foo: (createElement, props) => { return createElement('div', props); },
},
```

38elements:
The case does not work.
I think the feature is not necessary.
I think it is not necessary to use `createElement()` to make `scopedSlots`.
`slots` mounting option is not support it.
Would you please tell me why this feature is necessary?

38elements:
`scopedSlotFn()` receives only `props`.

```js
nodes = scopedSlotFn(props) || fallback
```

https://github.com/vuejs/vue/blob/5a9da95b8a865416f082952a48416ffc091e4078/src/core/instance/render-helpers/render-slot.js#L27

38elements:
`scopedSlotFn()` receives only `props`.

```js
nodes = scopedSlotFn(props) || fallback
```

https://github.com/vuejs/vue/blob/5a9da95b8a865416f082952a48416ffc091e4078/src/core/instance/render-helpers/render-slot.js#L27


If you think the feature is necessary, I will add an API as below.

```js
scopedSlots: {
  default: (createElement) => (props) => { return createElement('div', props); }
}
```

eddyerburgh:
Ok, I agree that the API does not look good. I'll merge the branch as it is 👍

eddyerburgh (approved):

38elements:
Thanks.

38elements:
#657 is not closed.
I want to confirm.
Do you think this API is necessary?

```js
scopedSlots: {
  default: (createElement) => (props) => { return createElement('div', props); }
}
```

eddyerburgh:
Not currently, but it's a feature request so we should keep it open until we've decided definitively. 

I also haven't had enough time to fully assess the scopedSlots and slots API

38elements:
Thank you for reply.
I think that it is better to remove `intend to impliment` label for #657.

kayandra:
@38elements thank you for this

PR #812, opened 2018-07-06
data should be a function

### LINKED ISSUE #797: data should be function
### Version
1.0.0-beta.20

### Reproduction link
[https://github.com/vuejs/vue-test-utils/blob/1ee03ac2d57eccd913027ecd51ac5777eeac5c84/test/specs/mounting-options/localVue.spec.js#L20](https://github.com/vuejs/vue-test-utils/blob/1ee03ac2d57eccd913027ecd51ac5777eeac5c84/test/specs/mounting-options/localVue.spec.js#L20)

### Steps to reproduce
Open this:
https://github.com/vuejs/vue-test-utils/blob/1ee03ac2d57eccd913027ecd51ac5777eeac5c84/test/specs/mounting-options/localVue.spec.js#L20
```
 data: { test: '' } 
```
should be written as
```
data: () => ({ test: '' })
```

### What is expected?
data: () => ({ test: '' })

 data: { test: '' } 

---
https://vuejs.org/v2/guide/components.html#data-Must-Be-a-Function

LinusBorg:
closes #797

### DIFF
--- a/test/specs/mounting-options/localVue.spec.js
+++ b/test/specs/mounting-options/localVue.spec.js
@@ -17,7 +17,7 @@ describeWithMountingMethods('options.localVue', mountingMethod => {
         template: `
         <div>{{test}}</div>
       `,
-        data: { test: '' }
+        data: ()=> ({ test: '' })
       }
       const localVue = Vue.extend()
       localVue.version = '2.3'
--- a/test/specs/mounting-options/localVue.spec.js
+++ b/test/specs/mounting-options/localVue.spec.js
@@ -17,7 +17,7 @@ describeWithMountingMethods('options.localVue', mountingMethod => {
         template: `
         <div>{{test}}</div>
       `,
-        data: ()=> ({ test: '' })
+        data: () => ({ test: '' })
       }
       const localVue = Vue.extend()
       localVue.version = '2.3'
--- a/test/specs/mounting-options/localVue.spec.js
+++ b/test/specs/mounting-options/localVue.spec.js
@@ -22,9 +22,9 @@ describeWithMountingMethods('options.localVue', mountingMethod => {
       const localVue = Vue.extend()
       localVue.version = '2.3'
       const wrapper = mountingMethod(TestComponent, {
-        localVue: localVue,
-        mocks: { test: 'some value' }
+        localVue: localVue
       })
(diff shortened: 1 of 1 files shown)


LinusBorg:
WTH do those tests fail now? confused ...

test/specs/mounting-options/localVue.spec.js
```diff
@@ -17,14 +17,14 @@ describeWithMountingMethods('options.localVue', mountingMethod => {
...
+        data: () => ({ test: '' })
       }
       const localVue = Vue.extend()
       localVue.version = '2.3'
       const wrapper = mountingMethod(TestComponent, {
-        localVue: localVue,
-        mocks: { test: 'some value' }
+        localVue: localVue
       })
+      wrapper.vm.test = 'some value'
```
LinusBorg, line 27: @eddyerburgh I had to remove the `mocks: { test: 'some value'}` and replace it with this, otherwise the test woudl fail.

I *suspect* that this is because you can't actually mock properties for proxied data properties, but that didn't fail before because the `data: { test: '' }` itself failed?

Did my change break the intended goal of the test?

LinusBorg:
Okay now I'm double-confused, the tests passed locally after https://github.com/vuejs/vue-test-utils/pull/812/commits/d8101dc05eddef4271d7bd2e8fd96f1fc5976401

on CircleCI it breaks with a new error ...

test/specs/mounting-options/localVue.spec.js
eddyerburgh: This test is the first one written for localVue, and itcould be rewritten. The intent is to check that VTU mounts using the localVue constructor, so a better test would create a localVue constructor with createLocalVue, add a property to the localVue prototype, and mount a component that relies on that property existing.

Would you be able to rewrite it?

LinusBorg: Will try tomorrow

LinusBorg: might take a few days until I really have the time :/

eddyerburgh:
Thanks for this Thorsten , I've fixed this in #835 :)

LinusBorg:
Yeah sorry, didn't have/make the time for this :/

PR #813, opened 2018-07-08
fix: improve slots option

### LINKED ISSUE #793: $parent of a child component defined on a slot, reference to itself
### Version
1.0.0-beta.20

### Reproduction link
[https://codesandbox.io/s/wqo716rm55](https://codesandbox.io/s/wqo716rm55)

### Steps to reproduce
Run the test written on sandbox

### What is expected?
It should work as works on the browser so `this.$parent` should return the parent of the component and not itself.

I have a scenario similar to what I reproduce in the sandbox. I have a few components that are inserted on the default slot of a "wrapper" component, each of these components are registered on the parent when are mounted using a method on the parent `this.$parent.register`.
All work on the browser but when I tried to test the `this.$parent` of the child component refers to itself and the test fails because it doesn't find the register method.

---
For example If I change `this.$parent` by `this.$parent.$parent` all works.

38elements:
This resolves #793.

### DIFF
--- a/packages/create-instance/create-slot-vnodes.js
+++ b/packages/create-instance/create-slot-vnodes.js
@@ -2,26 +2,33 @@
 
 import { compileToFunctions } from 'vue-template-compiler'
 
+function createVNodes (
+  vm: Component,
+  slotValue: string
+): Array<VNode> {
+  const el = compileToFunctions(`<div>${slotValue}</div>`)
+  const _staticRenderFns = vm._renderProxy.$options.staticRenderFns
+  // version < 2.5
+  if (!vm._renderProxy._staticTrees) {
+    vm._renderProxy._staticTrees = []
+  }
+  vm._renderProxy.$options.staticRenderFns = el.staticRenderFns
+  const vnode = el.render.call(vm._renderProxy, vm.$createElement)
+  vm._renderProxy.$options.staticRenderFns = _staticRenderFns
+  return vnode.children
+}
+
 function createVNodesForSlot (
-  h: Function,
+  vm: Component,
   slotValue: SlotValue,
   name: string,
-  vm: Component
 ): VNode | string {
   let vnode
   if (typeof slotValue === 'string') {
-    const el = compileToFunctions(`<div>${slotValue}</div>`)
-    const _staticRenderFns = vm._renderProxy.$options.staticRenderFns
-    // version < 2.5
-    if (!vm._renderProxy._staticTrees) {
-      vm._renderProxy._staticTrees = []
-    }
-    vm._renderProxy.$options.staticRenderFns = el.staticRenderFns
-    vnode = el.render.call(vm._renderProxy, h)
-    vm._renderProxy.$options.staticRenderFns = _staticRenderFns
-    vnode = vnode.children[0]
+    const vnodes = createVNodes(vm, slotValue)
+    vnode = vnodes[0]
   } else {
-    vnode = h(slotValue)
+    vnode = vm.$createElement(slotValue)
   }
   if (vnode.data) {
     vnode.data.slot = name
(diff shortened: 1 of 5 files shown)


packages/create-instance/create-slot-vnodes.js
```diff
@@ -18,7 +19,17 @@ function createVNodesForSlot (
   const el =
     typeof slotValue === 'string' ? compileToFunctions(slotValue) : slotValue
 
-  const vnode = h(el)
+  let vnode = h(el)
+  if (typeof slotValue === 'string') {
+    const vue = new Vue()
```
eddyerburgh: I think this would be a problem if you mount a component that uses localVue and the slot component accesses a property that is only added to the localVue prototype, can you write a test to check?

38elements:
This case does not work.
I will make another issue.

```js
slots: {
  default: '<div><component-with-parent-name /></div>'
}
```

packages/create-instance/create-slot-vnodes.js
38elements: I will add test.

38elements: Yes, it is.
Changing `localVue.prototype`, this does not work.
Using `vm._renderProxy` is correct if possible.
I will add test for changing `localVue.prototype`.

38elements:
I will add test for changing `localVue.prototype`.
I think this change is temporary solution.
I will seek another solution.

38elements:
I added a test for changing `localVue.prototype`.
I think this change is temporary solution.
I will seek another solution.

packages/create-instance/create-slot-vnodes.js
```diff
@@ -22,6 +23,12 @@ function createVNodesForSlot (
   let vnode = h(el)
   if (typeof slotValue === 'string') {
     const vue = new Vue()
+    const _vue = new _Vue()
```
eddyerburgh: why not just use the _vue.renderProxy rather than mixing both?

38elements: Since it raises an error.

38elements:
I will send  the pull request for another solution within 3 days.

38elements:
@eddyerburgh 
I sent the pull request for another solution.
It is #821.

38elements:
I think #821 is better than this.

packages/create-instance/create-slot-vnodes.js
```diff
@@ -18,22 +20,41 @@ function createVNodesForSlot (
...
     typeof slotValue === 'string' ? compileToFunctions(slotValue) : slotValue
 
-  const vnode = h(el)
+  let vnode = h(el)
+  if (typeof slotValue === 'string') {
+    const vue = new Vue()
+    const _vue = new _Vue()
+    for (const key in _vue._renderProxy) {
+      if (!(vue._renderProxy[key])) {
+        vue._renderProxy[key] = _vue._renderProxy[key]
```
eddyerburgh: Why not always use _vue.renderProxy?

packages/create-instance/create-slot-vnodes.js
```diff
@@ -18,22 +20,41 @@ function createVNodesForSlot (
...
+      }
+    }
+    try {
+      // $FlowIgnore
+      vnode = el.render.call(vue._renderProxy, h)
```
eddyerburgh: Do components without a render function work here?

For example:

```js
mount(TestComponent,{
  slots: { default: {
    template: '<div />'
  }}
})
```

Can you add a test for this kind of slot

eddyerburgh:
Could you also include a fix for #828? Your previous PR would have solved this (adding a parent div to template strings, and using the child value)

38elements:
I will try to add it.

packages/create-instance/create-slot-vnodes.js
38elements: ~~When `vue.renderProxy` is used, it raises an error.~~ 
When `_vue.renderProxy` is used, it raises an error.

```
TypeError: Cannot read property 'Ctor' of undefined
```

eddyerburgh: Do you mean when `_vue.renderProxy` is used?

38elements: It does not raise the error.

38elements: I am sorry.
Yes, I do.

38elements:
@eddyerburgh 
I changed this.
IMHO,
Since #828 does not uses slots, I think this pull request is not related it.
I do not know #734 well.
It seems that this might be related #734.
I think this might resolve #734.

eddyerburgh:
Ok sounds good. Can you add the tests from #734 to verify that this PR solves the issue?

38elements:
@eddyerburgh 
I added a test for #734.

eddyerburgh (approved):

eddyerburgh:
Nice work, thanks :)

38elements:
@eddyerburgh 
I am sorry.
#734 is not resolved.

38elements:
@eddyerburgh 
Thank you for reviewing.
I am sorry.
The test for #734 which is added at this pull request is incorrect.
#734 is not resolved.
I sent a pull request #847.

Issue #738, opened 2018-06-19
setProps doesn't update props in some circumstances


TheCycoONE:
### Version
1.0.0-beta.19

### Reproduction link
[https://codesandbox.io/s/1k6y65wp4](https://codesandbox.io/s/1k6y65wp4)

### Steps to reproduce
Create a component, like the one in the above reproduction, with multiple properties that are watched.
It is important that the watch function reads one of the properties (i.e. if you remove that switch statement from my test case the problem doesn't happen.)
Create a test for that component, call setProps on the property that is read in the watch.

### What is expected?
The property should be updated, the component should react to the updated property

### What is actually happening?
The property is never updated. props() shows the old value.

Note: this is a regression, it worked correctly in 1.0.0-beta.16 (I did not try other versions)

eddyerburgh:
Sorry, this is another problem with how we've implemented synchronous renders.

There's a PR open to fix this by adding an async option to Vue—https://github.com/vuejs/vue/pull/8240.

For the moment, the workaround is to set the `sync` mounting option to false and use Vue.nextTick/ setTimeout to await DOM updates (like you're doing):

```js
test('use Vue.nextTick', (done) => {
  const wrapper = mount(TestComponent, { sync: false })
  wrapper.trigger('click')
  Vue.nextTick(() => {
    expect(wrapper.text()).toBe('updated')
    done()
  })
})
```

TheCycoONE:
Setting sync false and waiting for Vue.nextTick doesn't appear to work around the `countingAlgo` prop not updating case that started happening after updating from beta 16 to beta 19.

It does workaround the DOM not updating when `setProp({value: 42})` is called which was needed in 16 as well.

HcySunYang:
@TheCycoONE @eddyerburgh 
`setProps` does not work in vue@2.2.6, behaves like this issue, but works fine in the latest version of vue.

TheCycoONE:
@HcySunYang I am on vue 2.5.16

agriffis:
@TheCycoONE I'm hitting this too. What are you doing as a workaround?

TheCycoONE:
Unfortunately I haven't found a workaround yet. For now I'm *.skip*ping those tests.

agriffis:
I guess `npm i -D @vue/test-utils@1.0.0-beta.17` works. 18, 19 and 20 are broken.

agriffis:
[...]

Issue #676, opened 2018-06-01
Fix synchronous behavior


eddyerburgh:
Before this library is ready for 1.0.0 we need to fix the synchronous behavior.

## Context

By default Vue performs DOM updates asynchronously.  When we created this library, we decided to make it perform updates synchronously, to improve the readability of tests. 

The original way that we implemented synchronous DOM updates was by running the instance render watcher when wrapper method, like `trigger` or `setProps` changed instance properties.

There were two issues with that approach. First, if you made a change that occurred outside of Vue Test Utils, you would need to call an `update` function that ran watchers. 

Second, there were bugs that were difficult to solve, such as watchers running when they should not. 

In beta.12 we changed the approach by setting all watchers as synchronous. This approach  [has caused a number of bugs](https://github.com/vuejs/vue-test-utils/issues?q=is%3Aissue+is%3Aopen+label%3Async)  because watchers cannot complete the full lifecycle. Some of these bugs are very difficult to fix.

## Solution

I've submitted [a PR](https://github.com/vuejs/vue/pull/8240) to Vue core that will allow us to set tests to run synchronously.  This will fix all the current bugs with synchronous.

## Problem

The problem is that this change will only be available in Vue > 2.6. This library supports all Vue 2.x versions, so we need to make a decision:

**Do we continue with the current buggy synchronous behavior, or remove synchronous behavior for Vue < 2.6 and encourage users to use `Vue.nextTick`/ `setTimeout` to run tests?**

What do people think?

lmiller1990:
I think it would be valuable to see a few examples of which tests will need to be modified if the sync behavior is moved. I don't want to write `Vue.$nextTick` in _all_ my tests, but if is just tests around components with `$watch`, for example, I think it's okay.  

Buggy behavior is never fun to deal with, and since we are still in beta, I think we should work towards where we _want_ the library to be, even if it means a change such as removing the sync behavior. As long as we provide a warning (and possibly a link to what changed, why, and how to fix tests that are failing due to the change) I think it's okay.

38elements:
I think that it is better to improve `sync` mounting option and treat its unresolved issues as limitations which is described in document.

Or Providing mock Vue which is added `async` option by each version and setting alias vue at webpack is resolve this issue.

webpack.test.config.js

```
// ...
  resolve: {
     alias: {
         vue: '@vue/vue23-for-test-utils'
      }
   }
// ...
```

AlbertBrand:
[...]

PR #820, opened 2018-07-12
test: demonstrate setProps bug through computed and created (#738)

agriffis:


eddyerburgh:
Thanks for the bug repro!

This is a bug with the synchronous implementation that changed in beta.16 (https://github.com/vuejs/vue-test-utils/issues/676).

This is a very serious issue that I'm focussed on fixing.

The solution at the moment (and possibly the long term solution) is to run the tests without sync mode:
```js
it('sets component prop referenced through computed and created', () => {
  const wrapper = mountingMethod(ComponentWithComputedCreated, {propsData: {prop1: 1}, sync: false})
  wrapper.setProps({prop1: 2})
  setTimeout(() => {
    expect(wrapper.vm.prop1).to.equal(2)
  })
}) 
```

agriffis:
@eddyerburgh Just for kicks, I rebased this to latest and it's great to see it works!

leifriksheim:
In my case using `setTimeout` will make all test pass, even if they shouldn't

eddyerburgh:
This should be fixed in beta.29

PR #821, opened 2018-07-12
fix: improve slots mounting option

### LINKED ISSUE #793: $parent of a child component defined on a slot, reference to itself

38elements:
This resolves #793.
This is related to #813.
This does not reslove #827.


38elements:
I will add some tests and improvements within 3 days.

packages/create-instance/create-render-slot.js
```diff
@@ -0,0 +1,62 @@
...
+import { compileToFunctions } from 'vue-template-compiler'
+
+const _renderSlot = Vue.prototype._t
+
+function createVNodes (
+  vm: Component,
+  slotValue: Component | string
+): ?Array<VNode> {
+  if (typeof slotValue === 'string') {
+    const compiledResult = compileToFunctions(`<div>${slotValue}{{ }}</div>`)
```
eddyerburgh: Why the empty interpolation `{{ }}`?

38elements: I do not understand.
It is not necessary.
I thought the compileToFunctions raises an error if it dose not exist at #274.

38elements:
I think it is necessary to add warning if  `compileToFunctions` is not existing.
I will add them.

packages/create-instance/create-render-slot.js
```diff
@@ -0,0 +1,64 @@
...
+function createVNodes (
+  vm: Component,
+  slotValue: Component | string
+): ?Array<VNode> {
+  if (typeof slotValue === 'string') {
+    // Since compileToFunctions is checked in createSlotVNodes(),
+    // it is not necessary to check compileToFunctions.
+    const compiledResult = compileToFunctions(`<div>${slotValue}</div>`)
+    const _staticRenderFns = vm._renderProxy.$options.staticRenderFns
+    vm._renderProxy.$options.staticRenderFns = compiledResult.staticRenderFns
```
eddyerburgh, line 17: Why do you save the renderFunctions and then reassign?

test/specs/mounting-options/slots.spec.js
```diff
@@ -568,4 +574,72 @@ describeWithMountingMethods('options.slots', mountingMethod => {
...
     }
   })
+
+  itDoNotRunIf(
+    mountingMethod.name === 'renderToString',
+    'mounts component with default slot if passed string in slot object',
+    () => {
+      const wrapper1 = mount(ComponentWithSlots, { slots: { default: 'foo<span>123</span>{{ foo }}' }})
+      expect(wrapper1.find('main').html()).to.equal('<main>foo<span>123</span>bar</main>')
+      const wrapper2 = mount(ComponentWithSlots, { slots: { default: '<p>1</p>{{ foo }}2' }})
```
eddyerburgh, line 584: I don't think this is how slots should behave. foo should not be resolved using the child vm that it's rendered in. If it uses any instance to resolve foo on, it should be the root instance. But I don't think we should support any variables in the slots option as it adds too much complexity

packages/create-instance/create-render-slot.js
```diff
@@ -0,0 +1,64 @@
...
+    const vnodes = compiledResult.render.call(
+      vm._renderProxy, vm.$createElement
+    ).children
+    vm._renderProxy.$options.staticRenderFns = _staticRenderFns
+    return vnodes
+  }
+  return [vm.$createElement(slotValue)]
+}
+
+export default function createRenderSlot (
```
eddyerburgh, line 27: What is the purpose of replacing the renderSlot alias? Does this solve the issue of parent being undefined?

test/specs/mounting-options/slots.spec.js
38elements: I am sorry.
Your pointing out is correct.
I misunderstood about it.

PR #825, opened 2018-07-13
fix: allow extended components as stubs

kellym:
The current test for accepting valid component stubs references an `ExtendedComponent`, but technically doesn't allow components that extend another component via the single-file means of `{ extends: Component }`.

This makes use of the existing shared `isVueComponent` function and combines it with the current stub-version of `isVueComponent`, which looks like it's present because `{ template }` support isn't included in the shared `isVueComponent`.


eddyerburgh (approved):
Thans 👍

Issue #819, opened 2018-07-12
Unit testing nested components - Child's `emitted` is undefined


bugsduggan:
### Version
1.0.0-beta.20

### Reproduction link
[https://github.com/bugsduggan/vue-test-utils-bug](https://github.com/bugsduggan/vue-test-utils-bug)

### Steps to reproduce
Clone the repo and run `npm run test:unit`. There will be one failing test indicating that the `ChildComponent`'s emitted is undefined.

### What is expected?
Having used `mount` instead of `shallowMount`, I was hoping to be able to see events that have been emitted by the `ChildComponent`.

### What is actually happening?
`ChildComponent`'s `emitted` function returns undefined.

---
I appreciate that this could be seen as testing the implementation of vue but I have 2 components that are very tightly coupled in my application and it is important for us to be able to check what events are being triggered by the child component while unit testing the parent.

If someone can help me narrow down the causes of this bug (and if it's not impossible or undesirable to fix), I would be happy to devote some hours into a pull request.

lmiller1990:
The event does seem to be triggered - however it is not added to the array returned by `emitted`. I think this is a bug. 

A long time ago, before `vue-test-utils`, I had  a similar situation. I ended up doing something like

```ts
let eventName: string = ''
let payload: Array = []

function createEventRecorder() {
  return (event: string, ...args: any[]): Vue =>
   {
     eventName = event
     payload = args
     return new Vue();
   };
}
```

```js
expect(eventName)...
expect(payload)...
```

And mocking out the `$emit` entirely. But now we have `vue-test-utils`, this should be nice and clean.

@eddyerburgh does this seem like a bug? If so, I can help isolate it with you. I made a PR against the `emitted` function a while ago, so I'm a bit familiar with how it works.

eddyerburgh:
Yes this is a bug. 

I believe the bug is because the child component is extended from Vue before the event logging mixin is added. @lmiller1990 it would be great if you could pick this up.

lmiller1990:
Ok thanks. Excited to work on this over the weekend. I need this in my app, too.

lmiller1990:
Ok, I investigated this. It works fine for the non typescript case. I think the bug must be related to TypeScript, somehow. Should we add some test cases using TypeScript to be able to test this? 

`emitted` has the correct properties in the case of normal `.vue` files, and also `.vue` files using `vue-property-decorator`. Introducing TypeScript seems to cause this bug.

[...]

PR #832, opened 2018-07-18
Fix bug where `Vue.mixin` overrides global components registered by plugins

lmiller1990:
I am working towards isolating and fixing a (potential) bug which might be causing #658 and #819 . I think something odd is happening around `Vue.mixin` when used with `localVue`.

I added a reproduction test and need some opinions on if this is the right way to go.
___

The last test in the diff (in test/specs/shallow-mount.spec.js)  with `it.only` has a comment like this:

```js
// localVue.use(myPlugin)
```

Uncommenting causes a failure which I don't think should be happening.

In `shallow-mount.js`, if you do `console.log(createComponentStubsForGlobals(vue))` __after__ `Vue.mixin` is called, the `options.components` is empty. That's why the bug in https://github.com/vuejs/vue-test-utils/issues/658 is occurring, I think, and others than reply on plugins or `Vue.mixin`.

Spend a long time on this but not sure why, but I feel like this is the source of several bug reports. @eddyerburgh do you have any thoughts/ideas (or am I going in the complete wrong direction with this? Does what I tried to explain above even make sense?)

Thanks for any ideas of input, I understand you are busy with many other issues and PRs.

### DIFF
--- a/packages/test-utils/src/shallow-mount.js
+++ b/packages/test-utils/src/shallow-mount.js
@@ -23,6 +23,7 @@ export default function shallowMount (
     delete component.components[hyphenate(component.name)]
   }
 
+  console.log(createComponentStubsForGlobals(vue))
   return mount(component, {
     ...options,
     components: {
--- a/test/resources/components/component-using-vuetify.vue
+++ b/test/resources/components/component-using-vuetify.vue
@@ -1,0 +1,17 @@
+<template>
+  <v-app>
+    <v-container>
+      <v-layout>
+        <v-flex>
+          <h1>hello</h1>
+        </v-flex>
+      </v-layout>
+    </v-container>
+  </v-app>
+</template>
+
+<script>
+export default {
+  name: 'ComponentUsingVuetify'
+}
+</script>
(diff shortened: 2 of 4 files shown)


eddyerburgh:
Thanks for the work @lmiller1990 

Yes I can see this would be the source of many bugs. I'm not sure how we could fix it. 

~~One solution would be to ask the plugin authors to use the _Vue constructor that's passed in to the mixin, rather than the global base constructor~~

They actually do this already, I need to investigate further to understand the problem.

eddyerburgh:
In fact, I'm sure that we need to ask any plugin authors to use the Vue constructor that they receive in the install function, like how vee-validate uses the constructor: https://github.com/baianat/vee-validate/blob/master/src/install.js#L9.

eddyerburgh:
Thanks for the PR. The issue was that Vue.mixin merges the component options object into a new object with the original component options object as a prototype of the new object. It's created here: https://github.com/vuejs/vue/blob/dev/src/core/util/options.js#L165.

The solution was to use `for ... in` rather than `Object.keys` to iterate through the global components to stub. Object.keys only iterates through the objects own properties, `for ... in` iterates through the properties in the object including properties in the prototype chain.

lmiller1990:
Ah, I see! Good to know. Thanks for letting me know, I didn't know the difference between those two. Keep up the great work on `vue-test-utils`.

PR #834, opened 2018-07-19
Render props on auto stubs

### LINKED ISSUE #708: Add child-Prop in stub of ChildComponent (when using snapshot)
I would like to see in my snapshot that I call my childcomponent with correct prop.

```
 <childcomponent-stub child-Prop="'some propsData that needs to be tested'"/>
````

Here is a failing test with my expected snapshot:

https://codesandbox.io/s/ql77047m9q

nothing to add here

Akryum:
closes #708


eddyerburgh (changes requested):
Great :) 

Can you add a test please?

eddyerburgh:
I'll add a test :)

Do you have an example where this.$attrs should be rendered?

eddyerburgh (approved):

shell362:
You are my heroes, guys! 👍

PR #856, opened 2018-07-27
fix: keep the overrides prototype information of component

kuitos:
add the component prototype override scenario and remove the version condition.
@eddyerburgh Actually I don't know why we need add a version condition here: https://github.com/vuejs/vue-test-utils/pull/709/commits/cacfda8acb2dfed54295b82fbd93e84205cec9d3 , any special reason?

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -110,7 +110,9 @@ export default function createInstance (
     component.options._base = _Vue
   }
 
-  const Constructor = vueVersion < 2.3 && typeof component === 'function'
+  // when component constructed by Vue.extend,
+  // use its own extend method to keep component information
+  const Constructor = typeof component === 'function'
     ? component.extend(instanceOptions)
     : _Vue.extend(component).extend(instanceOptions)
 
--- a/test/specs/mount.spec.js
+++ b/test/specs/mount.spec.js
@@ -153,6 +153,28 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'mount', () => {
     expect(stub).not.called
   })
 
+  it('overrides component prototype', () => {
+    const mountSpy = sinon.spy()
+    const destroySpy = sinon.spy()
+    const Component = Vue.extend({})
+    const { $mount: originalMount, $destroy: originalDestroy } = Component.prototype
+    Component.prototype.$mount = function (...args) {
+      mountSpy()
+      originalMount.apply(this, args)
+      return this
+    }
+    Component.prototype.$destroy = function () {
+      originalDestroy.apply(this)
+      destroySpy()
+    }
+
+    const wrapper = mount(Component)
+    expect(mountSpy).called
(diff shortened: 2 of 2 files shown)


eddyerburgh:
Yes, it's due to a bug where extended options aren't resolved correctly (https://github.com/vuejs/vue/issues/4976)

It was fixed in Vue 2.3

eddyerburgh (approved):

eddyerburgh:
This change is actually problematic.

We have an issue with components that already been extended (with Vue.extend). They never extend from the localVue class when they are instantiated, so they don't inherit assets like components set in `stubs`, and properties set in `mocks`.

The original code removed by this PR extended all components, including class components, EXCEPT in Vue < 2.3, where components extended with Vue.extend did not merge props correctly.

I would like to revert this PR, because it stops us from being able to add assets to the component without mutating it. But I can see that the previous code was causing you issues in your tests.

Can you give me some details about why your tests don't work when the component is extended?

kuitos:
@eddyerburgh Sure.
As we had overridden the prototype of extended component in my test, for some hijack requirements, when vue >= 2.3 in original code the process would go to `_Vue.extend(component).extend(instanceOptions)`.

From the implementation of `Vue.extend`:
```js
Vue.extend = function (extendOptions: Object): Function {
    extendOptions = extendOptions || {}
    const Super = this
    ...
    Sub.prototype = Object.create(Super.prototype)
    ...
```
We can see that if we call `_Vue.extend(component)`, the `Super` will represent `Vue`  and `Sub.prototype` will extend from `Vue.prototype` rather than `component.prototype`, then the component prototype which we had overridden would be lost.

> I would like to revert this PR, because it stops us from being able to add assets to the component without mutating it.

Shall we add some unit test to explain and guarantee this scenario?

eddyerburgh:
>Shall we add some unit test to explain and guarantee this scenario?

Yes definitely, I made a mistake by merging this PR because no tests were failing. 

I'd also like to make sure your use case works as expected, but I can't see an easy solution.

kuitos:
uh... Could you show some code to explain what u wanna achieve? Actually I don't really know yet😂

eddyerburgh:
```js
const Constructor = vueVersion < 2.3 && typeof component === 'function'
? component.extend(instanceOptions)
: _Vue.extend(component)
```

kuitos:
Could you please provide a test case to describe it?

eddyerburgh:
```js
it('adds variables to extended components', () => {
  const TestComponent = Vue.extend({
    template: `
      <div>
        {{$route.path}}
      </div>
    `
  })
  const $route = { path: 'http://test.com' }
  const wrapper = mountingMethod(TestComponent, {
    mocks: {
      $route
    }
  })
  const HTML =
    mountingMethod.name === 'renderToString' ? wrapper : wrapper.html()
  expect(HTML).contains('http://test.com')
})
```

PR #868, opened 2018-07-29
feat: export createWrapper

### LINKED ISSUE #337: Support wrapping an already mounted Vue instance
The other day while testing a class that manually mounts a Vue instance to an element in the page, I wanted to use the `Wrapper` utilities but found out that I can't just call `new Wrapper(myVueInstance);` as it seems that `Wrapper` and `WrapperArray` are not exposed within the public API.

I didn't dive deep into the code, so I might have missed a way to achieve what I want.

A quick look at the source and the documentation shows that only the following is exposed:

```js
export default {
  createLocalVue,
  config,
  mount,
  shallow,
  TransitionStub,
  TransitionGroupStub
}
```

I found vuejs/vue-test-utils/pull/328 which is close to what I'd want.

eddyerburgh:
Closes #337

Closes #328

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -6,4 +6,5 @@
 !!!include(docs/api/renderToString.md)!!!
 !!!include(docs/api/selectors.md)!!!
 !!!include(docs/api/createLocalVue.md)!!!
+!!!include(docs/api/createWrapper.md)!!!
 !!!include(docs/api/config.md)!!!
--- a/docs/api/createWrapper.md
+++ b/docs/api/createWrapper.md
@@ -1,0 +1,19 @@
+## createWrapper()
+
+- **Returns:**
+  - `{Wrapper}`
+
+- **Usage:**
+
+`createWrapper` creates a `Wrapper` for a mounted Vue instance, or an HTML element.
+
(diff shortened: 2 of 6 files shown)


38elements:
There is no type definition of `createWrapper()`.

docs/api/README.md
```diff
@@ -6,4 +6,5 @@
 !!!include(docs/api/renderToString.md)!!!
 !!!include(docs/api/selectors.md)!!!
 !!!include(docs/api/createLocalVue.md)!!!
+!!!include(docs/api/createWrapper.md)!!!
```
38elements, line 9: It is necessary to add link to below page.
https://github.com/vuejs/vue-test-utils/blob/8db502dfacf56f5cfc8a914da192f8a43dd5a162/docs/README.md

eddyerburgh:
Thanks, I'm adding it in a new PR 👍

38elements:
I removed the comments in below page, since they may cause confusion.
https://github.com/vuejs/vue-test-utils/commit/ebca3b3160f4b3c09f100fa4bb5deffdfb41f486

PR #869, opened 2018-07-29
docs: add arguments to createWrapper.md

eddyerburgh:


docs/api/createWrapper.md
```diff
@@ -1,4 +1,11 @@
-## createWrapper()
+## createWrapper(node, options)
```
38elements, line 1: IMHO, I think it is necessary to add the test for this option argument.

PR #871, opened 2018-07-29
feat: add support for JSX and strings without a slot-scope attribute

### LINKED ISSUE #657: JSX $scopedSlots support

eddyerburgh:
- remove window dependency from scoped slots
- allow template tags as root node

cc @38elements 

Closes #657


packages/create-instance/create-scoped-slots.js
```diff
@@ -36,25 +36,12 @@ function getVueTemplateCompilerHelpers (): { [name: string]: Function } {
...
 
 function validateEnvironment (): void {
-  if (window.navigator.userAgent.match(/PhantomJS/i)) {
-    throwError(
-      `the scopedSlots option does not support PhantomJS. ` +
-        `Please use Puppeteer, or pass a component.`
-    )
-  }
   if (vueVersion < 2.5) {
     throwError(`the scopedSlots option is only supported in ` + `vue@2.5+.`)
```
38elements:
```
`the scopedSlots option is only supported in ` + `vue@2.5+.`
```

↓

```
`the scopedSlots option is only supported in vue@2.5+.`
```

test/specs/mounting-options/scopedSlots.spec.js
```diff
@@ -105,25 +105,6 @@ describeWithShallowAndMount('scopedSlots', mountingMethod => {
...
-      const message =
-        '[vue-test-utils]: the scopedSlots option does not support a template tag as the root element.'
-      expect(fn)
-        .to.throw()
-        .with.property('message', message)
-    }
-  )
-
   itDoNotRunIf(
     vueVersion >= 2.5 || isRunningPhantomJS,
```
38elements: `vueVersion >= 2.5 || isRunningPhantomJS,` -> `vueVersion >= 2.5,`

packages/create-instance/create-scoped-slots.js
```diff
@@ -36,25 +36,12 @@ function getVueTemplateCompilerHelpers (): { [name: string]: Function } {
 }
 
 function validateEnvironment (): void {
-  if (window.navigator.userAgent.match(/PhantomJS/i)) {
-    throwError(
-      `the scopedSlots option does not support PhantomJS. ` +
-        `Please use Puppeteer, or pass a component.`
-    )
-  }
   if (vueVersion < 2.5) {
```
38elements: If template tag in scopedSlots mounting options works normally,
I think this limitation is not necessary.

38elements:
I think it is necessary to change document.
https://github.com/vuejs/vue-test-utils/blob/e840892fb5dc2a6f68304f21e5603648c3b1a744/docs/api/options.md#scopedslots

packages/create-instance/create-scoped-slots.js
38elements: If template tag in scopedSlots mounting options works normally,
I think this limitation is not necessary.
I added this limitation since template tag in scopedSlots mounting options did not work.

https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots

> In 2.5.0+, slot-scope is no longer limited to the <template> element, but can instead be used on any element or component in the slot.

eddyerburgh: Yes you can add to either a <template> tag or any other tag

38elements: I see.
~~Since `vueVersion < 2.5`  limitation is not necessary, `validateEnvironment()` is not necessary.~~
`vueVersion < 2.5`  limitation is not necessary, `vueVersion < 2.1`  limitation is necessary.

> New in 2.1.0+

packages/create-instance/create-scoped-slots.js
```diff
@@ -36,25 +36,12 @@ function getVueTemplateCompilerHelpers (): { [name: string]: Function } {
...
 
-function validateTempldate (template: string): void {
-  if (template.trim().substr(0, 9) === '<template') {
-    throwError(
-      `the scopedSlots option does not support a template ` +
-        `tag as the root element.`
-    )
-  }
-}
+const scopedSlotRe = /<[^>]+ slot-scope=\"(.+)\"/
```
38elements: IMHO, I think `slotScopeRe` is better.

packages/create-instance/create-scoped-slots.js
```diff
@@ -32,46 +32,45 @@ function getVueTemplateCompilerHelpers (): { [name: string]: Function } {
...
   })
+  helpers.$createElement = vue._renderProxy.$createElement
   return helpers
 }
 
 function validateEnvironment (): void {
-  if (vueVersion < 2.5) {
-    throwError(`the scopedSlots option is only supported in ` + `vue@2.5+.`)
+  if (vueVersion < 2.1) {
+    throwError(`the scopedSlots option is only supported in ` + `vue@2.1+.`)
```
38elements: `the scopedSlots option is only supported in ` + `vue@2.1+.`

↓

`the scopedSlots option is only supported in vue@2.1+.`

packages/create-instance/create-scoped-slots.js
```diff
@@ -32,54 +32,45 @@ function getVueTemplateCompilerHelpers (): { [name: string]: Function } {
...
-    const domParser = new window.DOMParser()
-    const _document = domParser.parseFromString(template, 'text/html')
-    const slotScope = _document.body.firstChild.getAttribute(
-      'slot-scope'
-    )
-    const isDestructuring = isDestructuringSlotScope(slotScope)
-    scopedSlots[name] = function (props) {
-      if (isDestructuring) {
-        return render.call({ ...helpers, ...props })
+  for (const s in scopedSlotsOption) {
```
38elements: IMHO, I think `slotScopeValue` is better than `s`.

38elements: I think `slotName` is better then `s`.

PR #875, opened 2018-07-29
Improve types for mocks and provide in config options

cesalberca:
This PR allows users of TypeScript to use bracket notation without the need to cast to `any`.

Before:

```js
;(config.mocks as any)['foo'] = {
  bar: 'baz'
}
```

After:

```js
config.mocks['foo'] = {
  bar: 'baz'
}
```

Also, continuing work from this PR: https://github.com/vuejs/vue-test-utils/pull/874, I set  `allowSyntheticDefaultImports` to true in order to import default export without errors.


eddyerburgh:
Is this PR complete?

cesalberca:
Yes it is! However I'm still confused about the signaturee of `methods`: https://vue-test-utils.vuejs.org/api/#config. Is `config.methods`  an object that receives a function that returns an object? Or an object that receives a function that **declares some functions**?

eddyerburgh:
No, it's an object that's merged into component instance methods. The example is confusing, can you please replace it with this:

```
config.methods['getData'] = () => {}
```

cesalberca:
If it's an object then the docs are not correct as:

```
config.methods['errors'] = () => {
  any: () => false
}
```
Is equivalent to:

```
config.methods['errors'] = function() {
 // This is a labeled statement inside the function, not an object
 any: () => false
}
```

If we wanted an object the docs should be as follows:

```
config.methods['errors'] = () => ({
  any: () => false
})
```

Note the parenthesis.

eddyerburgh:
Yes that's why I asked if you could replace the example. As you said, it's incorrect.

cesalberca:
Aaah, alright then! I'll change it then

eddyerburgh:
Can you fix the linting error please? 

https://circleci.com/gh/vuejs/vue-test-utils/7489?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

cesalberca:
Done, ready to merge!

eddyerburgh (approved):
Great work, thanks :)

PR #876, opened 2018-07-29
docs: add TypeScript guide

### LINKED ISSUE #255: How to use typescript correctly?
I'm not sure if this issue belongs to this project. However, I'm using vue-test-utils since the beginning (even when its name was Avoriaz). But I have some issue to use SFC with typescript and Jest. I was wondering if you planned to write more documentation about which test runners or libs we should use to work properly with Typescript + Jest + Vue-test-utils?

Thank you!

eddyerburgh:
Closes #255


38elements:
It is necessary to add the link for `docs/guides/using-with-typescript.md`.
https://github.com/vuejs/vue-test-utils/blob/dev/docs/README.md

PR #889, opened 2018-08-02
fix: reconcile the overridden prototype of component with _Vue mixins

kuitos:
refer to https://github.com/vuejs/vue-test-utils/pull/856

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -126,9 +126,23 @@ export default function createInstance (
     component.options._base = _Vue
   }
 
+  function getExtendedComponent (component, instanceOptions) {
+    const extendedComponent = component.extend(instanceOptions)
+    // to keep the possible overridden prototype and _Vue mixins,
+    // we need change the proto chains manually
+    // @see https://github.com/vuejs/vue-test-utils/pull/856
+    // code below equals to
+    // `extendedComponent.prototype.__proto__.__proto__ = _Vue.prototype`
+    const extendedComponentProto =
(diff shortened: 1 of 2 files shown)


kuitos:
@eddyerburgh could you donate some time to review it?

eddyerburgh (approved):
Good solution, thanks :)

eddyerburgh:
I had to revert this change again before the release. It was causing issues with this new test—https://github.com/vuejs/vue-test-utils/blob/dev/test/specs/mounting-options/localVue.spec.js#L68.


The problem with modifying the prototype chain, or the super chain ourselves is that it's very difficult to plan for edge cases. So I think the solution would be to provide an option to not extend the component, although that would mean stubs and mocks won't be applied.

If you have other suggested workaround I'll take a look.

kuitos:
I had tried to change the root prototype of vue component:
```js
function getRootProto (obj) {
    if (Object.getPrototypeOf(obj) === Vue.prototype) {
      return obj
    }
    return getRootProto(Object.getPrototypeOf(obj))
}

function getExtendedComponent(component, instanceOptions) {
  const extendedComponent = component.extend(instanceOptions)
  const root = getRootProto(extendedComponent.prototype)
  Object.setPrototypeOf(root, _Vue.prototype)
  return extendedComponent
}

const Constructor = typeof component === 'function'
    ? getExtendedComponent(component, instanceOptions)
    : _Vue.extend(component).extend(instanceOptions)
```

I think I'm in the right direction, but the unit test what you added still failed.
I need to check what made it failed.

PR #893, opened 2018-08-03
feat: support scopedSlots mounting option for functional component

38elements:
This is the change to support scopedSlots mounting option for functional component.


eddyerburgh (approved):
Thanks :)

38elements:
@eddyerburgh 
Thank you for reviewing.

Issue #695, opened 2018-06-09
Feature: Allow `find` and `findAll` to work with tags


lmiller1990:
### What problem does this feature solve?
More intuitive API.

### What does the proposed API look like?

There is an open PR to add this API:

```js
wrapper = shallowMount(ComponentWithChild)

expect(wrapper.find("child-component").exists()).toBe(true)
```

Which lets users do `find("tag")` rather than `find({ name })` or `find(Component)`.

There is some objection (see #692 and it's previous PR, #681 ). Here is my thoughts why I like this better than the existing APIs:

1. `name` API: `({ name: ... })` doesn't feel very intuitive. We allow `querySelector` like syntax for `find`, `({ name })` doesn't conform to this. Also, `name` is not always the same as the `tag`, and has nothing to do with what is actually rendered, which is what `find` is used to assert.

2. `component` API: `find(Component)` is okay, but I don't really like having to import every component I want to `find` to every test. It's just a bunch of boilerplate. It's easy enough to make `shallowMount` and `mount` global, in the same way `describe` is, but this is not an option for components.

The least amount of typing and reduced boilerplate is `find({ name })`, but I don't like this that much. If the name is different for the component (it shouldn't be, but it happens) I have to go and find the `.vue` file, and check the name. Then if someone else looks at the test, they will likely have the same thought (we are looking for `({ name: foo-component })`, but actually this component only renders `<bar-component>`). 

We can still continue to allow for `{ name: 'foo' }` to be used, as well. **I am not proposing to remove and existing feature, simply extend it**. 

I don't want to drag this out for too long, if people have too strong an opinion we can just drop it. That's my thoughts.

eddyerburgh:
I think this is an intuitive way of finding a component, although I'd like to hear what other people think.

lmiller1990:
I have a new use case for this. When I stub an external dependency (at work we are using something called `vue-flatpickr` which wraps a javascript datepicker):

```js
const wrapper = shallowMount(Foo, {
  stubs: {
    flatpickr: true
  }
})
```

I cannot use `{ name: 'flatPickr' }` to select the component (I guess the name is something else, no way to know). I could go and look into `node_modules` and see the correct `name`, but I don't want to look at some external dependency just to be able to assert it was rendered. For now we did

```js
const wrapper = shallowMount(Foo, {
  stubs: {
    flatpickr: "<div id='flat' />"
  }
})
```

[...]

PR #897, opened 2018-08-04
Allow `find` to work with component tag

lmiller1990:
Working on #695 

Will most likely need some more test cases, but working nicely.

### DIFF
--- a/packages/shared/validators.js
+++ b/packages/shared/validators.js
@@ -1,5 +1,12 @@
 // @flow
-import { throwError, capitalize, camelize, hyphenate } from './util'
+import {
+  camelize,
+  capitalize,
+  htmlTags,
+  hyphenate,
+  svgElements,
+  throwError
+} from './util'
 
 export function isDomSelector (selector: any): boolean {
   if (typeof selector !== 'string') {
@@ -74,6 +81,20 @@ export function isNameSelector (nameOptionsObject: any): boolean {
   }
 
   return !!nameOptionsObject.name
+}
+
+export function isTagSelector (selector: any) {
+  if (typeof selector !== 'string') {
+    return false
+  }
+  const pseudoSelectors = ['.', '#', '[', ':', '>', ' ']
+  if (htmlTags.includes(selector) ||
+    svgElements.includes(selector) ||
+    selector.split('').some(char => pseudoSelectors.includes(char))) {
+    return false
+  } else {
+    return true
+  }
 }
 
 export function templateContainsComponent (
--- a/packages/shared/validators.js
+++ b/packages/shared/validators.js
@@ -1,6 +1,5 @@
 // @flow
-import intersection from 'lodash/intersection'
-import concat from 'lodash/concat'
+// import concat from 'lodash/concat'
 import {
   camelize,
   capitalize,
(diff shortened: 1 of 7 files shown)


packages/shared/validators.js
```diff
@@ -76,6 +83,20 @@ export function isNameSelector (nameOptionsObject: any): boolean {
...
 }
 
+export function isTagSelector (selector: any) {
+  if (typeof selector !== 'string') {
+    return false
+  }
+  const pseudoSelectors = ['.', '#', '[', ':', '>', ' ']
+  if (htmlTags.includes(selector) ||
+    svgElements.includes(selector) ||
+    selector.split('').some(char => pseudoSelectors.includes(char))) {
```
eddyerburgh: I think we should throw an error if the user is using a component selector with a pseudo selector, because they might expect this to work and be confused by find returning false.

eddyerburgh (changes requested):
Looks good :)

Can you add test cases for the following component types:

- functional components
- sub class components (`Vue.extend(options)`)
- extended components ({ extends: AnotherCmp })

Also, as I commented I think we should throw a selector if tag selectors are mixed with pseudo selectors. Could you also write a test for this.

lmiller1990:
Wow, fast review, thanks! Going to work on this comments now.

PS I saw you were coming to Vuefest in Tokyo in November, hopefully we can catch up then 👍

lmiller1990:
All cases are working except for functional components. Functional components don't seem to get `tag` or `$options._componentTag` assigned - maybe this feature cannot support functional components? (not that familiar with the main differences, other than the fact they are stateless).

Just fixing a few things and I'll push the updated tests (with failing functional one for now).

eddyerburgh:
Yes you're correct, I overlooked functional components. This is problematic, because it means we won't be able to replace the Constructor selector with the tag selector 😕

I've also thought of another potential problem: render functions that use the constructor (`{ render: h => h(TestComponent)}`). 

The reason I was keen to go ahead with this feature was because I thought we could replace constructor selectors, but I hadn't considered functional components or render functions that use component constructors.

Sorry about this, but we need to put the brakes on this PR to figure out a solution. I'd like to get this feature added, but I'm concerned about the discrepancy between functional and normal components

lmiller1990:
Functional components are tricky. I agree... the API should not require users to know about specifics, like the difference between functional and regular components. I will think about this more.

Throwing an error for cases like `find("a > my-component")` is also very tricky. There are so many cases to consider, such as:

1. hierarchical selectors: `a > my-component`
2. id/classes: `my-component#some-id`

But I think I this is not too difficult to solve, just more tricky than I originally expected. 

Another thing I am sure will lead to confusion is even though you search with for a tag, say `my-component`, if you inspect the markup using `wrapper.html` you will use `<mycomponent-stub>`. I think that is a bit unintuitive (although using a constructor or name also has this problem).

I haven't done much React, but I wonder how Enzyme handles `find`.

eddyerburgh:
Enzyme parse string selectors, which we could do. There's not much need right now since we require a DOM environment, and have the `Element.matches` method which does the work for us (for DOM elements).

For components, enzyme uses the constructor/ function. The React render tree node has a reference to the constructor for class components, or function, for functional components. This means matching components is a case of `selector === node.type` (not sure on exact prop).

eddyerburgh:
I think this feature would be too complex to implement, so we'll continue with the current selector API instead

PR #899, opened 2018-08-04
docs: fix guides/using-with-typescript.md

38elements:


docs/guides/using-with-typescript.md
```diff
@@ -90,7 +90,7 @@ In order to use TypeScript files in tests, we need to set up Jest to compile Typ
 $ npm install --save-dev ts-jest
 ```
 
-Next, we need to tell Jest to process JavaScript test files with `ts-jest` by adding an entry under `jest.transform` in `package.json`:
+Next, we need to tell Jest to process TypeScript test files with `ts-jest` by adding an entry under `jest.transform` in `package.json`:
```
lmiller1990, line 93: nice catch

lmiller1990 (approved):
Looks good to me.

eddyerburgh (approved):

PR #912, opened 2018-08-10
fix: reconcile the overridden prototype of component and _Vue mocks and mixins

kuitos:
fixed https://github.com/vuejs/vue-test-utils/pull/889
@eddyerburgh please make a review

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -1,5 +1,6 @@
 // @flow
 
+import Vue from 'vue'
 import { createSlotVNodes } from './create-slot-vnodes'
 import addMocks from './add-mocks'
 import { addEventLogger } from './log-events'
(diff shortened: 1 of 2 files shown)


kuitos:
@eddyerburgh could you please donate some time to review it?

eddyerburgh:
Yes, I was intending to look at it today 👍

packages/create-instance/create-instance.js
```diff
@@ -109,10 +110,35 @@ export default function createInstance (
...
+    const extendedComponent = component.extend(_Vue).extend(instanceOptions)
+
+    // cache subclass constructor
+    component.options._Ctor[extendedComponent.cid] = extendedComponent
+
+    // to keep the possible overridden prototype and _Vue mocks on prototype,
+    // we need change the proto chains manually
+    // @see https://github.com/vuejs/vue-test-utils/pull/856
+    const root = getRootVueProto(extendedComponent.prototype)
+    Object.setPrototypeOf(root, _Vue.prototype)
```
eddyerburgh, line 134: I believe this would mutate components:

```js
const CompA = Vue.extend()
const CompB = CompA.extend().extend()
const savedProto = CompA.prototype.__proto__
const rootProto = CompB.prototype.__proto__.__proto__
Object.setPrototypeOf(rootProto, null)

console.log(savedProto === CompA.prototype.__proto__) // false
```

kuitos: yes it could if you set the __proto__ of rootProto to be null.
But here we actually did was `Object.setPrototypeOf(rootProto, _Vue.prototype)`
```js
const rootProto = CompB.prototype.__proto__.__proto__
Object.setPrototypeOf(rootProto, Vue.prototype)

console.log(savedProto === CompA.prototype.__proto__) // true
```

eddyerburgh: If you set it to `_Vue` then I think it would still mutate the protoype, it just wouldn't be as severe:

```js
const CompA = Vue.extend()
const CompB = CompA.extend().extend()
const _Vue = Vue.extend()

const savedProto = CompA.prototype.__proto__
const rootProto = CompB.prototype.__proto__.__proto__
Object.setPrototypeOf(rootProto, _Vue.prototype)

console.log(savedProto === CompA.prototype.__proto__) // false
```

kuitos: I agree with what you concerned about, I will add an unit test and find a way to guarantee the super component immutably.
What do you think? Any suggestion? :)

eddyerburgh: That would be great, I don't have any suggestions unfortunately. I spent some time two weeks ago investigating, but I couldn't find a solution that didn't mutate the original components

Issue #887, opened 2018-08-01
Slots are not rendered in snapshot when they are in v-expansion-panel of Vuetify


trollepierre:
### Version
1.0.0-beta.22

### Reproduction link
[https://codesandbox.io/s/43k3520wmw](https://codesandbox.io/s/43k3520wmw)

### Steps to reproduce
Slots are not rendered in snapshot when they are in v-expansion-panel of Vuetify
```
<template>
  <div id="my-component">
    <v-expansion-panel>
      <v-expansion-panel-content>
        Outside slot
        <div slot="header">Pierre</div>
      </v-expansion-panel-content>
    </v-expansion-panel>
  </div>
</template>
```

### What is expected?
```
<div
  id="my-component"
>
  <v-expansion-panel-stub>
    <v-expansion-panel-content-stub
      expandicon="$vuetify.icons.expand"
    >
      
        Outside slot
        <div slot="header">Header</div>
    </v-expansion-panel-content-stub>
  </v-expansion-panel-stub>
</div>
```

### What is actually happening?
```
<div
  id="my-component"
>
  <v-expansion-panel-stub>
    <v-expansion-panel-content-stub
      expandicon="$vuetify.icons.expand"
    >
      
      Outside slot
      
    </v-expansion-panel-content-stub>
  </v-expansion-panel-stub>
</div>

```

eddyerburgh:
We currently only render default slots for stubbed components

The change you're requesting would mean that we render every named slot component. I'll investigate.

ferm10n:
I was interested in this myself, and here was my solution:

Modifying https://github.com/vuejs/vue-test-utils/blob/dev/packages/shared/stub-components.js#L96
```javascript
export function createBlankStub (
  originalComponent: Component,
  name: string
): Component {
  const componentOptions = typeof originalComponent === 'function'
    ? originalComponent.extendOptions
    : originalComponent
  const tagName = `${name}-stub`

  // ignoreElements does not exist in Vue 2.0.x
  if (Vue.config.ignoredElements) {
    Vue.config.ignoredElements.push(tagName)
  }

  return {
    ...getCoreProperties(componentOptions),
    render (h, context) {
      let innerContent = null
      if (context) {
        innerContent = context.children
      } else {
        innerContent = Object.keys(this.$slots).map(slotName => {
          const slot = this.$slots[slotName]
          if (slotName !== 'default') {
            return slot
          } else {
[...]

PR #915, opened 2018-08-10
Feat: add support for rendering named slots (#887)

ferm10n:


lmiller1990 (commented):
Any chance you can add a unit test? If not I might be able to do it. Didn't realize named slots wasn't supported.

ferm10n:
I plan on it, just haven't had the time to focus on it yet. #887 was closed too so my PR might not be needed. I'll check tomorrow

ferm10n:
So it looks like e604f1c fixes this. I like that solution a lot better than mine lol, so I'm closing.
However there seems to be some trouble when using following the example on  https://vuejs.org/v2/guide/components-slots.html#Named-Slots, which uses the template element for named slots.

JSDOM does not serialize template contents, in accordance with the [spec](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element). 

@eddyerburgh do you have any input on this, with regards to handling template elements?

eddyerburgh:
Can you give me an example in code of the problem, and raise a new issue so I can track it

ferm10n:
https://codesandbox.io/s/p4mmxj9mx (Learning how to use this tool, I didn't know how to pull in the latest version that is on the dev branch here)

I'm not entirely sure what should happen here to be honest, but I do imagine that it shouldn't be blank.

[Here's](https://github.com/inikulin/parse5/blob/master/packages/parse5/lib/serializer/index.js#L91) the reason why you'd get `<template></template>`. I misspoke earlier, parse5 is the cause, which is used by JSDOM.

Issue #905, opened 2018-08-06
it should be wrapper.vm.$route.path  not $router.path


Rudy24:
```javascript

import { shallowMount } from '@vue/test-utils'

const $route = {
  path: '/some/path'
}

const wrapper = shallowMount(Component, {
  mocks: {
    $route
  }
})

wrapper.vm.$router.path // /some/path

it should be wrapper.vm.$route.path  not $router.path

```

eddyerburgh:
Would you like to make a PR making the change?

Rudy24:
@eddyerburgh ok

PR #923, opened 2018-08-15
fix router name :it should be wrapper.vm.$route.path not $router.path

Rudy24:
```javascript

const $route = {
  path: '/some/path'
}

const wrapper = shallowMount(Component, {
  mocks: {
    $route
  }
})

wrapper.vm.$router.path // /some/path

it should be wrapper.vm.$route.path  not $router.path

```

[pls check issue link] (https://github.com/vuejs/vue-test-utils/issues/905)


eddyerburgh (approved):
Thanks :)

PR #926, opened 2018-08-15
Docs: Add propsData to Mounting Options

### LINKED ISSUE #917: Docs: include propsData in mounting options
While migrating my project from tests that I wrote while Vue was in V1, I experienced a bit of friction figuring out how to use `shallowMount` while setting prop data. It took me a bit to track down `propsData`, as I'd never actually needed it in an application setting.

I understand that `propsData` is actually handled through the `extends` API, but it seems like such a common use case for this library that documentation would be beneficial.

I think it would be beneficial to have a `propsData` link included under the 'Mounting Options' doc. It could be explained within there that this is actually the responsibility of `extends` API and not `vue-test-utils`.

I'm happy to PR a change to the docs with info about `propsData`.

sagalbot:
### Changes
- Add propsData documentation to mounting options 
- Add link to functional component definition in `context` description
- Fix broken link to `parentComponent`

### Notes
- I used the `::: tip` container, but I don't see it used elsewhere. Happy to remove that if preferred.

Closes #917

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -9,6 +9,7 @@ Options for `mount` and `shallowMount`. The options object can contain both Vue 
 - [`mocks`](#mocks)
 - [`localVue`](#localvue)
 - [`attachToDocument`](#attachtodocument)
+- [`propsData`](#propsdata)
 - [`attrs`](#attrs)
 - [`listeners`](#listeners)
 - [`parentComponent`](#parentComponent)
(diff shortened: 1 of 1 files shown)


docs/api/options.md
```diff
@@ -206,6 +207,32 @@ Component will be attached to DOM when rendered if set to `true`.
...
+const wrapper = mount(Component, {
+  propsData: {
+    msg: 'aBC'
+  }
+})
+expect(wrapper.text()).toBe('aBC')
+```
+
+::: tip 
+It's worth noting that `propsData` is actually a [Vue API](https://vuejs.org/v2/api/#propsData), not a proprietary
```
eddyerburgh: Great explanation, although can you remove the word proprietary, it sounds too formal. So:

`It's worth noting that `propsData` is actually a [Vue API](https://vuejs.org/v2/api/#propsData), not a `vue-test-utils` option.`

sagalbot: Roger that, done.

eddyerburgh (approved):
Thanks for this :)

PR #940, opened 2018-08-22
docs: Improve axios mock on async testing

udebella:
Hello, first time I open a pull request on an open source project, I may have missed something.

I use the library for like a month now and found a little improvement in the docs by using [`Promise.resolve`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) directly.

I made the same change in other languages too.
I did not bother creating an issue for a tiny change like that, but I can create one if you like.

### DIFF
--- a/docs/guides/testing-async-components.md
+++ b/docs/guides/testing-async-components.md
@@ -8,9 +8,7 @@ The implementation of the `axios` mock looks like this:
 
 ``` js
 export default {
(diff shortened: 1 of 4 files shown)


eddyerburgh (approved):
Thanks, this is definitely an improvement. I also write Promises like this in tests 👍

PR #941, opened 2018-08-23
Add support for getting prop, classes and attributes by key

willsoto:
Matches a similar interface from Enzyme:
  - http://airbnb.io/enzyme/docs/api/ShallowWrapper/prop.html
  - http://airbnb.io/enzyme/docs/api/ShallowWrapper/hasClass.html

Just a tiny bit of syntactic sugar to make getting a particular prop easier.

_NB: I didn't see an issue for this when I looked, but will be happy to make one if it's necessary._

### DIFF
--- a/docs/api/wrapper/attributes.md
+++ b/docs/api/wrapper/attributes.md
@@ -1,8 +1,11 @@
-## attributes()
+## attributes([key])
 
-Returns `Wrapper` DOM node attribute object.
+Returns `Wrapper` DOM node attribute object. If `key` is provided, the value for the `key` will be returned.
 
-- **Returns:** `{[attribute: string]: any}`
+- **Arguments:**
+  - `{string} key`  **optional**
+
+- **Returns:** `{[attribute: string]: any} | string`
 
 - **Example:**
 
@@ -12,4 +15,5 @@ import Foo from './Foo.vue'
 
 const wrapper = mount(Foo)
 expect(wrapper.attributes().id).toBe('foo')
+expect(wrapper.attributes('id')).toBe('foo')
 ```
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -78,12 +78,15 @@ export default class Wrapper implements BaseWrapper {
   /**
    * Returns an Object containing all the attribute/value pairs on the element.
    */
-  attributes (): { [name: string]: string } {
+  attributes (key?: string): { [name: string]: string } | string {
     const attributes = this.element.attributes
     const attributeMap = {}
     for (let i = 0; i < attributes.length; i++) {
       const att = attributes.item(i)
       attributeMap[att.localName] = att.value
+    }
+    if (key) {
+      return attributeMap[key]
     }
     return attributeMap
   }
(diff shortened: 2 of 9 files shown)


eddyerburgh:
Thanks for this, I've been intending to add it for a while! Could you also update `classes`, and `attributes` in this PR to follow this pattern? I'd like to keep the APIs in sync.

willsoto:
Sure! `attributes` seems pretty obvious. For `classes` the return value is `string[]`, so do you prefer `indexOf` or `find` or something else?

eddyerburgh:
Great :)

`indexOf`, because `Array.prototype.find` isn't supported by IE or phantom JS

willsoto:
@eddyerburgh and what should the return value for `classes` be, `boolean`? Would it behave more like `hasClass`?

willsoto:
@eddyerburgh I assumed `classes('name')` should return a boolean. If not, it's easy enough to change 😄 

I successfully used function overloading for the `.d.ts` file, but I am not familiar enough with Flow to know if I did the types there correctly. As a result, I ended up having to alter some functions that utilized `attributes()` in order to satisfy the type checker. Ideally, I'd just function overload there as well and not have to rewrite any code, so looking for suggestions on that. 

Thanks!

docs/api/wrapper/attributes.md
```diff
@@ -13,3 +13,11 @@ import Foo from './Foo.vue'
 const wrapper = mount(Foo)
 expect(wrapper.attributes().id).toBe('foo')
 ```
+
```
eddyerburgh: I think we should just move the assertion to the above example, no need to duplicate the set up.

eg:

```
import Foo from './Foo.vue'	 import Foo from './Foo.vue'
 	 
const wrapper = mount(Foo)	const wrapper = mount(Foo)
expect(wrapper.attributes().id).toBe('foo')
expect(wrapper.attributes('id')).toBe('foo')
```

eddyerburgh: Same for the rest of the docs pages

packages/test-utils/src/wrapper.js
```diff
@@ -112,9 +115,17 @@ export default class Wrapper implements BaseWrapper {
         }
       })
       classes = classes.map(
-        className => cssModuleIdentifiers[className] || className
+        klass => cssModuleIdentifiers[klass] || klass
```
eddyerburgh: Why the name change? Can you replace it with `c` rather than `klass`

packages/test-utils/src/wrapper.js
```diff
@@ -766,7 +784,8 @@ export default class Wrapper implements BaseWrapper {
    */
   setSelected (): void {
     const tagName = this.element.tagName
-    const type = this.attributes().type
+    const attributes = this.attributes()
+    const type = typeof attributes === 'object' ? attributes.type : ''
```
eddyerburgh: Was this to fix flow errors? If so you can just add `// $FlowIgnore` above the original line. We know it's never going to return a string, so there's no need to guard against it.

We're actually going to replace flow with TypeScript soon, which would fix this issue. It's better to add flow ignores now rather than add extra code to satisfy flow.

packages/test-utils/src/wrapper.js
willsoto: `className` ended up shadowing the argument name so I ended up changing it here since it isn't user facing. I am not sure you can use `class` since it is a reserved word. How about `name`?

willsoto:
@eddyerburgh Thanks for the feedback. It should all be addressed.

eddyerburgh:
Sorry there was another comment that I thought I'd added—can you add an `arguments` section to the docs, like find—https://vue-test-utils.vuejs.org/api/wrapper/find.html?

willsoto:
@eddyerburgh done and rebased 😄

eddyerburgh (approved):
Thanks for this :)

Issue #967, opened 2018-09-12
Mocking named slots is not working


kieselsteini:
### Version
1.0.0-beta.25

### Reproduction link
[https://github.com/kieselsteini/vue-named-slot-issue](https://github.com/kieselsteini/vue-named-slot-issue)

### Steps to reproduce
- yarn install
- yarn test:unit

### What is expected?
- named slot "header" should contain "Hi, I am the header slot!"
- default slot should contain "Hi, I am the default slot!"

### What is actually happening?
- default slot contains both
```html
<div class="hello"><header></header> Hi, I am the default slot!Hi, I am the header slot!</div>
```

PR #979, opened 2018-09-23
fix: render all slots inside a <template> vnode

eddyerburgh:
- Wrap all slot vnodes created from a string in a template vnode

Fixes https://github.com/vuejs/vue-test-utils/issues/967

### DIFF
--- a/packages/create-instance/create-slot-vnodes.js
+++ b/packages/create-instance/create-slot-vnodes.js
@@ -4,9 +4,12 @@ import { compileToFunctions } from 'vue-template-compiler'
 
 function createVNodes (
   vm: Component,
-  slotValue: string
+  slotValue: string,
+  name
 ): Array<VNode> {
-  const el = compileToFunctions(`<div>${slotValue}</div>`)
+  const el = compileToFunctions(
+    `<div><template slot=${name}>${slotValue}</template></div>`
+  )
   const _staticRenderFns = vm._renderProxy.$options.staticRenderFns
   const _staticTrees = vm._renderProxy._staticTrees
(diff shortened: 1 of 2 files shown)


packages/create-instance/create-slot-vnodes.js
```diff
@@ -4,39 +4,32 @@ import { compileToFunctions } from 'vue-template-compiler'
...
-    vnode = vm.$createElement(slotValue)
-  }
-  if (vnode.data) {
-    vnode.data.slot = name
-  } else {
-    vnode.data = { slot: name }
+    return createVNodes(vm, slotValue, name)
   }
+  const vnode = vm.$createElement(slotValue)
+  ;(vnode.data || (vnode.data = {})).slot = name
```
lmiller1990, line 32: what does this line do?

eddyerburgh: It sets `vnode.data.slot` to equal the slot name. It's shorthand for:

```js
if(!vnode.data) {
  vnode.data = {}
}
vnode.data.slot = name
```

lmiller1990: I see! haven't seen this syntax before. I take it the preceeding `;` is required.

eddyerburgh: Yes, that's is one of the drawbacks of not including semi-colons, there are a few other cases where no semi-colons will cause errors.

lmiller1990: Fair enough.

Code looks good outside that, good job.

PR #1016, opened 2018-11-01
docs(ja): fix example command

numb86:
Correctly it is `test`, not `unit`. I think so.
The corresponding English page is also `test`.
https://vue-test-utils.vuejs.org/guides/testing-single-file-components-with-mocha-webpack.html


kazupon (approved):

eddyerburgh (approved):
Thank you

PR #1017, opened 2018-11-02
Update testing-single-file-components-with-jest.md

mighty-peach:
Изменил имя пакета vue-test-utils -> @vue/test-utils
Change package name vue-test-utils -> @vue/test-utils


eddyerburgh (approved):
Thanks 👍

Issue #150, opened 2017-11-03
Trigger events from child vue component


Aferz:
Hi.

I would like to be able to trigger events from children components of a given parent. Check this code out:

##### Parent Component (Only template):
```html
<div class="parent">
    <child class="child-component"
      @custom-event="console.log.call(null, 'Custom Event from Child Component')" 
      @click="console.log.call(null, 'Click Event from Child Component')">
    </child>

    <div class="child-div"
      @custom-event="console.log.call(null, 'Custom Event from Native DOM Element')" 
      @click="console.log.call(null, 'Click Event from Native DOM Element')" >
    </div>
</div>
```

##### Test:
```javascript
test('Testing triggers', () => {
    const wrapper = mount(Parent) // Parent Vue Component
    
    const childComponent = wrapper.find('.child-component') // Child Vue Component
    const childDiv = wrapper.find('.child-div') // Child Native DOM Element
    
    childComponent.trigger('custom-event')
    childDiv.trigger('custom-event')
    
    childComponent.trigger('click')
    childDiv.trigger('click')
})
```

The log generated in console is the following:

```
Custom Event from Native DOM Element
Click Event from Native DOM Element
```

As you can see, events triggered in Vue Components are not being emitted, so the parent can't catch them and act in consequence.

Is this expected? Maybe I'm doing something wrong ?
Thanks !

lmiller1990:
This will not work, when you do `mount` it will not stub the child component, but render it's actual contents, I think. Can you post `child-component`?

What you probably want to do is instead of `find('.child-component').trigger('custom-event')`, do `trigger` on the actual button in the child component (maybe give it an ID or something). So:

``` js
// child component
<template>
  <button id="child" @click="$emit('custom-event')">Btn</button>
</template>
```

``` js
// parent
<template>
  <child-component @custom-event="something" />
</template>
```

``` js
// Test 
const wrapper = mount(Parent)
wrapper.find('#child').trigger('click')
// this should work (or something like this)
```

If you post `child-component`, or better yet your repo, I can add some examples of how to do this (I have done it a lot in my projects, and had the same problem, but haven't got any snippet I can share at this very moment).

Will try and put a small demo repo together later today, since I remember stuggling a little to test parent/child events.

Aferz:
Mmmh I think I Understand...

My problem is that I'm trying to do TDD and what I was attempting to do is create a Parent component that needs to react to events emitted from a yet undefined Child Component.

[...]

PR #1018, opened 2018-11-02
docs: add emitting event from child component section

VictorCazanave:
As mentioned in https://github.com/vuejs/vue-test-utils/issues/150, it adds a tips section to show how to emit a custom event from a child component


eddyerburgh (approved):
Thanks

PR #1022, opened 2018-11-09
Fix scoped slots typing

healqq:
Scoped slot attribute accepts functions, but typing allows only ```Record<string, string>```
Updated typings, added test for scoped slots types


eddyerburgh (approved):
Thanks 👍

PR #1029, opened 2018-11-18
feat: undefined attributes parsed as $attrs

abilicz:
making setProps method more consistent with the actual Vue behavior.


eddyerburgh (approved):
Thanks for this :)

PR #1038, opened 2018-11-24
feat: create stubs in createElement

### LINKED ISSUE #973: Warning messages thrown about child components created with `vue.extend` even when parent component is mounted with `shallowmount`
### Version
1.0.0-beta.25

### Reproduction link
[https://github.com/vuejs/vue-test-utils/blob/0d1ddd1e52bc66ed4fb7a1388be612f874dc9c57/packages/create-instance/extend-extended-components.js#L42-L47](https://github.com/vuejs/vue-test-utils/blob/0d1ddd1e52bc66ed4fb7a1388be612f874dc9c57/packages/create-instance/extend-extended-components.js#L42-L47)

### Steps to reproduce
`ShallowMount` a component that has child components created with vue.extend

### What is expected?
as shallow mount automatically stubs all child components there should be no warning about how child components has been modified

Getting these warnings.

`[vue-test-utils]: The child component xyz has been modified to ensure it is created with properties injected by Vue Test Utils.
    This is because the component was created with Vue.extend, or uses the Vue Class Component decorator.
    Because the component has been modified, it is not possible to find it with a component selector. To find the component, you must stub it manually usi
ng the stubs mounting option, or use a name or ref selector.
    You can hide this warning by setting the Vue Test Utils config.logModifiedComponents option to false.`

### LINKED ISSUE #994: Unknown custom element: <router-link> when using `mount()`
### Version
1.0.0-beta.25

### Steps to reproduce
Component `Test.js`, which uses another component with `router-link` inside:

```js
import Vue from 'vue'
import Component from 'vue-class-component'

const Card = Component(class Card extends Vue {
  render () {
    return <router-link to='test'>Test</router-link>
  }
})

export default {
  render () {
    return <Card />
  }
}
```

Given the following test:

```js
// @flow
/* eslint-env jest */

import { mount, createLocalVue, RouterLinkStub } from '@vue/test-utils'
import VueRouter from 'vue-router'

import Test from './Offices'

const localVue = createLocalVue()
localVue.use(VueRouter)

describe('Test.js', () => {
  it('should render with a stub', () => expect(mount(Test , {
    localVue,
    stubs: { 'router-link': RouterLinkStub }
  }).element).toBeDefined())

  it('should render with the router', () => {

    expect(mount(Test , {
      localVue
    }).element).toBeDefined()
  })
})
```

### What is expected?
In both test cases `router-link` in `OfficeCard` should be stubbed and tests should pass without errors.

Instead, the following error will be logged to the console for both cases:

```
[Vue warn]: Unknown custom element: <router-link> - did you register the component correctly? For recursive components, make sure to provide the "name" option.
```

---
The issue is definitely related to the `vue-class-component` since the following code will pass without errors:

```js
import Vue from 'vue'
import Component from 'vue-class-component'

const Card = {
  render () {
    return <router-link to='test'>Test</router-link>
  }
}

export default {
  render () {
    return <Card />
  }
}
```

I assume it is because it directly extends `Vue` which prevents `test-utils` from effectively stubbing underlying components.

### LINKED ISSUE #995: Using a <component> tag does not stub the component when using shallowMount
### Version
1.0.0-beta.25

### Reproduction link
[https://codesandbox.io/s/wq7zw0k288](https://codesandbox.io/s/wq7zw0k288)

### Steps to reproduce
Open project and run test

### What is expected?
No text should appear because all the components are stubbed

The 'normal' component is stubbed, but all others are not not

---
I'm using `<component :is="aComponent" />` to dynamically choose question components on forms based on other data. However because shallow mounting is not stubbing them the entire form is being created when I just want to test the wrapper.

This may be a similar problem to https://github.com/vuejs/vue-test-utils/issues/959

eddyerburgh:
- Stub components during $createElement call. This way we will always stub components, including dynamically added components
- Extend components during $createElement call. This avoids the previous recursive lookup, and ensures we only extend components that are rendered
- Use tag in stub name

- Fixes #973
- Fixes #994
- Fixes #995

*Breaking change*: this will change the output of some stubs


afv:
Hello!

After this change I'm getting stubbed components on mounted (not shallow) components if the component was previously shallow mounted (on another test).

```javascript
it('renders the component without stubs', () => {
  const wrapper = mount(TheComponent, { store, localVue, propsData: {…} });
  // The component has no stubs, as expected
}

it('renders the component again without stubs', () => {
  const wrapper = mount(TheComponent, { store, localVue, propsData: {…} });
  // The component has no stubs, as expected
}
```

But:

```javascript
it('renders the component with stubs', () => {
  const wrapper = shallowMount(TheComponent, { store, localVue, propsData: {…} });
  // The component has stubs, as expected
}

it('renders the component with stubs, unexpected', () => {
  const wrapper = mount(TheComponent, { store, localVue, propsData: {…} });
  // The component has stubs too!
}
```

Why? :confused:

eddyerburgh:
The bug has been reported—https://github.com/vuejs/vue-test-utils/issues/1050

Follow that thread for updates

PR #1052, opened 2018-12-03
fix: Fix references to documents.

numb86:
I fixed the fragment identifier of the link of warn message.

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -233,7 +233,7 @@ export default class Wrapper implements BaseWrapper {
     warn(
       `hasAttribute() has been deprecated and will be ` +
       `removed in version 1.0.0. Use attributes() ` +
-      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#attributes`
+      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#attributes-key`
     )
 
     if (typeof attribute !== 'string') {
@@ -258,7 +258,7 @@ export default class Wrapper implements BaseWrapper {
     warn(
       `hasClass() has been deprecated and will be removed ` +
       `in version 1.0.0. Use classes() ` +
-      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#classes`
+      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#classes-classname`
     )
     let targetClass = className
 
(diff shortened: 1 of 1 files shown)


packages/test-utils/src/wrapper.js
```diff
@@ -233,7 +233,7 @@ export default class Wrapper implements BaseWrapper {
     warn(
       `hasAttribute() has been deprecated and will be ` +
       `removed in version 1.0.0. Use attributes() ` +
-      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#attributes`
+      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#attributes-key`
```
eddyerburgh:
```suggestion
      `instead—https://vue-test-utils.vuejs.org/api/wrapper/attributes.html`
```

packages/test-utils/src/wrapper.js
```diff
@@ -258,7 +258,7 @@ export default class Wrapper implements BaseWrapper {
     warn(
       `hasClass() has been deprecated and will be removed ` +
       `in version 1.0.0. Use classes() ` +
-      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#classes`
+      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#classes-classname`
```
eddyerburgh:
```suggestion
      `instead—https://vue-test-utils.vuejs.org/api/wrapper/classes.html`
```

packages/test-utils/src/wrapper.js
```diff
@@ -285,7 +285,7 @@ export default class Wrapper implements BaseWrapper {
     warn(
       `hasProp() has been deprecated and will be removed ` +
       `in version 1.0.0. Use props() ` +
-      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#props`
+      `instead—https://vue-test-utils.vuejs.org/api/wrapper/#props-key`
```
eddyerburgh:
```suggestion
      `instead—https://vue-test-utils.vuejs.org/api/wrapper/props.html`
```

eddyerburgh (changes requested):
Thanks, but those URLs are a problem, because they include the method arguments. Can you please use the suggested URLs instead

numb86:
I appreciate your gently advice.
I fixed it, so please review again.

eddyerburgh (approved):
Thanks :)

PR #1055, opened 2018-12-04
fix: Support extended functional components

kellym:
This fixes an issue where a component that extends another functional component is not treated as a functional component by vue-test-utils.

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -11,7 +11,11 @@ import {
 } from 'shared/compile-template'
 import extractInstanceOptions from './extract-instance-options'
 import createFunctionalComponent from './create-functional-component'
-import { componentNeedsCompiling, isPlainObject } from 'shared/validators'
+import {
+  componentNeedsCompiling,
+  componentHasProperty,
+  isPlainObject
+} from 'shared/validators'
 import { validateSlots } from './validate-slots'
 import createScopedSlots from './create-scoped-slots'
 import { createStubsFromStubsObject } from './create-component-stubs'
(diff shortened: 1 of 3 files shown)


kellym:
I guess an example would be helpful. I ran into this via this issue: 

```
// MyFunctionalComponent
export default {
  extends: BaseFunctionalComponent
}
```
```
// BaseFunctionalComponent
export default {
  functional: true
}
```

When mounting `MyFunctionalComponent`, it would throw an error when using `context` saying it wasn't functional.

kellym:
Just updated to account for `mixins`, since those are inherited similarly to `extends`.

eddyerburgh (changes requested):
The change looks great, but can you add two test cases for the change? That way we avoid regressions!

kellym:
You know, I had tests in the really old branch I originally wrote this, but I couldn't find where the `isVueComponent` tests, etc, went since I rebased.  Where should I put the tests?

eddyerburgh:
You could put them in `test/specs/wrapper/find.spec.js`

kellym:
I ended up adding them in what seemed to be more targeted:  `test/specs/mounting-options/context.spec.js`.

eddyerburgh:
Yes that's a more relevant file!

Can you also make sure the tests test that it renders with the correct context, maybe by trying to call a method that will only exist if the context is passed correctly

kellym:
Sorry for the delay on this PR. I think at some point during writing tests I found that this solution doesn't solve it across the board -- there was still a scenario where it was not treated as a functional component.

lmiller1990:
Hey @kellym - sorry for the slow reply, do you have an example of the case that this _doesn't_ cover by any chance?

PR #1057, opened 2018-12-07
docs: Use classes instead of hasClass.

numb86:
`hasClass()` has been deprecated.
Used `classes()` instead.


eddyerburgh (approved):
Thank you :)

Issue #827, opened 2018-07-16
this.$children is empty in the created lifecycle hook


KyleMayes:
### Version
1.0.0-beta.20

### Reproduction link
[https://codesandbox.io/s/q89qro9yq?module=%2Fsrc%2Fcomponents%2FTabs.spec.js](https://codesandbox.io/s/q89qro9yq?module=%2Fsrc%2Fcomponents%2FTabs.spec.js)

### Steps to reproduce
1. Run the unit tests (click on "Tests" at the bottom)

### What is expected?
The `mount` function in the test will render the tabs component as it is seen in the browser on the right, with three `li` elements in the `ul` element representing the three child tab components.

### What is actually happening?
In the tabs component, there is a created lifecycle hook:

```
created() {
  this.tabs = this.$children;
}
```

This works in the browser, but `$children` is empty when rendered by `vue-test-utils`.
This results in the `li` elements (which are produced by iterating over this `tabs` data field) are not rendered.

eddyerburgh:
`$children` is empty when a component initializes, and child instances are pushed to it when they are created.

This is however a bug with synchronous behavior. You can either fix it by setting `sync` to false and using `setTimeout` to wait for the next tick:

```js
it("renders with content", (done) => {
 // ..
  const tabs = mount(Tabs, { localVue, slots: { default: content }, sync: false });
  setTimeout(() => {
    expect(tabs.html()).toContain("li");
    done()
  })
});
```

Or force the component to re render:

```js
it("renders with content", () => {
  // ..
  const tabs = mount(Tabs, { localVue, slots: { default: content } });
  tabs.vm.$forceUpdate()
  expect(tabs.html()).toContain("li");
});
```

eddyerburgh:
I've released beta.29 which uses Vue `async` mode to render synchronously, but it causes a bug with this use case. 

Since this is an edge case, and there is a workaround we won't fix this issue.

The solution going forward is to set `sync` to `false`.

PR #1062, opened 2018-12-10
fix: use Vue async option for sync mode

### LINKED ISSUE #532: $listeners is readonly error
### Version
1.0.0-beta.14

### Reproduction link
[https://codesandbox.io/s/0y2o4p8zvv](https://codesandbox.io/s/0y2o4p8zvv)

### Steps to reproduce
Visit reproduction link.

### What is expected?
Should not warn

Warn that $listerns and $attrs is readonly

### LINKED ISSUE #594: Can't trigger a second click on a Vuetify checkbox.
### Version
1.0.0-beta.15

### Reproduction link
[https://codesandbox.io/s/8yypln8828](https://codesandbox.io/s/8yypln8828)

### Steps to reproduce
File Editor > src > components > HelloWorld.spec.js

The spec includes:
- Test on a single Vuetify checkbox (passes).
- Test on a dynamic checkbox (passes).
- Test on a dynamic Vuetify checkbox (fails).

### What is expected?
When a click event is triggered on the Vuetify checkbox, the related model should change from `undefined` to `true`. Then when the second click event is triggered on the same Vuetify checkbox, the related model should change from `true` to `false`.

When a click event is triggered on the Vuetify checkbox, the related model changes from `undefined` to `true`. However, when the second click event is triggered on the same Vuetify checkbox, the related model doesn't change from `true` to `false` like you'd expect.

---
I ran into this issue when upgrading from `1.0.0-beta.12` (where it worked as expected) to `1.0.0-beta.15`. So my assumption is that this is at least in someway related to `vue-test-utils`, and maybe not a `Vuetify` specific issue. Or maybe there's a preferred way to accomplish this, and my method just so happened to work in `1.0.0-beta.12`, but not anymore.

Here's the actual component:
https://github.com/vuetifyjs/vuetify/tree/dev/src/components/VCheckbox

### LINKED ISSUE #676: Fix synchronous behavior

### LINKED ISSUE #829: "Cannot read property '_transitionClasses' of undefined" when injecting vee-validate which is not used
### Version
1.0.0-beta.20

### Reproduction link
[https://cli.vuejs.org/guide/prototyping.html](https://cli.vuejs.org/guide/prototyping.html)

### Steps to reproduce
I have this strange error when I'm running a test on a component that's using vee-validate. What's even weirder is that **it doesn't happen when I actually use `this.errors` in that component**.
`this.errors` is an ErrorBag added by vee-validate, and it makes the error disappear even if I just add a dummy line like `console.log(this.errors);` or `this.errors;` into the tested component's created hook.

Test code:

    import { mount } from '@vue/test-utils';
    import { renderToString } from '@vue/server-test-utils';
    import Vue from 'vue';
    import VeeValidate from 'vee-validate';
    import VeeValidateConfig from '@/vee-validate-config';
    Vue.use(VeeValidate, VeeValidateConfig);
    import FolderSelect from '@/components/FolderSelect';

    describe('FolderSelect', () => {
    let wrapperDeep;

    beforeEach(() => {
        wrapperDeep = mount(FolderSelect);
    });

    it('renders select box if option "existing folder" is selected', () => {
        // this code is forcing component to use vee-validate:
        wrapperDeep.setData({folder: 'existing'}); 
    });

Output from `yarn test:unit`:

>     [Vue warn]: Error in directive validate bind hook: 
>     "TypeError: Cannot read property '_transitionClasses' of undefined"

It's coming from `node_modules/vue/dist/vue.runtime.common.js`, line 1739 and 589.

After I add to tested component:

    created () {
        console.log(this.errors);
    },
The error is gone! Why the error appears otherwise? Is Vue clearing vee-validate if it's not used and in result breaking these transitions? Doesn't help if I add `{{ errors }}` to the template, though.

I'm injecting vee-validate like this:

    export default {
        inject: ['$validator'],
        (...)

Not sure if this error is caused by Vue, vue-test-utils or vee-validate, but since it doesn't throw this error on dev and production build, it must be sth with vue-test-utils.

### What is expected?
The error should be gone.

I get "TypeError: Cannot read property '_transitionClasses' of undefined"

### LINKED ISSUE #1023: Vuelidate Error in render: “TypeError: Cannot read property 'proxy' of undefined”
### Version
1.0.0-beta.20

### Reproduction link
[https://github.com/ParsonsProjects/vue-cli-sandbox/tree/jest-vuelidate](https://github.com/ParsonsProjects/vue-cli-sandbox/tree/jest-vuelidate)

### Steps to reproduce
RUN `npm install`

RUN `npm run test:unit`

THEN a warning will be thrown in the test results.

### What is expected?
WHEN i run `npm run test:unit`

THEN no error is thrown by vuelidate

WHEN i run `npm run test:unit`

THEN the error is being thrown by vuelidate

---
This error is only being thrown when running the unit tests and not when running `npm run serve` which might be something to do with how the lifecycle of data is being handled when running tests. I am not 100% sure about this. 

When I was trying to debug the issue it looked like the new item was not being added to the array before vuelidate was looking for the new item.

Any help would be greatly appreciated. Thanks

### LINKED ISSUE #1034: False positive prop mutation error
### Version
1.0.0-beta.25

The issue is also inside `1.0.0-beta.20` which is the real version we use, but we cannot upgrade to `.25` due to https://github.com/vuejs/vue-test-utils/issues/1026, but the bug is inside `.25` as well. Reproduction repo uses `.25`.

### Reproduction link
[https://github.com/MechJosh0/vue-test-utils-prop-error](https://github.com/MechJosh0/vue-test-utils-prop-error)

### Steps to reproduce
```bash
npm run test:unit
```

Output
```bash
PASS  src/components/__tests__/Container.spec.js
  @/components/Container
    interaction
      √ it should display a child-block (33ms)

  console.error node_modules/vue/dist/vue.runtime.common.js:589
    [Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "mode"
    
    found in
    
    ---> <SimpleButton>
           <Anonymous>
             <Root>
```

### What is expected?
No error.

Producing an error saying that the prop `mode` is being mutated when it is in fact not.

### LINKED ISSUE #1035: Re-render does not occur synchronously when reactive data is changed in mounted life-cycle hook
### Version
1.0.0-beta.25

### Reproduction link
[https://github.com/Aaron-Pool/vue-test-utils-jest-example](https://github.com/Aaron-Pool/vue-test-utils-jest-example)

### Steps to reproduce
1. clone reproduction repo
2. `Run yarn`
3. run `yarn jest`

### What is expected?
Test should pass, because reactive data changed in the mount component, so the scoped slot function should have been called again on re-render.

No re-render occurs, even though the data updates as expected.

---
I assume this has something to do with async vs. sync paradigm that is happening in the test-utils.

It works correctly if I run the test like so:

```js
it('should initialize with dimension data', done => {
  const mockSlot = jest.fn();
  const comp = factory(mockSlot);
  comp.vm.$nextTick(() => {
    expect(mockSlot.mock.calls).toEqual([[{ width: null, height: null }], [{ width: 500, height: 500 }]]);
    done();
  })
});
```

### LINKED ISSUE #1046: setProps performance issues leads to RangeError: Maximum call stack size exceeded
### Version
1.0.0-beta.26

### Reproduction link
[https://github.com/AlexandreBonaventure/repro-test-utils-setProps](https://github.com/AlexandreBonaventure/repro-test-utils-setProps)

### Steps to reproduce
`git clone https://github.com/AlexandreBonaventure/repro-test-utils-setProps` > `yarn install` > `yarn test:unit`

### What is expected?
No performance issue across all tests

Performance is affected when using more and more messages in vue-i18n

---
Hello, 
I stumbled upon a performance issue when using wrapper.setProps and vue-i18n. For some reasons,  tests are running slower and slower with the increased amount of entries (0 -> 3000 lines) in vue-i18n messages (translations), eventually ending up in a RangeError. I know some people will tell me that I could either mock or get rid of vue-i18n in my test, but it is very convenient to do so to track untranslated messages in a vue app.  At my company we are using it extensively. Moreover, I don't think we should experience this kind a performance issue anyway and that let me believe this is a code smell in the way test-utils reorders dependencies. 

### with 3000 lines messages

NB: using a wrapper factory instead of setProps is fixing the issue as well.
Thanks

### LINKED ISSUE #1125: RangeError: Maximum call stack size exceeded when use vue-i18n and i18n large json
### Version
1.0.0-beta.28

### Reproduction link
[https://github.com/yoyo930021/vue-i18n-test-utils-reproduction](https://github.com/yoyo930021/vue-i18n-test-utils-reproduction)

### Steps to reproduce
I use vue-i18n and add vue-i18n to vue-test-utils in jest.
Use more large json in i18n.
And more `it` function 

AppNumberSupporter.spec.ts:
https://github.com/yoyo930021/vue-i18n-test-utils-reproduction/blob/master/tests/unit/AppNumberSupporter.spec.ts

i18n.ts:
https://github.com/yoyo930021/vue-i18n-test-utils-reproduction/blob/master/src/i18n.ts

zh-CN.json:
https://github.com/yoyo930021/vue-i18n-test-utils-reproduction/blob/master/src/locales/langs/zh-CN.json

run `yarn test:unit`

### What is expected?
```
  ● AppNumberSupporter.vue › return data when click odd

    RangeError: Maximum call stack size exceeded
        at Array.forEach (<anonymous>)

      at orderDeps (node_modules/@vue/test-utils/dist/vue-test-utils.js:2553:16)
          at Array.forEach (<anonymous>)
      at node_modules/@vue/test-utils/dist/vue-test-utils.js:2558:14
          at Array.forEach (<anonymous>)
      at orderDeps (node_modules/@vue/test-utils/dist/vue-test-utils.js:2553:16)
          at Array.forEach (<anonymous>)
      at node_modules/@vue/test-utils/dist/vue-test-utils.js:2558:14
```

and more duplicate error.

PASS Test.

---
I will try `sync: false`.
Better than not use, but unsolved.

eddyerburgh:
This PR fixes many bugs caused by the current synchronous implementation by using an async option added to Vue 2.5.18. 

- closes #676
- fixes #1125 
- fixes #1046
- fixes #1035
- fixes #1034 
- fixes #1023 
- fixes #829 (https://github.com/baianat/vee-validate/issues/1267)
- fixes #594
- fixes #532

Note:

- Causes new bug in #827

This change introduces a [bug shown in this fiddle](https://jsfiddle.net/sb9zn0oc/). Since this is an unusual usage of `$children` we have decided to move forward with this PR. However, there may be other edge cases. If there are many more edge cases, then we will remove sync functionality from Vue Test Utils.


38elements:
It may be necessary to revert #675.

eddyerburgh:
For the moment we will continue to run `sync` compat mode (for Vue < 2.5.18), so that change is still 
required :)

38elements:
I think it is necessary to change `.circleci/config.yml`.

packages/test-utils/src/mount.js
```diff
@@ -13,22 +13,53 @@ import config from './config'
...
+  if (syncOption === false) {
+    Vue.config.async = true
+    return false
+  }
+  if (!semVerGreaterThan(Vue.version, '2.5.17')) {
+    warn(
+      `Vue Test Utils runs in sync mode by default. Due to bugs, sync mode ` +
+      `requires Vue > 2.5.18. In Vue Test Utils 1.0 sync mode will only be ` +
+      `supported with Vue 2.5.18 running in development mode. If you are ` +
+      `unable to upgrade, you should rewrite your tests to run synchronously` +
```
ghost: I'm a little confused here: should this read

```suggestion
      `unable to upgrade, you should rewrite your tests to run asynchronously` +
```

?

eddyerburgh: Yes, thanks Winnie 👍

ryanlitalien:
Following along, is there anything else that needs to be added for this to get merged? Thanks!

agualis:
💪🤗

Issue #1067, opened 2018-12-13
Components manually imported via $options.components do not exist when mounted()


reginsmol:
### Version
1.0.0-beta.27

### Reproduction link
[https://github.com/Reginsmol/test-utils-components-option-reproduction](https://github.com/MattRynbrand/test-utils-components-option-reproduction)

### Steps to reproduce
- npm install
- npm run dev
- npm run unit

### What is expected?
It should print out the ChildComponents in the components option.

### What is actually happening?
it only contains "transition" and "transition-group" when doing a regular mount(), however using shallowMount() the child components (ChildComponent, ChildComponent2) exist in the object.

reginsmol:
Is there any update on this?

PR #1084, opened 2018-12-29
fix: no unnecessary extend

eddyerburgh:
Currently create a subclass of a component with `extend` two times, in order to apply instance options. This causes some component options to not have the same own props as they would in production (e.g. #1067). 

One solution is to stop calling `extend` twice. The solution in this PR means that you can no longer override component options like `methods`, or `computed`, using instance options. 

The instance options would instead be used to instantiate the root Vue instance, so `store`, `route`, can still be passed. 

This is a large breaking change:

- Remove `methods` option
- Removes ability to overwrite `methods`, `computed`, or other component properties with instance options

This PR is a proof of concept as I investigate alternative approaches to fixing #1067.


docs/api/options.md
```diff
@@ -292,29 +292,21 @@ When `sync` is `false`, the Vue component is rendered asynchronously.
 
 ## Other options
 
-When the options for `mount` and `shallowMount` contain the options other than the mounting options, the component options are overwritten with those using [extends](https://vuejs.org/v2/api/#extends).
+When the options for `mount` and `shallowMount` contain the options other than the mounting options, the options are added used when the Vue instance is initialized.
```
grusingh, line 295: When the options for `mount` and `shallowMount` contain the options other than the mounting options, the options are added ~used~ when the Vue instance is initialized.

docs/api/options.md
```diff
@@ -292,29 +292,21 @@ When `sync` is `false`, the Vue component is rendered asynchronously.
...
-  template: '<div>{{ foo() }}{{ bar() }}{{ baz() }}</div>',
-  methods: {
-    foo () {
-      return 'a'
-    },
-    bar () {
-      return 'b'
-    }
-  }
+  template: `
```
grusingh, line 299: This example could be improved to make context more clear. 
`expect(wrapper.text()).toBe('aBC')` would fail?

packages/create-instance/create-instance.js
```diff
@@ -90,8 +90,8 @@ export default function createInstance (
   // extend component from _Vue to add properties and mixins
   // extend does not work correctly for sub class components in Vue < 2.2
   const Constructor = typeof component === 'function'
-    ? _Vue.extend(component.options).extend(instanceOptions)
-    : _Vue.extend(component).extend(instanceOptions)
+    ? _Vue.extend(component.options)
```
grusingh, line 93: :+1:

grusingh (commented):
Removing calls to `extend` twice seems good. :)
Review summary: There seems a typo in documentation and one of code example could be improved.

JessicaSachs:
We're closing this experimental PR for now. The issue is still outstanding, but the API change and approach would break too many tests for our users.

We will revisit approaches to solving the initial issue as time goes on.

PR #1090, opened 2019-01-08
Fix the file names referenced in comments

nlitwin:
The test file example references the following import:
```js
import storeConfig from './store-config'
```
but the code example that is imported calls itself `store-config.spec.js` instead of `store-config.js`.

- [x] Fix the file reference in the comment of the JS file
- [x] Add a file reference to the spec file


eddyerburgh (approved):
Thanks :)

PR #1097, opened 2019-01-14
docs(mocha-webpack): replace mocha-webpack by mochapack fork

alexbonhomme:
Hey!

This PR fix current documentation about testing with webpack + mocha. 

As `mocha-webpack` seems not maintained anymore the current guide doesn't works anymore. A fork named `mochapack` fix the issue, so I fixed the doc accordingly.


eddyerburgh:
Thanks for this.

Although mocha-webpack not being maintained is an issue, https://github.com/sysgears/mochapack only includes the initial commit. We won't change the recommendation until there's a proven alternative

jblotus:
I arrived here after trying to follow the guide and use mocha-webpack. It looks like mochapack has been updated since this ticket.

jblotus:
I can report that I was able to run vue-test-utils successfully with webpack4 and mochapack.

nothingismagick:
The staleness of mocha-webpack was the reason why we didn't include it in @quasar/testing. Will look into this new repo and report back.

despreston:
+1 for moving to mochapack. I went down the same route of broken tests and finally discovering mochapack.

eddyerburgh:
OK, it's been a few months and mochpack is being actively maintained.

Can you resolve the conflicts in this PR @alexbonhomme?

alexbonhomme:
Sure, I'll have a look on it quickly

afontcu:
Thank you for handling this! 🎉

PR #1106, opened 2019-01-20
fix: handle errors in destroy

### LINKED ISSUE #1105: Destroy swallows errors in hooks
### Version
1.0.0-beta.28

### Reproduction link
[https://codesandbox.io/s/w75w7ko968](https://codesandbox.io/s/w75w7ko968)

### Steps to reproduce
1. have an error in a destroy hook
2. test this

### What is expected?
The hook stops at the error and this gets thrown in the test

The hook stops at the error and the test does _not_ know about the error

---
This is likely because Vue catches an error like this, and would show the red screen, but this is still something I'd like to catch in a test.

Likely we need to catch and rethrow the error in below code

https://github.com/vuejs/vue-test-utils/blob/c13cac3856b3202cc40d9a6ef975cf9655d713ce/packages/test-utils/src/wrapper.js#L145-L155

eddyerburgh:
- Throw if error thrown when instance destroyed

Fixes #1105


Haroenv:
Cool, thanks!

PR #1123, opened 2019-01-31
chore: add `repository.directory` field to each `package.json`

damianstasik:
Like in a few monorepos in `vuejs` GH org, I wanted to introduce the same changes here related to the new `package.json` field that lets you define a path to each package that will be available on their pages on `npmjs.com`. Here is the RFC with a few more details:
https://github.com/npm/rfcs/blob/latest/accepted/0010-monorepo-subdirectory-declaration.md

### DIFF
--- a/packages/server-test-utils/package.json
+++ b/packages/server-test-utils/package.json
@@ -14,7 +14,8 @@
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


eddyerburgh (commented):
Thank you :)

PR #1124, opened 2019-02-01
docs: Fix link to Vuex document

numb86:
Firstly, since the link of Japanese translation was wrong, I fixed it.
Next, since the redirect occurred in all language, the link destination was corrected.


eddyerburgh (approved):
Thank you :)

PR #1132, opened 2019-02-06
docs: correct explanation about nextTick in testing-async-components.md

HusamElbashir:
`$nextTick` always uses microtasks now in Vue 2.6 just like it did pre v2.5.0. In v2.5 macrotasks are used to handle some edge cases.

**References**:

1. [2.6 Internal Change: Reverting nextTick to Always Use Microtask](https://gist.github.com/yyx990803/d1a0eaac052654f93a1ccaab072076dd)

2. [fix: always use microtasks for nextTick](https://github.com/vuejs/vue/commit/850555d1faa9be7d8306adffd95c7dee5e58717f)

### DIFF
--- a/docs/guides/testing-async-components.md
+++ b/docs/guides/testing-async-components.md
@@ -53,7 +53,7 @@ it('fetches async when a button is clicked', () => {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


eddyerburgh (approved):
Thanks 👍

PR #1143, opened 2019-02-10
server-test-utils: render() returns Cheerio not string

### LINKED ISSUE #1131: server-test-utils: render() returns Cheerio not string (d.ts)
### Version
1.0.0-beta.29

### Reproduction link
[https://codesandbox.io/s/4xvk5l7mo9](https://codesandbox.io/s/4xvk5l7mo9)

### Steps to reproduce
[index.d.ts](https://github.com/vuejs/vue-test-utils/blob/v1.0.0-beta.29/packages/server-test-utils/types/index.d.ts#L58) declares `render()` as returning `string` while the [documentation](https://vue-test-utils.vuejs.org/api/render.html#render) states that it is indeed "CheerioWrapper" (`Cheerio`) - the latter is also what can be observed when inspecting the result.

(f3dfb1e81f43c05d8ae871dba9184d83a3b3548e, for #571, may have been an impediment for this in the past but is long fixed.)

### What is expected?
`render()` type definition declares `Cheerio` return value.

`render()` type definition declares `string` return value.

---
We created [a hack around this](https://gerrit.wikimedia.org/r/c/wikibase/termbox/+/487887/1/tests/unit/components/Sectionedit.spec.ts#14) using [@types/cheerio](https://www.npmjs.com/package/@types/cheerio).

wiese:
The documentation (https://vue-test-utils.vuejs.org/api/render.html#render)
as well as the implementation state render()'s return type as `Cheerio`,
the type definition should consequently do the same.

~Please take a good look at the dist files - there is quite a bit of iffy change, possibly due to different versions of tools being used. I could not find infrastructure instructions (e.g. `Dockerfile`) however. Please advise / rebuild.~

Introduced during #618
Fixes #1131

### DIFF
--- a/packages/server-test-utils/package.json
+++ b/packages/server-test-utils/package.json
@@ -24,7 +24,8 @@
   },
   "homepage": "https://github.com/vuejs/vue-test-utils#readme",
   "dependencies": {
-    "cheerio": "^1.0.0-rc.2"
+    "cheerio": "^1.0.0-rc.2",
+    "@types/cheerio": "^0.22.10"
   },
   "devDependencies": {
     "chalk": "^2.1.0",
--- a/packages/server-test-utils/types/index.d.ts
+++ b/packages/server-test-utils/types/index.d.ts
@@ -54,9 +54,9 @@ interface VueTestUtilsConfigOptions {
 
 export declare let config: VueTestUtilsConfigOptions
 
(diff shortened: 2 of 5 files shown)


eddyerburgh:
Thanks, can you reset the dist files to master (we only update them when we release a new version)

wiese:
Thanks, @eddyerburgh. Did so.

eddyerburgh (approved):
Thanks :)

PR #1157, opened 2019-02-17
Update testing-single-file-components-with-jest.md

StephenChips:
**TYPO FOUND**。the reg to match js file should be `^.+\.js$`, not `^.+\\.js$`.


eddyerburgh:
The double backslash is intentional—in JSON it's a special sequence that means Backslash character: https://tools.ietf.org/html/rfc7159#section-7

PR #1160, opened 2019-02-21
chore: switch to official image

gengjiawen:


eddyerburgh:
Thanks for the PR but I'm not sure we need this, it installs dependencies that we don't use. What's the benefit of us using it?

gengjiawen:
`vuejs/ci` image is from https://github.com/vuejs/vue-ci-image, I don't quite sure which part is redundant for this project.

Also, circle ci image is 707M, while the official one is 460M.

* https://microbadger.com/images/circleci/node:6.13.0-stretch-browsers
* https://microbadger.com/images/vuejs/ci

eddyerburgh (approved):
Ok, I didn't realise the circleci/node image was so large.

Thanks for this :)

PR #1165, opened 2019-03-02
docs: update key event example

### LINKED ISSUE #1162: Can't set `which` when triggering `keydown` event
### Version
1.0.0-beta.29

### Reproduction link
[https://codesandbox.io/s/8k2z89m41j?fontsize=14](https://codesandbox.io/s/8k2z89m41j?fontsize=14)

### Steps to reproduce
1. Open CodeSandbox URL.
2. Open Tests tab.

### What is expected?
The tests should pass. (The component and tests are copied from https://vue-test-utils.vuejs.org/guides/#testing-key-mouse-and-other-dom-events.)

An error: "Cannot assign to read only property 'which' of object '#<KeyboardEvent>'"

---
I ran into this bug while trying to follow the guide to test keyboard events in my component.

eostrom:
`event.which` is deprecated and (at least in recent Chrome) can't be set, which made the example test fail. `event.key` works.

Closes #1162.

### DIFF
--- a/docs/guides/dom-events.md
+++ b/docs/guides/dom-events.md
@@ -101,7 +101,6 @@ This component allows to increment/decrement the quantity using various keys.
   const KEY_DOWN = 40
   const KEY_UP = 38
   const ESCAPE = 27
-  const CHAR_A = 65
 
   export default {
     data() {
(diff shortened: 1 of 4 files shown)


eddyerburgh (approved):
Thank you

PR #1166, opened 2019-03-03
fix: stub model option

### LINKED ISSUE #965: Invalid behavior of `shallowMount` on testing a child component that uses custom `v-model`
### Version
1.0.0-beta.25

### Reproduction link
[https://github.com/chenxeed/reproduce-bug-vue-test-utils](https://github.com/chenxeed/reproduce-bug-vue-test-utils)

### Steps to reproduce
(Best to clone and try this github repo to reproduce the bugs on running the unit test)
https://github.com/chenxeed/reproduce-bug-vue-test-utils

To reproduce:
- Create a child component that uses custom `v-model`
- Create a parent component that uses the child component and use the `v-model` to pass the props. Make the props as "required"
- Test the parent component with `shallowMount` to test if the data passed to the child component through `v-model` is working or not

### What is expected?
The test should able to detect the passed data from the parent component to the child component with custom `v-model`. This is working in `mount`, but not `shallowMount`.

The value passed from parent component to child component is not detected, and it shows warning "[Vue warn]: Missing required prop" instead.

---
I need to be able to test this scenario with `shallowMount`, because I'm testing a Page component that has a lot of child components and I don't want to render all by using `mount`, since I just want to test the Page components logic and its interface with the child components.

eddyerburgh:
- Stub component `model` option

Fixes https://github.com/vuejs/vue-test-utils/issues/965

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -38,6 +38,7 @@ function getCoreProperties(componentOptions: Component): Object {
   return {
     attrs: componentOptions.attrs,
     name: componentOptions.name,
+    model: componentOptions.model,
     props: componentOptions.props,
     on: componentOptions.on,
     key: componentOptions.key,
--- a/test/specs/shallow-mount.spec.js
+++ b/test/specs/shallow-mount.spec.js
@@ -554,4 +554,26 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'shallowMount', () => {
       )
     }
   )
+
+  it('stubs model', () => {
+    const ChildComponent = {
+      template: '<div />',
+      model: {
+        prop: 'a',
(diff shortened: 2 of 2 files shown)


themeler:
Any possible release date? This warnings are really messing up my console during tests :(

ffxsam:
@themeler @eddyerburgh This is still not published, it seems. Would be great to see this at least released as a hotfix if possible?

rnicholus:
I'm also wondering when this will be published. Is there anything we can do to help?

adrichey:
Any chance this is getting released soon? This has been closed for a while and still hasn't seen a release.

rnicholus:
@adrichey see #1329

eddyerburgh:
I've release beta.30

PR #1193, opened 2019-03-24
feat: allow setData with dot strings as nested paths

### LINKED ISSUE #1184: Can wrapper.setData be used to set data which is a nested object without setting entire object?
As title.

P.S. What's the diff between setting a data directly (like `wrapper.vm.$data.foo = 'foo'`) versus using wrapper.setData (like `wrapper.setData({foo: 'foo'}`) ? Is it just is a shortcut?

If it doesn't, I suggest we can provide a new function that accept a object path selector instead of a simple data name, for instance:

```
// setting wrapper.vm.$data.time.date to 'now'
wrapper.setData({ 'time.date': 'now' })
```

briwa:
This closes #1184.

There is already a way to update nested data with `.setData`, albeit more verbose.
```javascript
console.log(wrapper.vm.$data)
// { foo: { bar: 'baz' } }

wrapper.setData({ foo: {  bar: 'baq' } })
// { foo: { bar: 'baq' } }  
```  

This PR proposes a less verbose way to do it:
```javascript
wrapper.setData({ 'foo.bar': 'baq2' } })
// { foo: { bar: 'baq2' } }
```
This resembles how nested data is being referenced when `watch`-ing values.

Both ways should still work fine, since passing in object as data is not just to update nested data, but to also update multiple data values at once.

What do you guys think? Thank you.

### DIFF
--- a/docs/api/wrapper/props.md
+++ b/docs/api/wrapper/props.md
@@ -2,7 +2,7 @@
 
 Return `Wrapper` `vm` props object. If `key` is provided, the value for the `key` will be returned.
 
-**Note the Wrapper must contain a Vue instance.**
+**Note: the Wrapper must contain a Vue instance.**
 
 - **Arguments:**
 
--- a/docs/api/wrapper/setData.md
+++ b/docs/api/wrapper/setData.md
@@ -4,7 +4,9 @@ Sets `Wrapper` `vm` data.
 
 setData works by recursively calling Vue.set.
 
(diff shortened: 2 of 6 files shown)


afontcu:
Hi @briwa! Thank you for coming up with this PR. Even though we see how this could be useful sometimes, we've decided not to  merge it to the library.

Again, thank you for your time and your willingness to contribute. Please do keep on making Vue Test Utils better :open_hands:!

PR #1203, opened 2019-04-05
Vue Test Utils trigger click with modifier keys

dnutiu:
Add an example for triggering a click event with event modifiers key


eddyerburgh (approved):
Thank you :)

PR #1229, opened 2019-05-09
Issue(1147): Add option to pretty print html components

### LINKED ISSUE #1147: Pretty-print html() output by default
My team is using `@vue/test-utils` for snapshot testing components but the snapshots are hard to parse because the HTML string output from `.html()` is not formatted in a readable way. If `html()` pretty-printed the HTML it would be a lot easier to review changes to the component-generated HTML.

`html()` would pretty-print the returned HTML string by default, perhaps using something like https://github.com/prettydiff/prettydiff. If necessary, a user could bypass pretty-printing by passing an option to `.html()`, perhaps `.html(false)` or `.html({ prettyPrint: false })`.

addisonssense:
This PR is to implement pretty printing for html components. 

Fixes #1147

### DIFF
--- a/flow/modules.flow.js
+++ b/flow/modules.flow.js
@@ -27,3 +27,7 @@ declare module 'cheerio' {
 declare module 'semver' {
   declare module.exports: any
 }
+
+declare module 'pretty' {
+  declare module.exports: any
+}
--- a/flow/wrapper.flow.js
+++ b/flow/wrapper.flow.js
@@ -5,7 +5,6 @@ import type WrapperArray from '~src/WrapperArray'
 
 declare type Selector = any
 declare type Components = { [name: string]: Component }
-declare type HtmlOptions = { prettyPrint: boolean }
 
 declare interface BaseWrapper {
   // eslint-disable-line no-undef
--- a/flow/wrapper.flow.js
+++ b/flow/wrapper.flow.js
@@ -5,6 +5,7 @@ import type WrapperArray from '~src/WrapperArray'
 
 declare type Selector = any
 declare type Components = { [name: string]: Component }
+declare type HtmlOptions = { prettyPrint: boolean }
 
 declare interface BaseWrapper {
   // eslint-disable-line no-undef
--- a/packages/test-utils/package.json
+++ b/packages/test-utils/package.json
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "dom-event-types": "^1.0.0",
-    "lodash": "^4.17.4"
+    "lodash": "^4.17.4",
+    "pretty": "^2.0.0"
   }
 }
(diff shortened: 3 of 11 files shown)


quinnlangille (approved):
Yes! I've been waiting for this 🥇

eddyerburgh:
Thank you for the PR, as per my comment in the issue: can you make pretty-printing the default behavior?

addisonssense:
@eddyerburgh done! I have updated the existing test to use the `prettyPrint: false` option. If its better to update the html being asserted instead of using the flag, I can definitely do that instead. I figured this method was less intrusive

bilyes (approved):
Very cool feature at a relatively low cost. Thanks for the initiative @addisonssense !

briwa:
@addisonssense I think existing tests should be updated with `prettyPrint: true` instead of `false`, since it should reflect the default behavior of `.html`? It will be consistent for future tests that use `.html`, since they would probably be written with the default behavior in mind (pretty-printed).

Just my $0.02 ✌️ But anyhow, nice PR!

addisonssense:
@briwa I think you make a good point. I'll do that 👍

TheJaredWilcurt:
A few months ago I proposed a different approach to this. `jest-serializer-vue` has options built in, but the options do not format (or "pretty print") the code very well.

* https://github.com/eddyerburgh/jest-serializer-vue/issues/13

### TLDR:

* Change the default settings being passed in to `pretty` to these:
  ```js
  const options = {
    indent_char: ' ',
    indent_inner_html: true,
    indent_size: 2,
    inline: [],
    sep: '\n',
    unformatted: ['code', 'pre']
  };
  ```
* Allow users to pass in their own options object. (Not sure where they would do this, `package.json`?)
* Have instructions of how to pass in the object with the previous settings (so users can opt out of breaking changes).

This allows us to change the defaults again in the future without worrying about breaking people's tests, as the defaults can be documented for each version.

eddyerburgh:
I intend to implement the changes in jest-serializer-vue but I haven't had time to work on it.

jest-serializer-vue will be formatted more heavily, whereas vue-test-utils can just use the defaults.

@addisonssense thanks for updating, but I wasn't clear in my previous comment. I don't think there should be an option to pretty-print or not. I don't think it should be configurable, and pretty printing should always be the default behavior. 

Can you update the PR to remove the `prettyPrint` option entirely?

addisonssense:
@eddyerburgh done! Pretty printing is now the default. The option to disable it is removed.

addisonssense:
@eddyerburgh it looks like my build is stuck on circleci. Would you mind restarting that step?

eddyerburgh (approved):
Thanks :)

PR #1230, opened 2019-05-10
polyfill Element.matches (fixes #1223)

### LINKED ISSUE #1223: Wrapper.find() does not work in IE11 without polyfill
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/eddyerburgh/vue-test-utils-karma-example](https://github.com/eddyerburgh/vue-test-utils-karma-example)

### Steps to reproduce
Using Karma test runner in IE11 environment, run the Counter.spec.js test.

### What is expected?
`wrapper.find('button')` should return a wrapper of the matching element.

No match is returned and an error is thrown that `trigger` must be provided a non-empty wrapper.

---
This is due to IE11 not supporting `Element.matches()`. Intead IE11 has the non-standard `Element.msMatchesSelector()`. MDN has a [simple polyfill](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Polyfill) that should be implemented and solves this problem:
```js
if (!Element.prototype.matches) {
  Element.prototype.matches = Element.prototype.msMatchesSelector || Element.prototype.webkitMatchesSelector;
}
```

I'm happy to open a PR to fix this, I just don't know where the correct place to install this polyfill would be.

markbrouch:
This fixes a bug (#1223) in IE and older versions of Chrome that use a non-standard name for `Element.matches()` by implementing a [polyfill](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Polyfill).

Please advise if there is a more appropriate place to implement this polyfill.

### DIFF
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -5,14 +5,6 @@ import {
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
 
-// Polyfill `Element.matches()` for IE and older versions of Chrome:
-// https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Polyfill
-if (!Element.prototype.matches) {
-  Element.prototype.matches =
-    Element.prototype.msMatchesSelector ||
-    Element.prototype.webkitMatchesSelector
-}
-
 export function vmMatchesName(vm, name) {
   return (
     !!name && (vm.name === name || (vm.$options && vm.$options.name === name))
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -5,6 +5,14 @@ import {
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
 
+// Polyfill `Element.matches()` for IE and older versions of Chrome:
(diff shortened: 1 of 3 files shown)


markbrouch:
@eddyerburgh it looks like jsdom globals are turned off in node env: https://github.com/vuejs/vue-test-utils/blob/dev/test/setup/mocha.setup.js#L3

This causes tests to fail on this change because `Element` does not exist outside of jsdom. What's the intent of disabling jsdom globals for node env?

eddyerburgh:
The tests run for both @vue/test-utils and @vue/server-test-utils. 

Can you add the polyfill in `mount.js`, which is only used in @vue/test-utils.

markbrouch:
> The tests run for both @vue/test-utils and @vue/server-test-utils.
> 
> Can you add the polyfill in `mount.js`, which is only used in @vue/test-utils.

I moved the polyfill to its own module that is invoked in `mount.js`.

eddyerburgh (approved):
Thank you :)

asinakdeniz:
Hi @eddyerburgh, 
As I understood, this fix has not been released yet, when are you planning to get release? Thanks...

PR #1231, opened 2019-05-10
vue-test-utils: allow `false` as a component stub value

garyo:
Updated types/index.d.ts to allow false to be passed as a component stub value
in mount or shallowMount. `false` in this case means don't use a stub for the
given component.

### DIFF
--- a/packages/test-utils/types/index.d.ts
+++ b/packages/test-utils/types/index.d.ts
@@ -22,7 +22,7 @@ type Slots = {
  * If it is an array of string, the specified children are replaced by blank components
  */
 type Stubs = {
-  [key: string]: Component | string | true
+  [key: string]: Component | string | boolean
 } | string[]
 
 /**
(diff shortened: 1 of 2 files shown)


packages/test-utils/types/index.d.ts
```diff
@@ -22,7 +22,7 @@ type Slots = {
  * If it is an array of string, the specified children are replaced by blank components
  */
 type Stubs = {
-  [key: string]: Component | string | true
+  [key: string]: Component | string | true | false
```
eddyerburgh:
```suggestion
  [key: string]: Component | string | boolean
```

eddyerburgh:
Can you also update the type file in packages/server-test-utils/types/index.d.ts?

garyo:
done.

eddyerburgh:
Hi @garyo, there's another types file for the `@vue/server-test-utils` package that's in packages/server-test-utils/types/index.d.ts that also needs to be updated. Can you make the changes in that file too?

garyo:
Sorry, I had that fix queued up, just forgot to add it! Here you go. This time for sure. :-)

eddyerburgh (approved):

eddyerburgh:
Thanks :)

Seminioni:
@eddyerburgh when this fix will be available in public release?

eddyerburgh:
By September 15 latest

Issue #1209, opened 2019-04-16
Different snapshot in coverage mode


adi-zz:
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/adi-zz/istanbul-snapshot-in-coverage-mode](https://github.com/adi-zz/istanbul-snapshot-in-coverage-mode)

### Steps to reproduce
npm i

npm run test

### What is expected?
`npm run test` generates 2 snapshots, for the same component. The 2 generated snapshots are logged in the console when running `npm run test`. The 2 generated snapshots should be identical.

### What is actually happening?
 When running jest in 'normal' mode, the generated snapshot is `<div><h1>parent</h1> <child-stub items="1,string,() => true"></child-stub></div>` (as expected).

When running in jest 'coverage' mode, the generated snapshot is `<div><h1>parent</h1> <child-stub items="1,string,() => {
              /* istanbul ignore next */
              cov_13kzki4jl6.f[2]  ;
              cov_13kzki4jl6.s[7]  ;
              return true;
            }"></child-stub></div>`. For this reason, running jest in coverage mode will fail, because the snapshot is different.

---
Maybe the issue is related to https://github.com/vuejs/vue-test-utils/issues/975.

eddyerburgh:
This issue will be fixed when #975 is implemented

KaiFuglem:
Is this fixed now that https://github.com/vuejs/vue-test-utils/pull/1834 has been merged?

PR #1237, opened 2019-05-20
fix(stubs): function in props should be rendered as a deterministic string

### LINKED ISSUE #975: Function properties in stubs should not render source code
### Version
1.0.0-beta.25

### Reproduction link
[https://codesandbox.io/s/135my5kk9j](https://codesandbox.io/s/135my5kk9j)

### Steps to reproduce
1. Go to the Tests tabs and run the test
2. Check the Console output in the Console tab

### What is expected?
wrapper.html() should return something like
```
<childcomponent-stub clickhandler="[Function]"></childcomponent-stub> 
```

```
<childcomponent-stub clickhandler="function bound handleClick() {
    [native code]
}"></childcomponent-stub> 
```

---
I ran into the bug because the snapshot tests failed on our CI environment, even though I had properly updates the snapshots locally. The failing spot was a Function property which was rendered differently on local and CI environments (due to Istanbuls coverage comments being added).

briwa:
This closes #975 and #1209.

I figured that the root cause was the stubs rendering function as it is, while it should just have been a string placeholder. My idea is to override all functions in props to a placeholder, and since it's stubbed, the actual function implementation doesn't really matter (CMIIW).

Let me know if this is good enough, or if you have any feedback. Thanks.

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -142,7 +142,7 @@ function validateStub(stub) {
 }
 
 function shapeStubProps(props) {
-  const shapedProps: Record<string, any> = {}
+  const shapedProps: Object = {}
   for (const propName in props) {
     if (!props.hasOwnProperty(propName)) {
       continue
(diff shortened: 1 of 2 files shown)


briwa:
Closing this for now since it seems like I'm breaking some tests in previous versions... Will debug later.

fallemand:
Hi @briwa! Did you have the chance to have a look? It would be great to have this working.

stinovlas:
Hello! Is this PR moving forward? It's quite annoying to create workarounds for testing snapshots with functions. The proposed solution looks good to me.

PR #1238, opened 2019-05-20
fix(stubs): allows stubbing by component name as well

briwa:
There are times when we assign a component as a child with a different name than the registered name. Consider this example:

```typescript
// timeline/index.ts
@Component
export default Timeline extends Vue {
...
}
```

```typescript
// parent/index.ts
import TimelineComponent from './timeline/index.vue'
import TimelineClass from './timeline/index'

@Component({
  components: { TimelineComponent }
})
export default Parent extends Vue {
  $refs: {
    timeline: TimelineClass
  }
  ...
}
```

```typescript
// parent/__tests__/index.ts
...

const wrapper = mount<ParentClass>(ParentComponent, {
  stubs: ['timeline']
})

expect(wrapper.find(TimelineComponent).exists()).toBe(true) // This would fail

const wrapper2 = mount<ParentClass>(ParentComponent, {
  stubs: ['TimelineComponent']
})

expect(wrapper2.find(TimelineComponent).exists()).toBe(true) // This would pass
```

I found out that resolving components for stubs is using the keys of `$options.components` instead of the component name, as seen [here](https://github.com/vuejs/vue-test-utils/blob/v1.0.0-beta.29/packages/create-instance/create-component-stubs.js#L138). In my opinion, it should be able to be resolved using the component name as well. See the test for details.

Since I'm not looking for a breaking change, I modified the code such that it can now be resolved by both the component name and the keys of `$options.components`. Let me know what you think.

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -146,13 +146,9 @@ export function createStubsFromStubsObject(
 ): Components {
   const originalComponentsWithName = {}
   for (const childKey in originalComponents) {
-    if (!originalComponents.hasOwnProperty(childKey)) {
-      continue
-    }
-
     const component = originalComponents[childKey]
-    // Have the name as an alias so it can be both the name
-    // and the `components` key
+    // Have the name as an alias so it can be referred to
+    // by both the name and the `components` key
     if (component.name !== childKey) {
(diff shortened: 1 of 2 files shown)


briwa:
Closing this for now since some unrelated tests are breaking...

PR #1240, opened 2019-05-22
Introduce autoDestroy config option

### LINKED ISSUE #1236: Support cleaning up wrappers automagically

Currently it is necessary to call `wrapper.destroy()` / `wrapper.vm.$destroy()` manually. Forgetting to do so may result in difficult to track down side effects and performance degradation (especially when not mocking Vuex).

Introduce a new `enableAutoDestroy()`  helper which accepts a hook function such as [`afterEach()`](https://jestjs.io/docs/en/api#aftereachfn-timeout) to call `wrapper.destroy()` as callback.

winniehell:
This introduces a new `autoDestroy` config option which allows to automatically call `wrapper.destroy()` when creating a new wrapper instance or by passing in a custom hook function (such as `afterEach`).

closes https://github.com/vuejs/vue-test-utils/issues/1236

### DIFF
--- a/packages/test-utils/src/config.js
+++ b/packages/test-utils/src/config.js
@@ -3,5 +3,6 @@ export default {
   mocks: {},
   methods: {},
   provide: {},
-  silent: true
+  silent: true,
+  autoDestroy: false
 }
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -15,6 +15,13 @@ import { matches } from './matches'
 import createDOMEvent from './create-dom-event'
 import { throwIfInstancesThrew } from './error'
 
+const wrapperInstances = []
+
+const destroyAllInstances = () => {
+  wrapperInstances.forEach(wrapper => wrapper.destroy())
+  wrapperInstances.length = 0
+}
+
 export default class Wrapper implements BaseWrapper {
   +vnode: VNode | null
   +vm: Component | void
@@ -66,6 +73,17 @@ export default class Wrapper implements BaseWrapper {
       (this.vnode[FUNCTIONAL_OPTIONS] || this.vnode.functionalContext)
     ) {
       this.isFunctionalComponent = true
+    }
+
+    const { autoDestroy } = config
+    if (autoDestroy) {
+      if (autoDestroy instanceof Function) {
+        autoDestroy(destroyAllInstances)
+      } else {
+        destroyAllInstances()
+      }
+
+      wrapperInstances.push(this)
     }
   }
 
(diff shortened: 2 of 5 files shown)


packages/test-utils/src/config.js
```diff
@@ -3,5 +3,6 @@ export default {
   mocks: {},
   methods: {},
   provide: {},
-  silent: true
+  silent: true,
+  autoDestroy: false
```
winniehell, line 7: `false` should be the default because that is the current behavior

packages/test-utils/src/wrapper.js
```diff
@@ -15,6 +15,12 @@ import { matches } from './matches'
 import createDOMEvent from './create-dom-event'
 import { throwIfInstancesThrew } from './error'
 
+const wrapperInstances = []
```
winniehell, line 18: This needs to be an array because when passing a hook function it can happen that multiple wrapper instances were created.

packages/test-utils/src/wrapper.js
```diff
@@ -67,6 +73,17 @@ export default class Wrapper implements BaseWrapper {
     ) {
       this.isFunctionalComponent = true
     }
+
+    const { autoDestroy } = config
```
winniehell, line 78: do we want to allow overriding this via wrapper options?

winniehell:
@eddyerburgh Can you please review?

souldzin:
@winniehell, there's an interesting side-effect to this approach. Since this is coupled to the constructor, the last component created will not be destroyed (or the `autoDestroy` callback will not called).

This might not be a problem, but it's a quirk that's probably worth highlighting in the docs. 🤷‍♀

winniehell:
> but it's a quirk that's probably worth highlighting in the docs

@souldzin Absolutely! Thank you for reminding me! :bowing_man: 

EDIT: added a note now

packages/test-utils/src/wrapper.js
```diff
@@ -67,6 +74,17 @@ export default class Wrapper implements BaseWrapper {
...
+    if (autoDestroy) {
+      if (autoDestroy instanceof Function) {
+        autoDestroy(destroyAllInstances)
```
winniehell, line 81: This line is wrong. It will for example result in `describe > it > shallowMount > new Wrapper > afterEach`. That way we have a lot of `afterEach` calls inside `it`s while we only want a single one **outside**.

winniehell:
I'm not sure how to fix https://github.com/vuejs/vue-test-utils/pull/1240#discussion_r286564435, so I'm going to close this. I have suggested an alternative approach in https://github.com/vuejs/vue-test-utils/issues/1236#issuecomment-494874313.

winniehell:
new attempt at https://github.com/vuejs/vue-test-utils/pull/1245

PR #1244, opened 2019-05-26
feat(wrapper array): Allow negative indices to be passed to .at()

paulgv:
Passing a negative index to .at() lets you retrieve elements by starting from the end of the wrapper array (i.e. last item is at index -1).

```js
import Foo from './Foo.vue'

const wrapper = shallowMount(Foo)
const divArray = wrapper.findAll('div')

const lastDiv = divArray.at(-1)
expect(lastDiv.is('div')).toBe(true)
```

### DIFF
--- a/packages/test-utils/src/wrapper-array.js
+++ b/packages/test-utils/src/wrapper-array.js
@@ -23,10 +23,13 @@ export default class WrapperArray implements BaseWrapper {
   }
 
   at(index: number): Wrapper | VueWrapper {
-    if (index > this.length - 1) {
-      throwError(`no item exists at ${index}`)
+    const normalizedIndex = index < 0 ? this.length + index : index
+    if (normalizedIndex > this.length - 1 || normalizedIndex < 0) {
+      let error = `no item exists at ${index}`
+      error += index < 0 ? ` (normalized to ${normalizedIndex})` : ''
+      throwError(error)
     }
-    return this.wrappers[index]
+    return this.wrappers[normalizedIndex]
   }
 
   attributes(): void {
--- a/docs/api/wrapper-array/at.md
+++ b/docs/api/wrapper-array/at.md
@@ -1,6 +1,7 @@
 ## at
 
 Returns `Wrapper` at `index` passed. Uses zero based numbering (i.e. first item is at index 0).
+If `index` is negative, indexing starts from the last element (i.e. last item is at index -1).
 
 - **Arguments:**
 
@@ -16,6 +17,10 @@ import Foo from './Foo.vue'
 
 const wrapper = shallowMount(Foo)
 const divArray = wrapper.findAll('div')
+
 const secondDiv = divArray.at(1)
-expect(secondDiv.is('p')).toBe(true)
+expect(secondDiv.is('div')).toBe(true)
+
+const lastDiv = divArray.at(-1)
+expect(lastDiv.is('div')).toBe(true)
 ```
--- a/test/specs/wrapper-array/at.spec.js
+++ b/test/specs/wrapper-array/at.spec.js
@@ -13,12 +13,40 @@ describeWithShallowAndMount('at', mountingMethod => {
     expect(p.classes()).to.contain('index-1')
   })
 
+  it('returns Wrapper at index from the end when index is negative', () => {
+    const TestComponent = {
+      template: '<div><p class="index-first" /><p class="index-last"/></div>'
+    }
+    const all = mountingMethod(TestComponent).findAll('p')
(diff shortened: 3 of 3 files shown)


packages/test-utils/src/wrapper-array.js
```diff
@@ -23,7 +23,8 @@ export default class WrapperArray implements BaseWrapper {
   }
 
   at(index: number): Wrapper | VueWrapper {
-    if (index > this.length - 1) {
+    index = index < 0 ? this.length + index : index
```
winniehell: should we

```suggestion
    const normalizedIndex = index < 0 ? this.length + index : index
```

Otherwise for an array of length 1, the error thrown for `wrapper.at(-2)` will be `no item exists at -1` which could cause confusion.

paulgv: Yeah I wasn't sure about that either, but at the same time, I think it's useful to see the computed index in the error message. So I changed the message when the index is negative, so we can see both the original argument and the computed one. What do you think @winniehell?

winniehell: looks good—thank you! :tada:

eddyerburgh:
Thanks for the PR. I haven't seen this in an API before, what are some examples of language features/ libraries that use -1 to mean the end of an array?

We could also add a `last` helper method. That would be easier to understand for me

paulgv:
@eddyerburgh The best example I can think of is Python, which has this behavior built-in for lists, eg:

```python
>>> list = [1, 2, 3]
>>> list[-1]
3
>>> list[-2]
2
>>> list[-3]
1
```

To me it would feel pretty natural to have this implemented in `vue-test-utils` (I actually have tried it multiple times only to be reminded it isn't supported).

A `last` helper would certainly be useful but not as flexible as supporting negative indices. A use case I can think of is for testing a pagination component where you might want to test the last three items, eg:

- `items.at(-1)` -> "Last page" button
- `items.at(-2)` -> "Next page" button
- `items.at(-3)` -> "Page #" button

winniehell:
> what are some examples of language features/ libraries that use -1 to mean the end of an array?

@eddyerburgh I hope you don't mind me jumping in here. :smiley: A similar API is used in

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring
- https://lodash.com/docs/4.17.11#nth

lmiller1990:
This feels a bit unintuitive to me. Since JS doesn't support this natively, I'm not sure it makes sense here, either.

winniehell:
@lmiller1990 In my comment above I have linked native JavaScript APIs that support negative indices.

If you think that this is not the expected behavior, what behavior would you expect when doing `wrapper.at(-1)`? should it return `undefined`?

lmiller1990:
I would expected undefined: `var a = [1,2,3,4,5]; a[-1] //=> undefined`.

I'm actually surprised to find there is some support for `-1` in JS at all - the more you know. I don't think this is a _bad_ idea - obviously it's convenient, especially for people coming from languages where this is supported. I doubt I'd expect this behaviour unless I had explicitly seen it in the docs.

If we want to follow `slice`, it would return an array? `[1,2,3,4,4].slice(-1) //=> [4]`.

`last` would make more senes to me.

paulgv:
> I would expected undefined: var a = [1,2,3,4,5]; a[-1] //=> undefined.

I agree that, when manipulating a native Array in Js, it feels counterintuitive to access a negative index like you would in Python, but doing that through an API (`.at(-1)`) feels fine to me. Especially given the fact that this is already possible with native Js APIs, like @winniehell pointed out.

> If we want to follow slice, it would return an array? [1,2,3,4,4].slice(-1) //=> [4].

The goal here is not to mimic `.slice()`, @winniehell was simply giving some examples where native Js APIs use negative indices. It wouldn't make any sense for `.at()` to return an array because it is designed to return a specific element from a wrappers list.

But I could imagine a `.slice()` method being implemented in **vue-test-utils**, I guess it could have some utility.

> last would make more senes to me.

Again, a `.last()` helper would certainly be useful but doesn't provide as much flexibility as supporting negative indices in `.at()` does. IMHO, if we were to implement a `.last()` helper, it should be a simple alias for `.at(-1)`.

lmiller1990:
Ok, I reckon we merge this up. Going to get few more reviews going.

dobromir-hristov (approved):

afontcu (approved):
LGTM!

Issue #1224, opened 2019-05-02
Give ability to customise Wrapper and/or WrapperArray


davidlandais:
### What problem does this feature solve?
I would like to propose you to give the ability to developers to register custom commands like wrapper.`findElementWithTestIdAttribute()`.isVisible() or wrapper.`findVueAndEmit(selector, ['event', ...args])`. This is my needs, that i work around with some helpers. This is for example, a feature shipped with Cypress. What do you think ? Does it sound a good idea ?

### What does the proposed API look like?
The API doesn't seems to propose this feature.

winniehell:
In https://github.com/vuejs/vue-test-utils/pull/1245#discussion_r289268663, @souldzin suggested using an `EventEmitter` to hook into `Wrapper` creation. A Vue instance as event hub (described [here](https://vuejs.org/v2/guide/migration.html#dispatch-and-broadcast-replaced)) would serve the same purpose.

We could then do the following:

```javascript
import { events } from '@vue/test-utils';

events.$on('wrapper-created', newWrapper => {
  Object.assign(newWrapper. {
    findElementWithTestIdAttribute() {
      // ...
    },

    isVisible() {
      // ...
    },

    findVueAndEmit() {
      // ...
    },
  });
});
```

or alternatively:

```javascript
import { Wrapper } from '@vue/test-utils';

Wrapper.events.$on('created', newWrapper => {
  // ...
});
```

@eddyerburgh What do you think?

winniehell:
@eddyerburgh If the above approach makes sense to you, I would be happy to come up with a pull request to unblock https://github.com/vuejs/vue-test-utils/issues/1236.

winniehell:
@JessicaSachs Can you please take a look at this?

JessicaSachs:
@winniehell for back and forth convo can you find me on Vue Land discord? It’s easier to chat there.

winniehell:
@JessicaSachs I didn't have the impression that there was much back and forth yet and also would like to have the result documented here. :sweat_smile: But sure, we can explore the different options on discord first. :+1:

JessicaSachs:
Agreed and agreed, but I want to talk it over with you and then we will
report back to GH so that other maintainers can weigh in.

JessicaSachs:
Implementing this feature request would make it easier to fix a bug (#1236) by way of an existing PR (#1245) that wants to depend on this new feature.

@winniehell and I will look into ways of solving #1236 without depending on this feature request.

[...]

Issue #1135, opened 2019-02-07
Always store selector in Wrapper created by wrapper.find(...)


ghost:
### What problem does this feature solve?

When using `wrapper.find(...)`, the returned new wrapper only knows the selector if nothing was found. This makes it difficult to debug tests with dynamic selectors and to create custom matchers like `expect(wrapper.find('.something')).toExist()`.

### What does the proposed API look like?

Make Wrapper always have a `selector` property when created with `.find(...)`—not only `ErrorWrapper` ( https://github.com/vuejs/vue-test-utils/blob/36adbf357fe2e6085d826e143b32d0fa37b6d938/packages/test-utils/src/error-wrapper.js#L9 ).

eddyerburgh:
This is a great idea :)

Would you like to implement it @gitlab-winnie ?

ghost:
> Would you like to implement it @gitlab-winnie ?

I can't make any Promise yet but have to await my workload for the next month. Definitely interested in getting my feet wet at the Vue test utils beach though. 🏖

winniehell:
I started working on this in https://github.com/vuejs/vue-test-utils/pull/1248.

PR #1245, opened 2019-05-27
feat: Introduce enableAutoDestroy() helper function

### LINKED ISSUE #1236: Support cleaning up wrappers automagically

winniehell:
This introduces a new `enableAutoDestroy()` helper function which allows to automatically call `wrapper.destroy()` by passing in a custom hook function (such as `afterEach`).

closes #1236


winniehell:
@souldzin can I ask you to take a look at this?

packages/test-utils/src/wrapper.js
```diff
@@ -582,3 +594,15 @@ export default class Wrapper implements BaseWrapper {
...
+
+export const enableAutoDestroy = (hook: Function) => {
+  if (Wrapper._instances) {
+    throwError('enableAutoDestroy cannot be called more than once')
+  }
+
+  Wrapper._instances = []
+
+  hook(() => {
+    ;(Wrapper._instances || []).forEach(wrapper => wrapper.destroy())
```
winniehell: should we throw an error here if `Wrapper._instances` is not defined?

packages/test-utils/src/wrapper.js
```diff
@@ -582,3 +594,15 @@ export default class Wrapper implements BaseWrapper {
     this.element.dispatchEvent(event)
   }
 }
+
+export const enableAutoDestroy = (hook: Function) => {
+  if (Wrapper._instances) {
+    throwError('enableAutoDestroy cannot be called more than once')
+  }
+
+  Wrapper._instances = []
```
winniehell: initializing this here feels a little weird. should instead `isAutoDestroyEnabled = true`?

packages/test-utils/src/wrapper.js
```diff
@@ -141,6 +146,13 @@ export default class Wrapper implements BaseWrapper {
       this.vm.$destroy()
       throwIfInstancesThrew(this.vm)
     }
+
+    // allow garbage collection
+    if (Wrapper._instances) {
+      Wrapper._instances = Wrapper._instances.filter(
```
winniehell: in combination with the hook callback this has `O(n²)` complexity (`n` being the wrapper instances). maybe we want to get rid of this.

winniehell (commented):
I have pushed the following changes:

- simplified the tests a little
- added a guard against calling the helper twice
- made the wrapper instances scoped inside Wrapper instead of the module to allow resetting them (for example in unit tests)

packages/test-utils/src/wrapper.js
```diff
@@ -582,3 +587,23 @@ export default class Wrapper implements BaseWrapper {
...
+  if (Wrapper._instances) {
+    throwError('enableAutoDestroy cannot be called more than once')
+  }
+
+  Wrapper._instances = []
+
+  hook(() => {
+    const { _instances } = Wrapper
+    if (!_instances) {
+      return
```
winniehell: should we throw an error here?

docs/api/enableAutoDestroy.md
```diff
@@ -0,0 +1,25 @@
+## enableAutoDestroy(hook)
+
+- **Arguments:**
+
+  - `{Function} hook`
+
+- **Usage:**
+
+`enableAutoDestroy` allows to destroy all created `Wrapper` instances using the passed hook function (for example [`afterEach`](https://jestjs.io/docs/en/api#aftereachfn-timeout)).
```
souldzin: **nitpick:** Wording is a little off here.

```suggestion
`enableAutoDestroy` will destroy all created `Wrapper` instances using the passed hook function (for example [`afterEach`](https://jestjs.io/docs/en/api#aftereachfn-timeout)).
```

packages/test-utils/src/wrapper.js
souldzin: I agree 🤔...

Just a thought, but I wonder if we should put all (or at least some) of this logic and tracking of instances in a separate `auto_destroy` module and just expose a way to hook into the constructor? Something like...

```javascript
// wrapper.js
// ref: https://nodejs.org/docs/latest-v10.x/api/events.html#events_class_eventemitter
import EventEmitter from 'events';

// Maybe this bit could even go in a separate module...
export const wrapperHooks = new EventEmitter();

export default class Wrapper  implements BaseWrapper {
  // ...
  constructor() {
    // ...
    wrapperHooks.emit('create', this);
  }
}
```

packages/test-utils/src/wrapper.js
souldzin: This would only happen if something manually overwrote `Wrapper._instances` to some falsey value? I think that warrants an error with a friendly (yet stern) message :wink:

packages/test-utils/src/wrapper.js
```diff
@@ -582,3 +587,23 @@ export default class Wrapper implements BaseWrapper {
...
+    }
+
+    while (_instances.length > 0) {
```
souldzin: **question:** Would a `forEach` and then setting `_instances = [];` be faster? Since these arrays aren't structured like stacks I'm not sure if popping is the best way to iterate 🤔...

packages/test-utils/src/wrapper.js
```diff
@@ -24,6 +24,7 @@ export default class Wrapper implements BaseWrapper {
   +options: WrapperOptions
   isFunctionalComponent: boolean
   rootNode: VNode | Element
+  static _instances: ?Array<Wrapper>
 
   constructor(
```
souldzin: **issue:** I think I just found a problem with doing this at the `Wrapper` ctor. Wrappers are created for all components found in the tree on `mount` or `shallowMount`. For instance, when I call `wrapper.find('...')`, it returns another wrapper to me, so I think we'd end up capturing this nested wrapper and double destroying it 😞.

We should probably hook into the `mount` and `shallowMount` functions instead. WDYT?

test/specs/wrapper.spec.js
```diff
@@ -15,4 +16,41 @@ describeWithShallowAndMount('Wrapper', mountingMethod => {
...
+      enableAutoDestroy(callback => {
+        hookCallback = callback
+      })
+      const wrapper = mountingMethod({ template: '<p>con tent</p>' })
+      sandbox.spy(wrapper, 'destroy')
+
+      hookCallback()
+
+      expect(wrapper.destroy).calledOnce
+    })
```
souldzin, line 46: **praise:** Nice way of testing this! 👍

souldzin (commented):
Thanks for the ping and iterating on this @winniehell! Back to you :soccer:

packages/test-utils/src/wrapper.js
winniehell: brought this up in https://github.com/vuejs/vue-test-utils/issues/1224#issuecomment-497658163

packages/test-utils/src/wrapper.js
winniehell: I have no idea what happened here. Probably being too tired when I thought this would  be a good idea. :sleeping:

packages/test-utils/src/wrapper.js
winniehell: > we'd end up capturing this nested wrapper and double destroying it

@souldzin yes, I think you are right. Even though nothing explodes when destroying twice, it may significantly decrease performance.

> We should probably hook into the mount and shallowMount functions instead.

We would not auto destroy wrappers created via `createWrapper()` or `new Wrapper` then. :thinking:

winniehell: https://github.com/vuejs/vue-test-utils/pull/1248 would allow us to filter out wrappers created by `wrapper.find()` / `wrapper.findAll()`.

winniehell:
It looks like we want to solve https://github.com/vuejs/vue-test-utils/issues/1224 and https://github.com/vuejs/vue-test-utils/issues/1135 first. :smiley:

packages/test-utils/src/wrapper.js
souldzin: Maybe we can tell in the ctor if the component is a "root" component or not 🤔... I guess those are the only ones we'd want to automagically destroy.

winniehell: > Maybe we can tell in the ctor if the component is a "root" component 

yes, we can after https://github.com/vuejs/vue-test-utils/pull/1248 because only non-root wrappers would have `wrapper.selector` defined.

packages/test-utils/src/create-wrapper.js
```diff
@@ -3,16 +3,20 @@
...
+import { trackInstance } from './auto-destroy'
 
 export default function createWrapper(
   node: VNode | Component,
   options: WrapperOptions = {}
 ): VueWrapper | Wrapper {
   const componentInstance = node.child
   if (componentInstance) {
-    return new VueWrapper(componentInstance, options)
+    const wrapper = new VueWrapper(componentInstance, options)
```
dobromir-hristov, line 14: You are no longer returning the wrapper, wouldnt that continue the execution?

packages/test-utils/src/auto-destroy.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+
+import { throwError } from 'shared/util'
+import Wrapper from './wrapper'
+
+let wrapperInstances: Array<Wrapper>
```
dobromir-hristov: Why not instantiate with an empty array? Or this is a way to check if it was ever called before?

packages/test-utils/src/create-wrapper.js
```diff
@@ -3,16 +3,20 @@
...
+    trackInstance(wrapper)
   }
-  return node instanceof Vue
-    ? new VueWrapper(node, options)
-    : new Wrapper(node, options)
+  const wrapper =
```
dobromir-hristov, line 18: We are also skipping the return here.

dobromir-hristov (commented):
Looks good, two small comments.

packages/test-utils/src/auto-destroy.js
winniehell: > Or this is a way to check if it was ever called before?

@dobromir-hristov Exactly! We can introduce a boolean `isAutoDestroyEnabled` if the above is too subtle.

dobromir-hristov: I personally dont mind it like this, Its just a clever way of doing two things at once. A boolean would be easier to understand at first glance.

test/specs/wrapper.spec.js
```diff
@@ -15,4 +16,37 @@ describeWithShallowAndMount('Wrapper', mountingMethod => {
         .with.property('message', message)
     })
   })
+
+  describe('enableAutoDestroy', () => {
```
winniehell, line 20: We somehow need to reset state here: https://circleci.com/gh/vuejs/vue-test-utils/13058?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Not quite sure how yet... :thinking:

dobromir-hristov: I suspect the state inside the module is kept between the tests? This is normal as modules once imported, are used from the cache.

I know how to do this with Jest, however not sure with Mocha...

You could also export an internal method to clear it? That is how we did it with Vuelidate 1 before, for a similar case.

winniehell: Maybe we want a `disableAutoDestroy`?

dobromir-hristov: I dont see why someone would use it, but when there is a way to enable, should be a way to disable :)

winniehell: > I dont see why someone would use it

Yes, I agree. It was mainly meant for testing but I found a different solution now. :muscle:

dobromir-hristov: Wouldnt exporting a `reset` method, that sets the boolean to false have been enough? :D Object.assign-ing like that the state is clever, but also bending the rules hehe

winniehell: @dobromir-hristov I don't mind a `reset` either. :smiley: 

Do you see why my

```
import { autoDestroyState } from '~vue/test-utils/src/auto-destroy'
```

doesn't work though? (https://circleci.com/gh/vuejs/vue-test-utils/13062?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Do I have to publicly expose the state / reset function?

dobromir-hristov: I think you should use relative paths?

winniehell: ah, that's not very pretty. :grimacing: but :man_shrugging:

winniehell: That doesn't seem to work either. :thinking: 

https://circleci.com/gh/vuejs/vue-test-utils/13065?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

dobromir-hristov: Does it work on your local?

winniehell: nope :disappointed:

dobromir-hristov (approved):
Do you think we should document the `resetAutoDestroyState`? I am not sure if it makes sense for ppl to need it though 😆

lmiller1990:
I think if we have the feature, it should be documented. Maybe a sentence or two explaining when/why it's useful?

The code seems good, no problems here 👍

afontcu (approved):
LGTM! 👍 

Regarding documenting `resetAutoDestroyState`, we should do so, but I wouldn't stress it too much. Maybe just mention it in the `enableAutoDestroy` section?

dobromir-hristov:
Same, just mention it exists. Maybe add them to the API docs, because I find myself search there more, than in the guides.

PR #1247, opened 2019-05-30
chore: Use config file for Prettier

winniehell:
Introduces a [configuration file for Prettier](https://prettier.io/docs/en/configuration.html). This has the advantages of a cleaner `package.json` and better IDE support.


winniehell:
@eddyerburgh What do you think?

eddyerburgh (approved):
Looks good, thanks :)

PR #1248, opened 2019-05-30
feat: Store selector in Wrapper.find() / .findAll()

### LINKED ISSUE #1135: Always store selector in Wrapper created by wrapper.find(...)

winniehell:
Stores the selector that is passed to `Wrapper.find()` / `Wrapper.findAll()` as a property in the newly created `Wrapper` / `WrapperArray` instance.

closes https://github.com/vuejs/vue-test-utils/issues/1135

### DIFF
--- a/packages/test-utils/src/error-wrapper.js
+++ b/packages/test-utils/src/error-wrapper.js
@@ -17,9 +17,11 @@ const buildSelectorString = (selector: Selector) => {
 }
 
 export default class ErrorWrapper implements BaseWrapper {
+  selector: Selector
   selectorString: string
 
   constructor(selector: Selector) {
+    this.selector = selector
     this.selectorString = buildSelectorString(selector)
   }
 
--- a/packages/test-utils/src/error-wrapper.js
+++ b/packages/test-utils/src/error-wrapper.js
@@ -18,60 +18,56 @@ const buildSelectorString = (selector: Selector) => {
 
 export default class ErrorWrapper implements BaseWrapper {
   selector: Selector
-  selectorString: string
 
   constructor(selector: Selector) {
     this.selector = selector
-    this.selectorString = buildSelectorString(selector)
   }
 
   at(): void {
     throwError(
-      `find did not return ${
-        this.selectorString
-      }, cannot call at() on empty Wrapper`
+      `find did not return ${buildSelectorString(
+        this.selector
+      )}, cannot call at() on empty Wrapper`
     )
   }
 
   attributes(): void {
     throwError(
-      `find did not return ${
-        this.selectorString
-      }, cannot call attributes() on empty Wrapper`
-
+      `find did not return ${buildSelectorString(
+        this.selector
+      )}, cannot call attributes() on empty Wrapper`
     )
   }
 
   classes(): void {
     throwError(
-      `find did not return ${
-        this.selectorString
-      }, cannot call classes() on empty Wrapper`
-
+      `find did not return ${buildSelectorString(
+        this.selector
+      )}, cannot call classes() on empty Wrapper`
     )
   }
 
   contains(): void {
     throwError(
-      `find did not return ${
-        this.selectorString
-      }, cannot call contains() on empty Wrapper`
+      `find did not return ${buildSelectorString(
+        this.selector
+      )}, cannot call contains() on empty Wrapper`
     )
   }
 
   emitted(): void {
     throwError(
-      `find did not return ${
-        this.selectorString
-      }, cannot call emitted() on empty Wrapper`
+      `find did not return ${buildSelectorString(
+        this.selector
+      )}, cannot call emitted() on empty Wrapper`
     )
   }
 
   emittedByOrder(): void {
     throwError(
-      `find did not return ${
-        this.selectorString
-      }, cannot call emittedByOrder() on empty Wrapper`
+      `find did not return ${buildSelectorString(
+        this.selector
+      )}, cannot call emittedByOrder() on empty Wrapper`
     )
   }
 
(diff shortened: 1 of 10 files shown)


packages/test-utils/src/error-wrapper.js
```diff
@@ -1,56 +1,73 @@
...
+const buildSelectorString = (selector: Selector) => {
+  if (selector.type === REF_SELECTOR) {
+    return `ref="${selector.value.ref}"`
+  }
+
+  if (typeof selector.value === 'string') {
+    return selector.value
+  }
+
+  return 'Component'
```
winniehell, line 16: I believe that we could use the component name at least in some cases here. However I'd like to keep that as a follow-up.

packages/test-utils/src/error-wrapper.js
```diff
@@ -1,56 +1,73 @@
...
 
 export default class ErrorWrapper implements BaseWrapper {
-  selector: string
+  selector: Selector
+  selectorString: string
 
-  constructor(selector: string) {
+  constructor(selector: Selector) {
     this.selector = selector
+    this.selectorString = buildSelectorString(selector)
```
winniehell: maybe building the string on demand is sufficient?

packages/test-utils/src/wrapper.js
```diff
@@ -214,7 +212,9 @@ export default class Wrapper implements BaseWrapper {
     const wrappers = nodes.map(node => {
       // Using CSS Selector, returns a VueWrapper instance if the root element
       // binds a Vue instance.
-      return createWrapper(node, this.options)
+      const wrapper = createWrapper(node, this.options)
+      wrapper.selector = selector
+      return wrapper
     })
     return new WrapperArray(wrappers)
```
winniehell: should we store the selector in `WrapperArray`, too?

winniehell (commented):
I believe most of it is done but I'd like to add some more tests here.

packages/test-utils/src/wrapper.js
winniehell: yes, we should because `class WrapperArray implements BaseWrapper`

packages/test-utils/src/error-wrapper.js
```diff
@@ -1,57 +1,72 @@
...
+  if (typeof selector.value === 'string') {
+    return selector.value
+  }
+
+  return 'Component'
+}
 
 export default class ErrorWrapper implements BaseWrapper {
-  selector: string
+  selector: Selector
```
winniehell, line 20: This is a breaking change. Are we ok with that?

winniehell:
@eddyerburgh Can you please take a look?

packages/test-utils/src/error-wrapper.js
eddyerburgh: Yes, there are big breaking changes in the next release anyway (removing sync mode)

eddyerburgh:
Thanks, look great. Before we merge can you update the type definition for Wrapper?

winniehell:
> Before we merge can you update the type definition for Wrapper?

@eddyerburgh I'm sorry, I'm not completely sure what you mean by that. :sweat: 

The flow definition has already been updated: https://github.com/vuejs/vue-test-utils/pull/1248/files#diff-dff4d7307e5e694d79396aacecd4399e

Do I need to update anything else, too?

winniehell:
ah, I think I found what you were referring to: https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/types/index.d.ts

I will update that file in a bit. :+1:

ghost:
@eddyerburgh Can you please take another look here?

winniehell:
@eddyerburgh Any chance to get this merged?

packages/test-utils/types/index.d.ts
```diff
@@ -69,6 +69,7 @@ interface BaseWrapper {
 
   trigger (eventName: string, options?: object): void
   destroy (): void
+  selector: Selector | void
```
eddyerburgh, line 72: This isn't actually a Selector type, it's an object with a value and type: https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/get-selector.js#L38

Because the library isn't written in TS, we have tests (https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/types/test/wrapper.ts) that access and set the wrapper options using the types. Can you write a test that accesses selector.value and selector.type and update this types file.

winniehell: > This isn't actually a Selector type

@eddyerburgh Thank you for catching this! :+1:

It actually is of type Selector for the ErrorWrapper: https://github.com/vuejs/vue-test-utils/blob/70ac83a95324841ea9839228bf9eb2ab6603a3ad/packages/test-utils/src/error-wrapper.js#L22

I will update the other places to use `rawSelector` (which is of type Selector), too, because that prevents us from defining a new type and consumers of the API don't need to do `selector.value` to get the selector they provided in the first place.

> we have tests (https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/types/test/wrapper.ts)

Sure, I can add tests there! :+1: 

Right now I'm a little confused by the structure of that file though because it has neither test cases in the usual style (no `it('something happens', () => { ... })` / `test(...)` blocks) nor assertions (only implicit ones). Could you provide me with an example of how such a test would look like?

winniehell: > It actually is of type Selector for the ErrorWrapper: 

Ah, it turns out this was wrong actually—but the Selector type is currently `any`:

https://github.com/vuejs/vue-test-utils/blob/70ac83a95324841ea9839228bf9eb2ab6603a3ad/flow/wrapper.flow.js#L6

I will adjust the places that create ErrorWrapper instances to use the rawSelector as well then.

eddyerburgh: Sure, the test script just compiles the file to make sure there aren't any compiler errors when using the Vue Test Utils API, caused by incorrect typing.

A test would look something like this:

```js
let s: string = wrapper.selector.value
```

winniehell: Thanks! :bowing_man: I'll add the tests and ping you again once they are ready.

winniehell:
@eddyerburgh I have updated all places to output the rawSelector (that is passed to `find` / `findAll`), added some TypeScript tests, and CI is still green. :green_apple: :tada: 

Can you please have another look at this?

winniehell:
@eddyerburgh Please let me know if there is anything I can do to get this finished. I would love to work on the other contributions that depend on this.

JessicaSachs:
Hi @winniehell  thank you for the great work! Hoping to get this merged shortly.

afontcu (approved):

afontcu:
Nice addition! Looks like the PR has some conflicts with master, would you have the time to fix them?

Thanks for your patience! Hopefully we will land this soon :)

winniehell:
@afontcu done :heavy_check_mark:

winniehell:
Thank you @JessicaSachs! :bowing_man:

PR #1249, opened 2019-05-31
Update common-tips.md

7inspire:
expect(wrapper.html()).toContain('Emitted!')
so the text '触发！' change to  'Emitted!'


eddyerburgh (approved):
Thank you

PR #1251, opened 2019-06-02
docs: adds await-promises example

visualjerk:


eddyerburgh:
Thanks for the PR, but I'm not convinced we should add this. Currently we recommend the flush-promises npm package. What does this package do internally? Where can I access the source code (I assume you're the author of it).

visualjerk:
No worries :) We needed to test a transition component without using a stubbed version of the transition. While flush-promises is useful when it comes to immediately resolved promises, it does not do the job, when the promises actually take time to be resolved.

Await-promises stores all promises, that are invoked between `.collect()` and `.stop()`. Then `.wait()` simply performs a `Promise.all()` on these promises, that you can await.

You can find the [source code here](https://github.com/visualjerk/await-promises/blob/master/src/index.js).

eddyerburgh:
The library looks useful, but I don't think we should add it to the docs, we can't cover every corner case that can be encountered by testing, otherwise the documentation would become too large

PR #1257, opened 2019-06-11
Respect "hidden" attributes when testing Wrapper.isVisible()

stevegrunwell:
Elements that are hidden via the HTML `hidden` attribute were erroneously being reported as being visible by the `isVisible()` method, even though they were not actually visible in the DOM.


LinusBorg (approved):
Looking good!

eddyerburgh (approved):
Thanks 👍

PR #1283, opened 2019-07-26
adding 'f' and 't' key codes to keydown trigger

achhunna:
Hoping to add `f` and `t` key codes to `keydown.` `trigger` method. These two keys are used in media player apps, like YouTube, to trigger full screen and theater mode features. Also updated the docs with the key codes.


eddyerburgh:
Thankyou for the PR, but I think it would be strange to just add f and t keys, and I don't think we should add modifiers for all keys

Issue #1252, opened 2019-06-03
Clear/Reset emitted events


jpenna:
### What problem does this feature solve?
When building a component and checking for the events emitted by this component we use `component.emitted('eventX')[0]` and do the tests. On the next test, we build the component again and use the same assertion.

When the component is more complex, it would be better to just reset the emitted event instead of rebuilding the component. This is common behavior on testing libraries for function (sinon.js, Jest, for example).

### What does the proposed API look like?
`component.resetEmitted()` would be enough.

I can try doing it if this is a desired feature.

marcosnav:
I'd like this feature to be implemented, meanwhile I'm doing the following to reset/clear the emitted events:

`component._emitted[event] = []`

lmiller1990:
Hi! Sorry it took so long for someone to follow up on this.

I see why this is a valid concern, but couldn't this be as easily solved with a factory function? Eg:

```js
const createWrapper = () => {
  return mount(MyComplexComponent, {
    // ... a ton of options
  })
}
```

Then you test becomes:

```js
test('It works', () => {
  const wrapper = createWrapper()
  // that's it!
})
```

You can also make `createWrapper` take arguments if you want to set data, props etc.
Likely you want to reset the _entire_ component, not just the the emitted events? It seems more ideal to have an entirely fresh copy of the component for each test. There is a good article about it in the Vue.js cookbook: https://vuejs.org/v2/cookbook/unit-testing-vue-components.html#Real-World-Example

What do you think?

jpenna:
Hey @lmiller1990 ! This is an option, and that's how I was doing it, but recreating the component every time is a little over kill if we just want to test a behavior triggered from the same action again and again, for example. 

In the case I'm talking about, I'd like to reset just the emitted events, like `jest.clearAllMocks()` or  sinon's `spy.resetHistory()`.

Sorry I don't have a real use case to include here right now.

lmiller1990:
Thanks for the reply. 

If there is a real performance bottleneck with recreating the component, I think that is something we should fix.

At this point, I don't think this is feature something we will be including in Vue Test Utils. If this was to become a feature one day, I'd be in support of a `resetComponent` method, that would reset _everything_ to the base state. Since that can be accomplished with a factory function, though, this isn't something we will be including.

Thanks for your understanding!

jpenna:
[...]

PR #1289, opened 2019-08-02
clearEmitted method for Wrapper vm

marcosnav:
Proposal for wrapper.clearEmitted() method.

Addressing #1252 

"When the component is more complex, it would be better to just reset the emitted event instead of rebuilding the component. This is common behavior on testing libraries for function (sinon.js, Jest, for example)."


dobromir-hristov:
Closed, as per #1252

PR #1299, opened 2019-09-01
Correcting link in docs to Lerna

tomasbjerre:
Changing from `https://lernajs.io/` to `https://lerna.js.org/`.


eddyerburgh (approved):
Thanks :)

PR #1300, opened 2019-09-02
Change the way of setting props

### LINKED ISSUE #1140: Prop not updated when immediate watcher accessing it (async mode)
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/rndmerle/repro-vue-test-utils-sync/tree/setprops-watcher](https://github.com/rndmerle/repro-vue-test-utils-sync/tree/setprops-watcher)

### Steps to reproduce
git clone -b setprops-watcher https://github.com/rndmerle/repro-vue-test-utils-sync.git

yarn && yarn test

### What is expected?
Hello test passing, with logging on console:
```
foo undefined
foo xxx
```

Hello test **not** passing, with logging on console:
```
foo undefined
```

---
To make the test pass, either:
- switch immediate to false (and uncomment created() content)
- change the watcher handler function to make anything as long it's not accessing the foo prop

xanf:
Since we're already wrapping a component into a parent one we can get rid of explicitly setting props on component, instead modifying them being passed from parent component instead.

This fixes weird behaviors in async mode, when prop is immediately reset  to it's original value
Fixes #1140

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -1,10 +1,7 @@
 // @flow
-
-import Vue from 'vue'
 import pretty from 'pretty'
 import getSelector from './get-selector'
 import { REF_SELECTOR, FUNCTIONAL_OPTIONS, VUE_VERSION } from 'shared/consts'
-import config from './config'
 import WrapperArray from './wrapper-array'
 import ErrorWrapper from './error-wrapper'
 import { throwError, getCheckedEvent, isPhantomJS } from 'shared/util'
@@ -478,8 +475,6 @@ export default class Wrapper implements BaseWrapper {
    * Sets vm props
    */
   setProps(data: Object): void {
-    const originalConfig = Vue.config.silent
-    Vue.config.silent = config.silent
     if (this.isFunctionalComponent) {
       throwError(
         `wrapper.setProps() cannot be called on a functional component`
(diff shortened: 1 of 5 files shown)


eddyerburgh:
Thanks for the PR!

This PR fixes `setProps` for root components, but not child components.

For example:

```js
it.only('fails when run on child component', async () => {
  const prop1 = 'prop 1'
  const ChildComponent = {
    template: '<div class="prop1">{{prop1}}</div>',
    props: ['prop1']
  }
  const TestComponent = {
    template: '<child-component />',
    components: {
      ChildComponent
    }
  }
  const wrapper = mountingMethod(TestComponent)
  wrapper.find(ChildComponent).setProps({ prop1 })
  await Vue.nextTick()
  expect(wrapper.find('.prop-1').element.textContent).to.equal(prop1)
})
```

The solution is to only allow `setProps` on root-level components. Can you update `setProps` to trow an error if it's called on a non-root component?

xanf:
@eddyerburgh Nice catch. Will do - it seems perfectly logical for me to allow setProps only on root component :)

eddyerburgh:
Hi @xanf, are you planning on adding the fix?

xanf:
@eddyerburgh Thank you for ping. Felt a bit off my radar, will update PR this week

afontcu:
Hi @xanf! Did you find the opportunity to work on this? :)

xanf:
@afontcu thank you for ping, will have something on Monday I hope, was busy with upgrading GitLab codebase for beta.30 release :)

packages/test-utils/src/wrapper.js
```diff
@@ -480,38 +475,26 @@ export default class Wrapper implements BaseWrapper {
...
-        // $FlowIgnore : Need to call this twice to fix watcher bug in 2.0.x
-        this.vm[key] = data[key]
+      if (this.vm && this.vm.$parent) {
+        this.vm.$parent.vueTestUtils_childProps[key] = data[key]
       }
     })
     // $FlowIgnore : Problem with possibly null this.vm
     this.vm.$forceUpdate()
-    Vue.config.silent = originalConfig
+    // We need to explicitly trigger parent watcher to support sync scenarios
```
lmiller1990, line 518: I think we can remove this, sync mode is gone now.

test/specs/wrapper/setProps.spec.js
```diff
@@ -244,6 +245,17 @@ describeWithShallowAndMount('setProps', mountingMethod => {
     expect(wrapper.vm.propA).to.equal('value')
   })
 
+  it('correctly sets props in async mode when component has immediate watchers', async () => {
+    const wrapper = mountingMethod(ComponentWithWatchImmediate, {
+      sync: false
```
afontcu, line 250: I might be missing something here, but I think `sync` is gone as per #1141?

JessicaSachs: Correct.

JessicaSachs: I pulled it out of the src. That's a typo. Thanks for the catch

xanf:
Closing this, since #1618 was merged

PR #1304, opened 2019-09-08
Adding get method to Wrapper

### LINKED ISSUE #1298: get method on Wrapper
Sometimes you "look for something" and sometimes you "get something". If you look for something, it may not exists. If you get something, it is an error if it does not exist.

In my test cases I want to use a get method on the wrapper. And get a clear error describing that what I was looking for cannot be found.

In Wrapper.js

```javascript
  /**
   * Gets first node in tree of the current wrapper that
   * matches the provided selector.
   */
  get(rawSelector: Selector): Wrapper {
    const found = find(rawSelector)
    if (found instanceof ErrorWrapper) {
      throw new Error(`Unable to find ${rawSelector} within: ${this.html()}`)
    }
    return found
  }
```

tomasbjerre:
To have it fail with a clear error message when selector does not match an existing element.

Closes #1298


lmiller1990:
Code looks fine, but is this additional API necessary? I don't fully understand the use case. If you cannot `find` something, you will surely get an error when you try to use it, anyway, or you will be making an assertion can get a decent output from the test runner.

tomasbjerre:
> you will surely get an error when you try to use it

Yes, and it will not be helpful.

I dont want a generic error like:
```
FAIL src/components/FCalendar/FCalendar.spec.ts (12.516s)
  ● expect next month button event to change currentMonth from 1 to 2

    [vue-test-utils]: find did not return .next-month-viewasd, cannot call trigger() on empty Wrapper

      31 |     expect(wrapper.vm.$data.currentYear).toBe(2019);
      32 |     expect(wrapper.vm.$data.currentMonth).toBe(1);
    > 33 |     wrapper.find('.next-month-viewasd').trigger('click');
         |                                         ^
      34 |     expect(wrapper.vm.$data.currentMonth).toBe(2);
      35 | });
      36 |

      at throwError (node_modules/@vue/test-utils/dist/vue-test-utils.js:1417:9)
      at ErrorWrapper.trigger (node_modules/@vue/test-utils/dist/vue-test-utils.js:2264:3)
      at Object.<anonymous> (src/components/FCalendar/FCalendar.spec.ts:33:41)
```

Instead I want something like:

```
FAIL src/components/FCalendar/FCalendar.spec.ts (12.516s)
  ● expect next month button event to change currentMonth from 1 to 2

    [vue-test-utils]: Unable to find .next-month-viewasd within: <div><span>whatever html was renderd and did not contain the element.....</span</div>

      31 |     expect(wrapper.vm.$data.currentYear).toBe(2019);
      32 |     expect(wrapper.vm.$data.currentMonth).toBe(1);
    > 33 |     wrapper.find('.next-month-viewasd').trigger('click');
         |                                         ^
      34 |     expect(wrapper.vm.$data.currentMonth).toBe(2);
      35 | });
      36 |

      at throwError (node_modules/@vue/test-utils/dist/vue-test-utils.js:1417:9)
      at ErrorWrapper.trigger (node_modules/@vue/test-utils/dist/vue-test-utils.js:2264:3)
      at Object.<anonymous> (src/components/FCalendar/FCalendar.spec.ts:33:41)
```

I often find myself doing `console.log` just to see what the dom looks like, because the error message does not help me.

lmiller1990:
I agree it's useful to have better debugging messages - but adding an entire API doesn't seem like the correct solution. Rather, can we improve `find`?

One idea could be for `find` to throw the error you described if it doesn't find an element. Then if you do `wrapper.find(...).trigger`, the initial `find` that found nothing would throw an error before `trigger` gets called. What do you think?

tomasbjerre:
Changing find like that would break the API. Lots of people upgrading would get failing test cases because of that.

How would you assert that an element does, or does not, exist?

My point is that there is a difference between searching and getting. If you search with find, you may not find it, and that is totally ok.

lmiller1990:
Good point, I forgot the `expect(wrapper.find('does-not-exist').exists()).toBe(false)` type of assertion. I guess this could make sense.

I think the code is fine, will need to wait on a maintainer (@eddyerburgh , I guess?) to approve and merge.

Also, do you have a use case for `getAll`? It might be nice to match `find` and `findAll` APIs with `get` and `getAll`.

lmiller1990 (approved):
Nice job using the existing functionality of `find`.

lmiller1990:
@tomasbjerre  I'll be coming on board as a maintainer in the near future, and would like to get this one merged up I have the repo access to do so. I just realized since this does add a new API, we will need to add docs for it. Is this something you are able to do (if not, I can take a look at doing it).

tomasbjerre:
Ok! I can make an attempt =)

tomasbjerre:
I added some docs now.

docs/api/wrapper/get.md
```diff
@@ -0,0 +1,19 @@
+## get
+
+Works just like [find](../find.md) but will throw an error if selector not
```
lmiller1990: Can be worded a little better - "but will throw an error if nothing matching the given selector is found".

lmiller1990 (changes requested):
Just a small grammar improvement! Also, some tests seem to be failing.

tomasbjerre:
The failing test seems to be toggling, looks like a timeout. I dont think that is introduced in this PR.

docs/api/wrapper/get.md
tomasbjerre: I made that change.

tomasbjerre (commented):
Changes made

lmiller1990:
LGTM, will merge when I get merge rights in the next week or two. Nice job!

lmiller1990 (approved):

lmiller1990:
I had another idea inspired by [this repo](https://github.com/hmsk/jest-matcher-vue-test-utils) in response to your earlier question, "How would you assert that an element does, or does not, exist?"

Could this make sense as a custom matcher? `expect(wrapper.find('#el')).notToExist()` reads and feels really nice to me, much more so than `expect(wrapper.find('#el').exists()).toBe(false)`.

tomasbjerre:
I often test a condition where I assume existens. And that is when the get method is useful 

I don't want to spend my time and write code to test existens. I simply want to assume it exists and if it does not, i want a clear error. I'd like my test cases to be clear about other conditions and those conditions implicitly assumes the element actually exists. Like asserting the `html()` equals `5`, the `5` ofcourse means that the element both exists and equals `5`. In this case it is nice to have an error stating it was, perhaps, `4 not 5`. And also, the implicit existens assertion, could fail with `the element with selector X does not exist in html Y`.

 * Looks like you are suggestion different ways of testing existens. And I argue that existens is not what I want to test but instead simply assume.

 * I also don't see how those assertions helps me figure out what the dom looked like. I want to avoid having to console.log the wrapper.html function.

 * I also don't see how anyone can browse the API and find such a method. If I type "`wrapper.`" then my IDE will tell me there is a "`get`" method.

dobromir-hristov (approved):
What is the status on this? Generally this looks OK to me, hope ppl dont get confused when to use what.
You have added docs, so thats good.

docs/api/wrapper/get.md
```diff
@@ -0,0 +1,19 @@
+## get
+
+Works just like [find](../find.md) but will throw an error if nothing matching
+the given selector is found. You should use `find` when searching for an element
+that may not exist. You should use this method when getting an element that should
+exist and it will provide a nice error message if that is not the case.
+
+```js
```
afontcu, line 8: Should we add an example showing the "happy path" for `get`? I mean, just getting an existing element in addition to showcase that it would throw otherwise.

afontcu:
Should `get` also throw an error if more than 1 matching element is found?

docs/api/wrapper/get.md
lmiller1990: done

lmiller1990:
@afontcu I think that it's fine not to throw an error for 1+ elements. If that is a use case, we should include a `getAll` function (perhaps `getAll({ count: n })`).

Although `findAll` would probably be fine for that.

lmiller1990:
Fixed merge conflicts, let's merge on CI passing. I wish GH had the feature GitLab has (auto merge on pass). We can dream!

PR #1308, opened 2019-09-13
docs: Add example for 'listeners' mount option

rahulcs:
I thought an example to understand usage of 'listeners'  would be useful & make it more obvious.

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -248,6 +248,23 @@ Please see [Other options](#other-options).
 
 Set the component instance's `$listeners` object.
 
+Example:
+
+```js
+const Component = {
+  template: '<div v-on:click="$emit('click')"><slot></slot></div>'
+}
+const onClick = jest.fn()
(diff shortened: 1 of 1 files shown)


JessicaSachs:
👍🏻 Thank you! Will merge shortly.

docs/api/options.md
```diff
@@ -248,6 +248,23 @@ Please see [Other options](#other-options).
 
 Set the component instance's `$listeners` object.
 
+Example:
+
+```js
+const Component = {
+  template: '<div v-on:click="$emit(\'click\')"><slot></slot></div>'
```
afontcu: Looks great!

Could we keep this as simple and as focused as possible?

```suggestion
  template: '<button v-on:click="$emit(\'click\')"></button>'
```

(extra point for using accessible examples 😇)

rahulcs: Thanks for reviewing 🙂

afontcu (approved):

PR #1309, opened 2019-09-13
generate named slots content for stubbed components via shallowMount

### LINKED ISSUE #1307: Named slots' content is not generated for stubbed components via shallowMount
Higher-order components make use of named slots. It's possible to test higher-order components via shallowMount because named slot content can be mocked, but it's not possible to test components which consume higher-order components because their generated named slot output is not available in the stubbed content.

https://github.com/vuejs/vue-test-utils/pull/782 added support for outputting default slot content, but not named slots.

To be clear, an example of a component which needs this support is:

```
<!-- myComponent.vue -->
<template>
  <HigherOrderComponent foo="bar">
    default slot content
    <template v-slot="header">header slot content</template>
    <template v-slot="footer">footer slot content<HigherOrderComponent foo="waldo"/></template>
  </HigherOrderComponent>
</template>
```

Currently, this is stubbed out to only:

`<higherordercomponent-stub foo="bar">default slot content</higherordercomponent-stub>`

The end user experience should be that components which generate named slot content "just work" when testing with shallowMount.

The API would not change.

chriswa:
resolves #1307

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -110,6 +110,9 @@ export function createStubFromComponent(
   // ignoreElements does not exist in Vue 2.0.x
   if (Vue.config.ignoredElements) {
     Vue.config.ignoredElements.push(tagName)
+    if (Vue.config.ignoredElements.indexOf('template-stub') === -1) {
+      Vue.config.ignoredElements.push('template-stub')
+    }
   }
 
   return {
@@ -117,30 +120,38 @@ export function createStubFromComponent(
     $_vueTestUtils_original: originalComponent,
     $_doNotStubChildren: true,
     render(h, context) {
-      return h(
-        tagName,
-        {
-          ref: componentOptions.functional ? context.data.ref : undefined,
-          attrs: componentOptions.functional
-            ? {
-                ...context.props,
-                ...context.data.attrs,
-                class: createClassString(
-                  context.data.staticClass,
-                  context.data.class
-                )
-              }
-            : {
-                ...this.$props
-              }
-        },
-        context
-          ? context.children
-          : this.$options._renderChildren ||
-              getScopedSlotRenderFunctions(this).map(x =>
-                this.$options.parent._vnode.data.scopedSlots[x]()
-              )
+      const attrs = componentOptions.functional
+        ? {
+            ...context.props,
+            ...context.data.attrs,
+            class: createClassString(
+              context.data.staticClass,
+              context.data.class
+            )
+          }
+        : {
+            ...this.$props
+          }
+
+      const slots = context ? context.slots() : this._renderProxy.$slots
+
+      // ensure consistent ordering of slots (default first, then alphabetical)
+      const sortedSlotEntries = Object.entries(slots)
+      sortedSlotEntries.sort(([slotNameA], [slotNameB]) =>
+        slotNameA === 'default'
+          ? -1
+          : slotNameB === 'default'
+          ? 1
+          : slotNameA.localeCompare(slotNameB)
       )
+
+      const children = sortedSlotEntries.map(([slotName, slotChildren]) =>
+        slotName === 'default'
+          ? slotChildren
+          : h('template-stub', { attrs: { slot: slotName } }, slotChildren)
+      )
+
+      return h(tagName, { attrs }, children)
     }
   }
 }
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -5,8 +5,8 @@ import {
   throwError,
   camelize,
   capitalize,
-  hyphenate,
-  keys
+  hyphenate
+  // keys
 } from '../shared/util'
 import {
   componentNeedsCompiling,
@@ -80,24 +80,41 @@ function resolveOptions(component, _Vue) {
   return options
 }
 
-function getScopedSlotRenderFunctions(ctx: any): Array<string> {
-  // In Vue 2.6+ a new v-slot syntax was introduced
-  // scopedSlots are now saved in parent._vnode.data.scopedSlots
-  // We filter out the _normalized and $stable key
-  if (
-    ctx &&
-    ctx.$options &&
-    ctx.$options.parent &&
-    ctx.$options.parent._vnode &&
-    ctx.$options.parent._vnode.data &&
-    ctx.$options.parent._vnode.data.scopedSlots
-  ) {
-    const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
-    return keys(slotKeys).filter(x => x !== '_normalized' && x !== '$stable')
-  }
-
-  return []
+function getAttributes(
+  functional: boolean,
+  functionalCtx: any,
+  componentCtx
+): Object {
+  return functional
+    ? {
+        ...functionalCtx.props,
+        ...functionalCtx.data.attrs,
+        class: createClassString(
+          functionalCtx.data.staticClass,
+          functionalCtx.data.class
+        )
+      }
+    : componentCtx.$props
 }
+
+// function getScopedSlotRenderFunctions(ctx: any): Array<string> {
+//   // In Vue 2.6+ a new v-slot syntax was introduced
+//   // scopedSlots are now saved in parent._vnode.data.scopedSlots
+//   // We filter out the _normalized and $stable key
+//   if (
+//     ctx &&
+//     ctx.$options &&
+//     ctx.$options.parent &&
+//     ctx.$options.parent._vnode &&
+//     ctx.$options.parent._vnode.data &&
+//     ctx.$options.parent._vnode.data.scopedSlots
+//   ) {
+//     const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
+//     return keys(slotKeys).filter(x => x !== '_normalized' && x !== '$stable')
+//   }
+//
+//   return []
+// }
 
 export function createStubFromComponent(
   originalComponent: Component,
@@ -120,19 +137,7 @@ export function createStubFromComponent(
     $_vueTestUtils_original: originalComponent,
     $_doNotStubChildren: true,
     render(h, context) {
-      const attrs = componentOptions.functional
-        ? {
-            ...context.props,
-            ...context.data.attrs,
-            class: createClassString(
-              context.data.staticClass,
-              context.data.class
-            )
-          }
-        : {
-            ...this.$props
-          }
-
+      const attrs = getAttributes(componentOptions.functional, context, this)
       const slots = context ? context.slots() : this._renderProxy.$slots
 
       // ensure consistent ordering of slots (default first, then alphabetical)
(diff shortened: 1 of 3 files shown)


chriswa:
Hi! This is my first ever OSS contribution. Please let me know any feedback you may have! :)

chriswa:
I did create an Issue for this PR, but I guess I failed to link it:
https://github.com/vuejs/vue-test-utils/issues/1307

ndchelsea:
This is a super helpful fix... What's the chances of this being merged?

GiacomoVoss:
Any updates on this? It's really helpful and would make snapshot testing with Vuetify components a lot better.

lmiller1990:
Hey @chriswa , awesome first OSS contribution. I will review this now and give you feedback so w can get this merged.

packages/create-instance/create-component-stubs.js
```diff
@@ -86,31 +86,48 @@ export function createStubFromComponent(
...
+            )
+          }
+        : {
+            ...this.$props
+          }
+
+      const slots = context ? context.slots() : this._renderProxy.$slots
+
+      // ensure consistent ordering of slots (default first, then alphabetical)
+      const sortedSlotEntries = Object.entries(slots)
```
lmiller1990, line 144: Nice usage of `Object.entries`

packages/create-instance/create-component-stubs.js
```diff
@@ -86,31 +86,48 @@ export function createStubFromComponent(
...
+      sortedSlotEntries.sort(([slotNameA], [slotNameB]) =>
+        slotNameA === 'default'
+          ? -1
+          : slotNameB === 'default'
+          ? 1
+          : slotNameA.localeCompare(slotNameB)
```
lmiller1990, line 150: I would have used a simple `>` and `<` for the string compare. Not sure it makes a big difference - it is common to use characters like `á` in code? (my language does not have those characters, so I have never seen this before).

If we are doing to support using `localeCompare`, maybe we should have a test containing a `á` character vs an `a` character. Although probably not strictly necessary.

test/specs/shallow-mount.spec.js
```diff
@@ -90,8 +91,40 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'shallowMount', () => {
...
+        '    <child-stub></child-stub>\n' +
+        '  </template-stub>\n' +
+        '  <template-stub slot="header">Hello</template-stub>\n' +
+        '</child-stub>'
+    )
+  })
+
+  it('renders named slots for functional components', () => {
+    const localVue = createLocalVue()
+    localVue.component('child', {
```
lmiller1990: nice job! it's easy to forget the case of functional components.

packages/create-instance/create-component-stubs.js
```diff
@@ -86,31 +86,48 @@ export function createStubFromComponent(
   // ignoreElements does not exist in Vue 2.0.x
   if (Vue.config.ignoredElements) {
     Vue.config.ignoredElements.push(tagName)
+    if (Vue.config.ignoredElements.indexOf('template-stub') === -1) {
```
lmiller1990, line 130: Why is this necessary?

lmiller1990 (commented):
This is an awesome first OSS contribution. Left a few comments, only one really requires any action (first one about ignored elements).

Might also be good to rebase on the last develop. Great job! I can't wait to get this one merged in.

lmiller1990:
@chriswa left some comments, let me know what you think. Excited to get this one in.

packages/create-instance/create-component-stubs.js
dobromir-hristov: Just a thought, this kind of ternary is super confusing. I would deff advise for some sort of reordering or other.

packages/create-instance/create-component-stubs.js
dobromir-hristov: So the `template-stub` stays as `<template-stub />` in the finished render, correct?

packages/create-instance/create-component-stubs.js
```diff
@@ -86,31 +86,48 @@ export function createStubFromComponent(
...
+      const sortedSlotEntries = Object.entries(slots)
+      sortedSlotEntries.sort(([slotNameA], [slotNameB]) =>
+        slotNameA === 'default'
+          ? -1
+          : slotNameB === 'default'
+          ? 1
+          : slotNameA.localeCompare(slotNameB)
+      )
+
+      const children = sortedSlotEntries.map(([slotName, slotChildren]) =>
```
dobromir-hristov: So this should render the default slot, and all others as children of the `template-stub`? Wont this be a super breaking change for snapshots?

lmiller1990: True, this will break all snapshots using the original behavior. We probably should have some snapshot tests in VTU as well to capture these regressions. But then we will need to move to Jest, I think? Or we can do a manual comparison between strings.

Probably not the biggest deal, we just need to make sure that's clear in release notes. Something like:

NOTE: You will need to update all snapshots when updating.

packages/create-instance/create-component-stubs.js
```diff
@@ -80,25 +80,42 @@ function resolveOptions(component, _Vue) {
...
+//     ctx.$options.parent._vnode.data &&
+//     ctx.$options.parent._vnode.data.scopedSlots
+//   ) {
+//     const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
+//     return keys(slotKeys).filter(x => x !== '_normalized' && x !== '$stable')
+//   }
+//
+//   return []
+// }
+
```
JessicaSachs, line 118: So @lmiller1990 I'm mostly concerned about why no tests break when I comment this out

packages/create-instance/create-component-stubs.js
```diff
@@ -5,8 +5,8 @@ import {
   throwError,
   camelize,
   capitalize,
-  hyphenate,
-  keys
+  hyphenate
+  // keys
```
JessicaSachs, line 9: Commented out for formatting until we can figure out how to merge everything in.

packages/create-instance/create-component-stubs.js
```diff
@@ -110,36 +127,53 @@ export function createStubFromComponent(
...
+      // if (context) {
+      //   children = sortedSlots
+      // } else if (this.$options._renderChildren) {
+      //   children = this.$options._renderChildren
+      // } else {
+      //   children = getScopedSlotRenderFunctions(this).map(x =>
+      //     this.$options.parent._vnode.data.scopedSlots[x]()
+      //   )
+      // }
+
```
JessicaSachs, line 169: This was initially how I expanded out and merged dev and this PR, but was failing so I commented it out and reverted to mainly this PR's version of rendering slots

test/specs/shallow-mount.spec.js
```diff
@@ -130,7 +136,11 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'shallowMount', () => {
     expect(wrapper.find({ name: 'Foo' }).exists()).to.equal(true)
     expect(wrapper.find('.new-example').exists()).to.equal(true)
     expect(wrapper.html()).to.equal(
-      '<foo-stub>\n' + '  <p class="new-example">text</p>\n' + '</foo-stub>'
+      '<foo-stub>\n' +
+        '  <template-stub slot="newSyntax">\n' +
```
JessicaSachs, line 140: The need to update all the assertions with `template-stub` worry me. This will be a breaking change :/

afontcu:
Hi! As mentioned [here](https://github.com/vuejs/vue-test-utils/issues/246#issuecomment-608739968), we're not merging this before v1 is released.

Just in case someone stumbles upon this PR – there are 2 workarounds:

1. use mount
2. write your own stub that implements the named slots


thanks for the effort put on this one, though! 🤗

lmiller1990:
I thought about this a bit more, I'm not convinced this behavior is actually correct. I think children of stubbed components, including slots, should not be rendered. I am in favor of closing this as a wontfix.

We need to figure out what shallowMount should/shouldn't do.

lmiller1990:
This looks to be a significant a breaking change at this point. I also am not sure this actually makes sense - I think stubbed children should not render their slots.

I'm going to close this one. Thanks for the work - sorry the team was unresponsive and this didn't get the attention it should have when the PR was opened.

We are looking to a v1 release soon, and we will revisit shallowMount and what it should be rendering for the Vue 3 integration.

chenxeed:
Hi @lmiller1990 , just sharing my opinion here regarding your thought:

> I think children of stubbed components, including slots, should not be rendered.

I'm afraid if there's multiple interpretation of defining this use-case.

I agree that generally, `shallowMount` shouldn't render the child component and simply stub them.

However, slot is different. Slot is a way for the parent to "inject" the content they expect in the child component. The child is responsible for the placement of the slot, but it's the parent responsibility to ensure the content injected are working as expected.

Thus, testing the slot content is the parent responsibility.

For example, when we have a component that needs to load a modal component like [b-modal](https://bootstrap-vue.org/docs/components/modal) with specific content for the named slot:

```vue
// parent component
<template>
  <b-modal v-model="visible">
    <template #modal-header>
      {{ title }}
      <button @click="onClickModalHeader">
    </template>
</b-modal>
```

On testing the parent component, I don't care with how the `<b-modal>` renders the slot since I believe `b-modal` will do its job properly. I just want to make sure the content of the header are rendering the title that I expect and the button inside are working.

_note: example with jest_

```js
const parentComponentWrapper = shallowMount(parentComponent);

parentComponent.setData({ title: 'The Parent Title' });
const modal = parentComponent.findComponent({ name: 'BModal' });
expect(modal.html()).toContain('The Parent Title'); // better if can assert which slot it's expecting
modal.find('button').trigger('click')
// assert the side effect from `onClickModalHeader`
```

The idea is, slots are used by parent for the parent's logic, and we need to test them too. Using `mount` to test them will defeat the purpose of `shallowMount` to only render components from the parent layer.

I beg you to reconsider :bow:  and let me know if my example above doesn't makes sense or not applicable. Thank you :bow:

dobromir-hristov:
This is taken care of in Vue 2. Default slots are rendered behind a flag, named slots only via a special helper. :)

chenxeed:
Thanks to respond @dobromir-hristov 

> named slots only via a special helper

Can you clarify what's the special helper you mentioned about?

Is it the manual stub that @danieltian suggested:
```ts
const wrapper = shallowMount(RootComponent, {
  stubs: { ChildComponent }
}

expect(wrapper.find(ChildComponent).text()).toContain('some text');
```

dobromir-hristov:
Sort of. In shallowMount default slots can be auto rendered, but not named and scoped slots. We cant know that data, as it is deeply buried in the render function tree of the components you are stubbing. Does it make sense?

I tried to make this for a while, bit there is just no sensible way to do render all slots, while still stubbing.

chenxeed:
@lmiller1990 @dobromir-hristov  I've seen the discussion has been going on heavily in #1307 which has better explanation than mine. My appreciation to you and the core team for the stable release, and taking in the concern of this issue :heart: 

> Does it make sense?

Yes, seems like the issue here is more towards the implementation, not the goal behind.

I agree with @lmiller1990 [suggestion](https://github.com/vuejs/vue-test-utils/issues/1307#issuecomment-617503092) that this behavior can be made as optional.
I'll followup the discussion from #1307 and look forward in the next vue-test-utils.

Again, thank you! :1st_place_medal:

PR #1316, opened 2019-10-08
Add required `namespaced` property

zukilover:


kazupon:
Thank you for your feedback!

Here, this example is sample for vue-test-utils.
We don't have to be so strict about Vuex.

Thanks!

PR #1318, opened 2019-10-10
docs(vi): using-with-vuex

luubinhan:


kazupon:
Unfortunately 😞 , In about new translation docs, we don't accept, because we don't always have the resources to maintain high-quality translation documents.

You can fork it, and maintain it. 😄

PR #1320, opened 2019-10-14
Template querySelector support

earnubs:
Handle selectors passed in the `template` option of the Vue instance (https://vuejs.org/v2/api/#template), currently test-utils gives an unfriendly error as the selector is passed as a template string:

```
browser.js:954 [Vue warn]: Error compiling template:

#foo

- Component template requires a root element, rather than just text.
```

Repro here: https://jsfiddle.net/earnubs/c5d2mea7/

### DIFF
--- a/test/specs/mount.spec.js
+++ b/test/specs/mount.spec.js
@@ -162,9 +162,8 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'mount', () => {
   })
 
   itDoNotRunIf(
-    process.env.TEST_ENV === 'node',
-    'compiles templates from querySelector',
-    () => {
+    !(navigator.userAgent.includes && navigator.userAgent.includes('node.js')),
+  'compiles templates from querySelector', () => {
       const template = window.createElement('div')
       template.setAttribute('id', 'foo')
       template.innerHTML = '<div>foo</div>'
--- a/test/specs/mount.spec.js
+++ b/test/specs/mount.spec.js
@@ -161,6 +161,26 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'mount', () => {
     expect(wrapper.html()).to.equal(`<div>foo</div>`)
   })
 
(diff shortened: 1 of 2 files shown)


dobromir-hristov:
A few notes:
1. I have not thought about this case tbh, but I can see it being useful for people not using bundlers.
2. lets remove the dist files, we can generate these on release.
3. The test has a check, you can use the `itDoNotRunIf`.
4. I am not sure how good of an idea it is to use `document` or `window` directly like that.

lmiller1990:
This is a great addition! Nice job.

As pointed out, try something like `itDoNotRunIf(TEST_ENV === "node")`. Full API for `conditional-specs`: https://github.com/eddyerburgh/conditional-specs.

Your suspicion about mutating window/document were correct @dobromir-hristov. For this reason, before and after each test, the window variable is reset, to keep things clean. See [here](https://github.com/vuejs/vue-test-utils/blob/3115396f9b2e991b290ce8c7c7b27218e368d70e/test/specs/mount.spec.js#L21).

I'm happy to approve this once

- [ ] use `itDoNotRunIf`
- [ ] remove `dist` files

Are you able to make these changes, @earnubs ?

test/specs/mount.spec.js
```diff
@@ -161,6 +161,26 @@ describeRunIf(process.env.TEST_ENV !== 'node', 'mount', () => {
     expect(wrapper.html()).to.equal(`<div>foo</div>`)
   })
 
+  it('compiles templates from querySelector', () => {
+    if (
+      !(navigator.userAgent.includes && navigator.userAgent.includes('node.js'))
```
lmiller1990: see comments

lmiller1990 (changes requested):

earnubs:
Test is failing, I'll take a look at this tonight when I have a bit more time.

lmiller1990 (approved):

PR #1336, opened 2019-10-30
Fixed syntax mistake (`shallowMount` => `mount`)

nonsocode:
`shallowMount` doesn't render children components so the example would not have worked. I changed it to `mount`


vvanpo:
`shallowMount()` just _stubs_ child components, said stubs are still `Vue` instances. So calling `$emit()` on stubbed components still works as expected.

kazupon (approved):

PR #1338, opened 2019-10-30
Added async/await example

karatekaneen:
Added example of using async/await on functions that are using Promises under the hood.


kazupon (approved):
Thanks!

PR #1340, opened 2019-11-02
Use `mount()` consistently in Wrapper documentation

vvanpo:
The `setSelected()` documentation examples imported `mount`, but called `shallowMount()`.


kazupon (approved):
Thanks!

PR #1353, opened 2019-11-22
[Manual] Amending commands to install the testing frameworks

lacrioque:
If you want to add testing to an already working project, the search for the  command to install the correct plugins can be tedious, thus the request to add this in the "getting started" paragraph.


lmiller1990 (approved):
Seems good, I'll just do `yarn format` to get this passing on CI and merge it up. Thanks 👍

PR #1354, opened 2019-11-22
Add provided/inject support for Composition API components

Stoom:
It looks like with the new composition api components that the provider changed from `vm.provide` to `vm._provided`.  This PR adds both so that composition apis that use injection can be tested.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "javascript.validate.enable": false,
+    "flow.pathToFlow": "${workspaceFolder}/node_modules/.bin/flow"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,4 @@
 {
-    "javascript.validate.enable": false,
-    "flow.pathToFlow": "${workspaceFolder}/node_modules/.bin/flow"
+  "javascript.validate.enable": false,
+  "flow.pathToFlow": "${workspaceFolder}/node_modules/.bin/flow"
 }
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "eslint-plugin-flowtype": "^2.46.1",
     "eslint-plugin-markdown": "^1.0.0-beta.6",
     "eslint-plugin-vue-libs": "^2.1.0",
-    "flow-bin": "^0.111.0",
+    "flow-bin": "^0.66.0",
     "jsdom": "^12.0.0",
     "jsdom-global": "^3.0.2",
     "karma": "^3.1.4",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "test:types": "tsc -p packages/test-utils/types && tsc -p packages/server-test-utils/types"
   },
   "dependencies": {
-    "@vue/composition-api": "^0.3.2",
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.2.2",
     "babel-loader": "^7.1.3",
@@ -89,5 +88,8 @@
     "vuex": "^3.0.1",
     "webpack": "^3.0.1",
     "webpack-node-externals": "^1.6.0"
+  },
+  "devDependencies": {
+    "@vue/composition-api": "^0.3.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "eslint-plugin-flowtype": "^2.46.1",
     "eslint-plugin-markdown": "^1.0.0-beta.6",
     "eslint-plugin-vue-libs": "^2.1.0",
-    "flow-bin": "^0.66.0",
+    "flow-bin": "^0.111.0",
     "jsdom": "^12.0.0",
     "jsdom-global": "^3.0.2",
     "karma": "^3.1.4",
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "test:types": "tsc -p packages/test-utils/types && tsc -p packages/server-test-utils/types"
   },
   "dependencies": {
+    "@vue/composition-api": "^0.3.2",
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.2.2",
     "babel-loader": "^7.1.3",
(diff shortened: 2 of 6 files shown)


Stoom:
The latest build fail is flow issues in `wrapper`.  Looks to be an issue with the `dev` branch

test/resources/components/component-with-inject-composition.vue
```diff
@@ -0,0 +1,20 @@
...
+</template>
+
+<script>
+import { createComponent, inject } from '@vue/composition-api'
+
+export default createComponent({
+  name: 'component-with-inject-composition',
+  setup: () => {
+    const fromMount = inject('fromMount')
+    const setInBeforeCreate = 'created'
```
dobromir-hristov: What is the purpose of this `setInBeforeCreate` property? We dont use it in the assertions.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
dobromir-hristov, line 1: Probably not a great idea. We should not keep such files in the repo.

dobromir-hristov (dismissed):
Looks good.

package.json
```diff
@@ -31,6 +31,7 @@
     "test:types": "tsc -p packages/test-utils/types && tsc -p packages/server-test-utils/types"
   },
   "dependencies": {
+    "@vue/composition-api": "^0.3.2",
```
JessicaSachs: I believe this should be in devDependencies because it's only used during testing.

.vscode/settings.json
JessicaSachs: Hmmm... I'm not a VSCode user. I can go download VSCode and try to setup our repo with it, but from googling, it seems like this file is needed for flow https://code.visualstudio.com/docs/languages/javascript#_can-i-use-other-javascript-tools-like-flow

test/resources/components/component-with-inject-composition.vue
JessicaSachs: It was previously used in a similar Resource test component ([here](https://github.com/vuejs/vue-test-utils/blob/be4033e2e22440255441a9ab9e1fe717a2c456a0/test/resources/components/component-with-inject.vue)) so I think he was trying to maintain parity with the previous example. (setup now maps to "beforeCreate" and "created" seen [here](https://github.com/vuejs/composition-api-rfc/blob/master/api.md#lifecycle-hooks))

We should test for it in the assertions.

JessicaSachs (changes requested):
Thank you so much for this PR 🎉! Two small changes and one question:

1. Can you place the composition API in the devDependencies?
2. Can you assert that `setInBeforeCreate` is equal to `created` in your test?

Question: Would it be helpful for you as a VSCode user if the settings.json file was there when you first cloned the project?

package.json
Stoom: @JessicaSachs  I've placed it in `devDependencies`, but I think there's a lot that can go in there (`devDependencies` is being introduced in this PR for the `package.json`)

JessicaSachs: Good point. I need to clean that up

JessicaSachs (approved):

JessicaSachs:
The tests are failing due to some `prettier` issue on dev that has since been fixed. Can you rebase and re-push?

Stoom:
Should be good now @JessicaSachs

test/specs/mounting-options/provide.spec.js
```diff
@@ -86,8 +106,23 @@ describeWithShallowAndMount('options.provide', mountingMethod => {
...
+        localVue
+      })
+
+      expect(wrapper.html()).to.contain('_')
+    }
+  )
+
   it('config with function throws', () => {
-    config.provide = () => {}
+    config.provide = () => { }
```
JessicaSachs: Ah. This is the guy breaking the linter. `npm run format` should clean that up for you in the future. My apologies for the runaround

Stoom: 🤞

JessicaSachs:
I think the flow upgrade also caused a failure... Lemme revert the flow bin locally and test what happens

Stoom:
Looks like flow errors?

JessicaSachs:
Yeah, pin the version to 0.66.0 and re-run yarn install. That fixes the errors for me.

Stoom:
> Yeah, pin the version to 0.66.0 and re-run yarn install. That fixes the errors for me.

`^0.66.0`?

JessicaSachs:
^0.66.0 Will be fine for now. I'm probably going to pin it at 0.66.0 after the merge, until we have a chance to fix the Element issues so someone doesn't accidentally upgrade it.

test/specs/mounting-options/provide.spec.js
```diff
@@ -58,6 +61,23 @@ describeWithShallowAndMount('options.provide', mountingMethod => {
     }
   )
 
+  itDoNotRunIf(
+    !injectSupported || mountingMethod.name === 'renderToString',
+    'supports setup in composition api component',
+    () => {
+      if (!injectSupported) return
```
dobromir-hristov, line 68: What is the point of this? It is already checked above?

test/specs/mounting-options/provide.spec.js
```diff
@@ -58,6 +61,23 @@ describeWithShallowAndMount('options.provide', mountingMethod => {
...
+
+      const localVue = createLocalVue()
+      localVue.use(VueCompositionApi)
+      const wrapper = mountingMethod(CompositionComponentWithInject, {
+        provide: { fromMount: '_' },
+        localVue
+      })
+
+      expect(wrapper.vm.setInSetup).to.equal('created')
```
dobromir-hristov, line 77: How is this connected to provide/inject?

dobromir-hristov (commented):
Those test cases look a bit unpolished. Overall is OK.

.vscode/settings.json
afontcu: I had to configure this myself too when setting up the local env yesterday. Not sure if committing it here is the best approach (I've done so in my projects), but it's gonna be needed for every VSC user.

PR #1362, opened 2019-12-04
fix: whitelist tags to be ignored with `disabled` attribute

### LINKED ISSUE #1321: 'disabled' attribute prevents triggering events on all HTML tags
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/micgro42/vue-test-utils-disabled-bug](https://github.com/micgro42/vue-test-utils-disabled-bug)

### Steps to reproduce
1. check out minimal reproduction repository
2. run `npm install`
3. run `npm run test:unit`

### What is expected?
Test should pass, because the `disabled` attribute on a link (HTML `<a>` tag) is ignored by browsers and thus the component should handle the event (in this case: emit it to the parent component/wrapper)

Test fails, because wrapper.trigger() does nothing 😭.

---
This was caused by #424 in order to fix #395.
Unfortunately, this solution is too far-reaching. This also disables events, for example, on links with the `disabled` attribute.
While that might be the desired final state, it is not something that is provided by the browser and thus needs to be programmed and asserted by tests. Just plainly disabling the wrapper.trigger method in that case prevents me from testing this.

hiendv:
This PR resolves #1321.
The list follows [W3C Recommendation](https://www.w3.org/TR/2014/REC-html5-20141028/disabled-elements.html#disabled-elements) with the addition of `<command>` and `<keygen>` as listed by [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled).

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -574,8 +574,25 @@ export default class Wrapper implements BaseWrapper {
       )
     }
 
-    // Don't fire event on a disabled element
-    if (this.attributes().disabled) {
+    /**
+     * Avoids firing events on specific disabled elements
+     * See more: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
+     */
+
+    const supportedTags = [
+      'BUTTON',
+      'COMMAND',
+      'FIELDSET',
+      'KEYGEN',
+      'OPTGROUP',
+      'OPTION',
+      'SELECT',
+      'TEXTAREA',
+      'INPUT'
+    ]
+    const tagName = this.element.tagName
+
+    if (this.attributes().disabled && supportedTags.indexOf(tagName) > -1) {
       return
     }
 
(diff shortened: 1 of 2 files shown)


dobromir-hristov (commented):
Lets split this into two tests. One for disabled elements(buttons, etc), one for those like anchors. 

Just so it's clear? Maybe even a custom component in there?

Overall I think the PR makes very much sense.

hiendv:
So should I write a list of tests for valid `disabled` elements: `<button>`, `<input>`, etc and a list of tests for others? I can imagine two blocks for these lists would be lengthy because each element has different events to test on. Refactoring these tests using a loop would be possible though. Thanks for your review.

dobromir-hristov:
No, not for all (I dont think its necessary), just one for disablable (lol) elements and one for such that cant be (like anchor).

hiendv:
I think the PR is ready for your review again @dobromir-hristov ;)

dobromir-hristov (approved):

afontcu (approved):
LGTM!

PR #1373, opened 2019-12-16
Guide on using vue test utils without a build step

### LINKED ISSUE #1192: Documentation on how to use test-utils without babel/webpack
The documentation explains how to use vue-test-utils with different runners and build setups but not how to test a plain component without babel / webpack. Since Vue itself works great without using single file components or a build step it would be great to provide at least one simple example on how to use vue-test-utils in that case.

I have tried some obvious ways and asked for help on the forum and SO more than two weeks ago but since no one could help this really seems to be not documented at all.

https://forum.vuejs.org/t/how-to-test-plain-vue-components-not-single-file-components/58415
https://stackoverflow.com/questions/55052983/how-to-test-plain-vue-components-not-single-file-components

```js
var testUtils=require('@vue/test-utils'), Vue=require('vue');

require('jsdom-global')();    

testUtils.mount({ template: '<div>test</div>' });
// or
testUtils.mount(Vue.component('test', { template: '<div>test</div>' }));
```
would be obvious but fails with this error:
```
@vue/test-utils/dist/vue-test-utils.js:2471
var componentInstance = node.child;
TypeError: Cannot read property 'child' of undefined
```

lmiller1990:
Add a guide explaining how to use Vue test utils without a test runner (just Node.js).

resolves #1192

### DIFF
--- a/docs/guides/usage-without-a-build-step-node.md
+++ b/docs/guides/usage-without-a-build-step-node.md
@@ -20,7 +20,7 @@ Now we need to require the libraries. There is a slight caveat, explained in a c
 // jsdom-global must be required before vue-test-utils,
 // because vue-test-utils expects a DOM (real DOM, or JSDOM)
 // to exist.
-require('jsdom-global')();
+require('jsdom-global')()
 
 const assert = require('assert')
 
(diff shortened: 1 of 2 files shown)


docs/guides/usage-without-a-build-step-node.md
```diff
@@ -0,0 +1,55 @@
+## Usage Without a Build Step
+
+While it is common to build Vue applications using tools [Webpack](https://jestjs.io/) to bundle the application, `vue-loader` to leverage Single File Components, and [Jest](https://jestjs.io/) to write expressive tests, it is possible to use Vue Test Utils with much less. The minimal requirements for Vue Test Utils, aside from the library itself are:
```
JessicaSachs: The link to webpack is incorrect.

JessicaSachs (changes requested):
Great guide. There’s an incorrect link, but that’s it.

dobromir-hristov (approved):

afontcu (approved):

JessicaSachs (approved):

PR #1374, opened 2019-12-17
fix: translation missing

### LINKED ISSUE #1267: There is no translation of attachToDocument in Japanese document
I want to improve the Japanese document.
https://github.com/vuejs/vue-test-utils/blob/dev/docs/api/options.md#attachtodocument

There is no traslation of "When attaching to the DOM, you should call wrapper.destroy() at the end of your test to remove the rendered elements from the document and destroy the component instance"

We can't understand it. Could I translate it?

Domにアタッチされた際に、テストの最後でwrapper.destroy()を呼び出さなければなりません。レンダリングされたelementをdocumentから取り除いて、コンポーネントインスタンスを壊さなければならないからです。

kazupon:
closes #1267

### DIFF
--- a/docs/ja/api/options.md
+++ b/docs/ja/api/options.md
@@ -194,6 +194,8 @@ expect(wrapper.vm.$route).toBeInstanceOf(Object)
 
 `true` に設定されている場合、描画時にコンポーネントは DOM にアタッチされます。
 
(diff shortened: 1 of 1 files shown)


kazupon:
ping @JessicaSachs 
I've fixed!

kazupon:
This PR is translated from translation issue #1267.
we will merge due to correct.

Issue #41, opened 2017-09-05
Add component instances to slots on mount


alexjoverm:
Is there a way to somehow pass component instances to the slots property of mount? The use case is, say I have a component that has a required property:

```javascript
export default {
    name: 'Message',
    props: {
      message: {
        type: String,
        required: true,
        validator: value => value.length > 1
      }
  }
```

Then if in a test I wanna add a list of Message as the default slots, I cannot do it like it's documented: I'll get a `"message" property is required` error:

```javascript
mount(MessageList, {
  slots: {
    default: messages
  }
})
```

I tried using mount and getting the vm out of it, or also extending the component and using `new Vue(...)` but I get the error `[Vue warn]: Failed to mount component: template or render function not defined.`:

```javascript
beforeEach(() => {
  const indexes = [0, 1, 2]
  const messages = indexes.map(i => mount(Message, {
    propsData: { message: `Message ${i}` }
  }).vm)

  cmp = mount(MessageList, {
    slots: {
      default: messages
    }
  })
})
```

So, how's it possible to accomplish that?

eddyerburgh:
There isn't currently a way to pass mounted components to slots.

There's some work being done in avoriaz to add the option to pass Wrappers as slots. But right now there's an unsolved bug - https://github.com/eddyerburgh/avoriaz/issues/113.

I'd like to add this functionality, but don't have time to work on this feature right now. I'd welcome a PR and could give you some guidance if you were willing to make one.

alexjoverm:
Thanks, right now I also don't have the time, but probably in the coming days I can get some and lend a hand in there. I'll contact you for that guidance in that case ;).

alexjoverm:
FYI, as a workaround, this seems to work:

```javascript
mount(MessageList, {
  slots: {
    default: {
      render(h) {
        return h(Message, { props: { message: 'hey' }  })
      }
    }
  }
})
```

mouafa:
try this:
`
const tableCol = table.find(TableColumn)
expect(tableCol.vm. show).toBe('firstName')
`

eddyerburgh:
We aren't going to implement this at the moment. You can pass a component, but not an instance.

phobetron:
@eddyerburgh Is this functionality anywhere on the roadmap? I have some functionality that can't really be tested without access to instance properties, and workarounds don't always work well.

eddyerburgh:
It's  not on the roadmap. I would review a PR if somebody wants to add it.

[...]

PR #1376, opened 2019-12-17
docs: add an example for slots components with props

hiendv:
Because many people including me (:rofl:) were [struggling to mount slots components with props](https://github.com/vuejs/vue-test-utils/issues/41#issuecomment-343901582), this PR introduces an example for the case.
Example:
```vue
<!-- Component.vue -->
<template>
  <div><slot/></div>
</template>

<!-- Child.vue -->
<template>
  <p>{{ foo }}</p>
</template>
<script>
export default {
  props: {
    foo: {
      type: String,
      required: true
    }
  }
}
</script>

<!-- Expected mounting -->
<component>
  <child foo="bar"></child>
  <child foo="qux"></child>
</component>
```
Btw, I don't really know if this fits into the `api` docs or common tips.

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -68,7 +68,7 @@ const yourComponent = {
       required: true
     }
   },
-  render (h) {
+  render(h) {
     return h('p', this.foo)
   }
 }
@@ -83,11 +83,13 @@ const wrapper = shallowMount(Component, {
     qux: '<my-component />',
     quux: '<your-component foo="lorem"/><your-component :foo="yourProperty"/>'
   },
-  stubs: { // used to register custom components
+  stubs: {
+    // used to register custom components
     'my-component': MyComponent,
     'your-component': yourComponent
   },
-  mocks: { // used to add properties to the rendering context
+  mocks: {
+    // used to add properties to the rendering context
     yourProperty: 'ipsum'
   }
 })
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -74,6 +74,37 @@ const wrapper = shallowMount(Component, {
 expect(wrapper.find('div')).toBe(true)
 ```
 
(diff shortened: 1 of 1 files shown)


afontcu (approved):
LGTM 👍  I think it would be "too much" to add this to the `Common tips` section, as this is not a common scenario – as far as I can tell!

docs/api/options.md
```diff
@@ -74,6 +74,37 @@ const wrapper = shallowMount(Component, {
 expect(wrapper.find('div')).toBe(true)
 ```
 
+You can also provide slots components with props.
```
JessicaSachs: I like this a lot, but can you add the context for _why_ this is useful? If you used your Pull Request description (including component code snippets) that would be perfect. It didn't make any sense until I saw what the template for "Component" looked like.

JessicaSachs (changes requested):
Just needs a little more context to explain to the developer what kind of Components this will help test. Everything looks great implementation-wise ✅

docs/api/options.md
hiendv: I'm working on it and trying to merge two snippets into one. The original snippet is not including the stubs for `<my-component />` so should I remove mine also?

hiendv: Line 70: 
```
- qux: '<my-component />'
+ qux: '<my-component foo="bar" :foo="val"/>'
```
and `mocks` for `val`?

JessicaSachs: Ah! I see... I missed the original snippet above 😄If you'd like to merge and refine yours more, please feel free to. I do like the full context that your PR example has and think it would help illustrate the "Why" a bit more

JessicaSachs (approved):
💯Super! Thank you.

Issue #1295, opened 2019-08-14
Issue with Vue config keyCode aliases and triggering keydown events


tbcorr:
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/tbcorr/vue-test-utils-issue](https://github.com/tbcorr/vue-test-utils-issue)

### Steps to reproduce
Run the unit tests and look at the failing tests.

### What is expected?
I would expect keydown event handlers bound to specific keyCode aliases would only be called when the specific key is pressed.

### What is actually happening?
It appears that keydown event handlers bound to keydown events with custom keyCode aliases are being called for all keydown events, regardless of the key which triggered the event.

---
I'm not sure if this is only related to triggering keydown events. It just so happens I noticed this issue while testing keydown events. I'm also not sure if this is only related to the keys I was testing which are Home, End, Page Up, Page Down. Also, in browser I get the expected behavior which makes me think this is related to vue-test-utils.

aholzner:
This seems to be working with me.
```
import { shallowMount } from '@vue/test-utils';
import App from '@/App';

describe('App', () => {
  describe('when pressing the home key', () => {
    let wrapper;

    beforeAll(() => {
      wrapper = shallowMount(App);

      wrapper.find('.keys').trigger('keydown', {
        key: 'home',
      });
    });

    it('should update homePressed', () => {
      expect(wrapper.vm.homePressed).toBe(true);
    });

    it('should not update endPressed', () => {
      expect(wrapper.vm.endPressed).toBe(false);
    });

    it('should not update upPressed', () => {
      expect(wrapper.vm.upPressed).toBe(false);
    });

    it('should not update downPressed', () => {
      expect(wrapper.vm.downPressed).toBe(false);
    });

    it('should not update pageUpPressed', () => {
      expect(wrapper.vm.pageUpPressed).toBe(false);
    });

    it('should not update pageDownPressed', () => {
      expect(wrapper.vm.pageUpPressed).toBe(false);
    });
  });
});

```

I also noticed that supplying key code doesn't work.

tbcorr:
@aholzner thanks for the comment! I modified my tests to to match how you triggered the keydown event and my tests are now passing, which is great!

I guess the issue only manifests when using the vue-test-utils key aliases vs specifying the key in the event options.

aholzner:
@tbcorr Awesome! 👍

phegman:
[...]

Issue #1285, opened 2019-07-31
Unable to trigger keydown event with custom keyCode


igor-nabebin:
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/igor-nabebin/vue-test-utils-jest-example](https://github.com/igor-nabebin/vue-test-utils-jest-example)

### Steps to reproduce
1. Clone repo
2. Run "npm install"
3. Run "npm run test"

### What is expected?
A keydown event with keyCode = 65 is fired, test passed.

### What is actually happening?
A keydown event with keyCode = 0 is fired, test failed.

andrewvasilchuk:
Encountered the same problem.

PR #1378, opened 2019-12-18
fix: use the keyCodes provided in options from wrapper.trigger

JessicaSachs:
This will address #1285 and #1295 once merged.

Previously `wrapper.trigger('keydown', { keyCode: 65 })` would trigger a keydown event with a `keyCode` of `0`. This is incorrect.

To fix this, we must listen to the keyCode if the modifier doesn't exist.

### DIFF
--- a/packages/test-utils/src/create-dom-event.js
+++ b/packages/test-utils/src/create-dom-event.js
@@ -24,36 +24,46 @@ const modifiers = {
   pagedown: 34
 }
 
-function createEvent(
-  type,
-  modifier,
-  { eventInterface, bubbles, cancelable },
-  options
-) {
+function getOptions(eventParams) {
+  const { modifier, meta, options } = eventParams
+  const keyCode = modifiers[modifier] || options.keyCode || options.code
+
+  return {
+    ...options, // What the user passed in as the second argument to #trigger
+
+    bubbles: meta.bubbles,
+    meta: meta.cancelable,
+
+    // Any derived options should go here
+    keyCode,
+    code: keyCode
+  }
+}
+
+function createEvent(eventParams) {
+  const { eventType, meta = {} } = eventParams
+
   const SupportedEventInterface =
-    typeof window[eventInterface] === 'function'
-      ? window[eventInterface]
+    typeof window[meta.eventInterface] === 'function'
+      ? window[meta.eventInterface]
       : window.Event
 
-  const event = new SupportedEventInterface(type, {
+  const event = new SupportedEventInterface(
+    eventType,
     // event properties can only be added when the event is instantiated
     // custom properties must be added after the event has been instantiated
-    ...options,
-    bubbles,
-    cancelable,
-    keyCode: modifiers[modifier]
-  })
+    getOptions(eventParams)
+  )
 
   return event
 }
 
-function createOldEvent(
-  type,
-  modifier,
-  { eventInterface, bubbles, cancelable }
-) {
+function createOldEvent(eventParams) {
+  const { eventType, modifier, meta } = eventParams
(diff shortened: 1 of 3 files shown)


packages/test-utils/src/create-dom-event.js
```diff
@@ -35,14 +72,12 @@ function createEvent(
...
+  const event = new SupportedEventInterface(
+    type,
     // event properties can only be added when the event is instantiated
     // custom properties must be added after the event has been instantiated
-    ...options,
-    bubbles,
-    cancelable,
-    keyCode: modifiers[modifier]
-  })
+    sanitizeOptions(type, modifier, { bubbles, cancelable }, options)
```
JessicaSachs: Ah. I think the cheapest fix for this would be the following, which keeps feature parity and will let the modifier win over the keyCode.

```js
{
    ...options,
    bubbles,
    cancelable,
    keyCode: modifiers[modifer] || options.keyCode || options.code
}
```

packages/test-utils/src/create-dom-event.js
```diff
@@ -24,6 +24,19 @@ const modifiers = {
   pagedown: 34
 }
 
+function getOptions(modifier, { bubbles, cancelable }, options) {
```
dobromir-hristov: Maybe make the third param part of the second? No need for a third separate parameter.

packages/test-utils/src/create-dom-event.js
```diff
@@ -24,6 +24,19 @@ const modifiers = {
...
+  const keyCode = modifiers[modifier] || options.keyCode || options.code
```
dobromir-hristov, line 29: `code` and `keyCode` are already spread in, if provided, here we are just ensuring we dont unnecessarily overwrite then with modifiers, right?

packages/test-utils/src/create-dom-event.js
JessicaSachs: Keeping parity with the calling method.

packages/test-utils/src/create-dom-event.js
JessicaSachs: Yes. Previously the bug was that modifiers was falsey, yet still overwriting the keyCode.

packages/test-utils/src/create-dom-event.js
```diff
@@ -24,6 +24,19 @@ const modifiers = {
...
+
+  const derivedOptions = { keyCode, code: keyCode }
```
afontcu: I'm sure I might be missing something here, but why do we need to create the `derivedOptions` variable instead of directly adding them to the return statement?

```js
  return {
    ...options, // What the user passed in as the second argument to #trigger
    bubbles,
    cancelable,
    keyCode,
    code: keyCode
  }
```

JessicaSachs: So that it’s very clear what order wins in the override/merge.  We’re gonna have more props in the future. The extra object acts as a logical grouping and communicates why certain properties win over others.

JessicaSachs: I’m happy to in-line it and use a comment block as a delimiter if that’s preferred

packages/test-utils/src/create-dom-event.js
JessicaSachs: After fixing the failing Karma tests, I refactored the method signatures to be more compact

lmiller1990 (approved):
Looked over the previous comments that have been addressed, this all seems reasonable - will give it a quick test tonight (and try find edge cases) but code looks ok. Nice job!!

dobromir-hristov (approved):

afontcu (approved):

PR #1380, opened 2019-12-19
Ensure setChecked() and setSelected() only trigger DOM events when state is changed (fix #1339)

### LINKED ISSUE #1339: wrapper.setSelected() triggers a change event even if the option is already selected
### Version
1.0.0-beta.29

### Reproduction link
[https://codepen.io/vvanpo/pen/MWWroOK](https://codepen.io/vvanpo/pen/MWWroOK)

### Steps to reproduce
1. Mount a component with a `<select>` element.
2. Call `setSelect()` multiple times on one of the select element's options.
3. Observe the `<select>` element wrapper's `emitted('change').length` is equal to the number of times `setSelect()` was called.

### What is expected?
A `<select>` element emits a `change` event only when the element value has actually changed.

`setSelected()` triggers a `change` event every time it is called, regardless of whether the underlying value was changed.

vvanpo:
`Wrapper.setChecked()` and `Wrapper.setSelected()` should not trigger DOM events when called on already checked or selected elements.

Closes #1339

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -375,6 +375,10 @@ export default class Wrapper implements BaseWrapper {
         )
       }
 
+      if (this.element.checked === checked) {
+        return
+      }
+
       if (event !== 'click' || isPhantomJS) {
         // $FlowIgnore
         this.element.selected = true
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -404,6 +404,10 @@ export default class Wrapper implements BaseWrapper {
     }
 
     if (tagName === 'OPTION') {
+      if (this.element.selected) {
+        return
+      }
+
       // $FlowIgnore
       this.element.selected = true
       // $FlowIgnore
--- a/test/specs/wrapper/setChecked.spec.js
+++ b/test/specs/wrapper/setChecked.spec.js
@@ -57,22 +57,22 @@ describeWithShallowAndMount('setChecked', mountingMethod => {
 
   it('should trigger a change event when called on a checkbox', () => {
     const listeners = { change: sinon.spy() }
-    const input = mountingMethod({
-      template: `<input type="checkbox" v-on="$listeners">`,
-    }, { listeners })
 
-    input.setChecked()
+    mountingMethod(
+      { template: `<input type="checkbox" v-on="$listeners">` },
+      { listeners },
+    ).setChecked()
 
     expect(listeners.change).to.have.been.called
   });
 
   it('should not trigger a change event if the checkbox is already checked', () => {
     const listeners = { change: sinon.spy() }
-    const input = mountingMethod({
-      template: `<input type="checkbox" checked v-on="$listeners">`,
-    }, { listeners })
 
-    input.setChecked()
+    mountingMethod(
+      { template: `<input type="checkbox" checked v-on="$listeners">` },
+      { listeners },
+    ).setChecked()
 
     expect(listeners.change).not.to.have.been.called
   });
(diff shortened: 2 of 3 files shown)


test/specs/wrapper/setChecked.spec.js
```diff
@@ -55,6 +55,43 @@ describeWithShallowAndMount('setChecked', mountingMethod => {
     expect(wrapper.find('.counter').text()).to.equal('4')
   })
 
+  it('should trigger a change event when called on a checkbox', () => {
```
afontcu:
```suggestion
  it('triggers a change event when called on a checkbox', () => {
```

test/specs/wrapper/setChecked.spec.js
```diff
@@ -55,6 +55,43 @@ describeWithShallowAndMount('setChecked', mountingMethod => {
...
+          @click="listener"
+        >
+      `,
+      methods: { listener }
+    }).setChecked()
+
+    expect(listener).to.have.been.called
+  })
+
+  it('should not trigger a change event if the checkbox is already checked', () => {
```
afontcu:
```suggestion
  it('does not trigger a change event if the checkbox is already checked', () => {
```

test/specs/wrapper/setChecked.spec.js
```diff
@@ -89,6 +126,41 @@ describeWithShallowAndMount('setChecked', mountingMethod => {
     expect(wrapper.find('.counter').text()).to.equal('4')
   })
 
+  it('should trigger a change event when called on a radio button', () => {
```
afontcu:
```suggestion
  it('triggers a change event when called on a radio button', () => {
```

test/specs/wrapper/setChecked.spec.js
```diff
@@ -89,6 +126,41 @@ describeWithShallowAndMount('setChecked', mountingMethod => {
...
+          @click="listener"
+        >
+      `,
+      methods: { listener }
+    }).setChecked()
+
+    expect(listener).to.have.been.called
+  })
+
+  it('should not trigger a change event if the radio button is already checked', () => {
```
afontcu:
```suggestion
  it('does not trigger a change event if the radio button is already checked', () => {
```

test/specs/wrapper/setSelected.spec.js
```diff
@@ -39,6 +39,44 @@ describeWithShallowAndMount('setSelected', mountingMethod => {
     expect(wrapper.text()).to.contain('selectA')
   })
 
+  it('should trigger a change event on the parent <select>', () => {
```
afontcu:
```suggestion
  it('triggers a change event on the parent select', () => {
```

test/specs/wrapper/setSelected.spec.js
```diff
@@ -39,6 +39,44 @@ describeWithShallowAndMount('setSelected', mountingMethod => {
...
+      methods: { change }
+    })
+      .findAll('option')
+      .at(1)
+      .setSelected()
+
+    expect(change).to.have.been.called
+  })
+
+  it('should not trigger an event if already selected', () => {
```
afontcu:
```suggestion
  it('does not trigger an event if called on already selected option', () => {
```

afontcu (commented):
Looks good! I've added some comments related to test titles, to keep them consistent with previous ones :)

Thank you for this! 🙌

afontcu (approved):
This looks good to me! 👍 

Pinging @dobromir-hristov and @JessicaSachs as they shared their thoughts on the issue :)

packages/test-utils/src/wrapper.js
```diff
@@ -375,6 +375,10 @@ export default class Wrapper implements BaseWrapper {
         )
       }
 
+      if (this.element.checked === checked) {
```
dobromir-hristov, line 378: We already do this on line 356 for `checkboxes`, I guess this is for `radio` as well?
I dont think there is another type of input element, that requires this, no?

dobromir-hristov (commented):
Just a minor comment :)

packages/test-utils/src/wrapper.js
vvanpo: Exactly, yeah I don't think there is any other types of input that require this.
I had found that this check used to exist when `setChecked()` was first introduced, but it got removed for radios during a refactor, likely by mistake.

lmiller1990 (approved):
Has an approve and all tests passing, let's merge this up. Nice job!!

PR #1382, opened 2019-12-21
Add commit message linting

dobromir-hristov:
## Description

Adds commit message linting before each commit.
Applies the conventional-commit pattern, also known as Angular preset.


dobromir-hristov:
Netlify's node version is 8.x where one of the packages depends on v10.x
Should I try do bump down the version of the package or should we set a higher version of Node on Netlify?

JessicaSachs:
I’d like to upgrade our node version if it is not problematic. [8.x is reaching EOL.](https://github.com/nodejs/Release#release-schedule) (AWS isn’t supporting lambdas written in it as of 2020... I imagine other providers will follow suit)

I think we should make a new issue to track that work.

afontcu:
I have mixed feelings here.

Having conventional commits is a great idea, but adding a commit linting step for everyone might feel like a blocker when starting to contribute to the library.

If we use a **squash merge** strategy, we can customise the commit message that ends up in dev/master, so users could use whatever message they like, and we'd be the ones responsible for setting the right message.

This way we would have the best of both worlds: Angular-based commit messages while keeping it simple for contributors.

I've been using this approach both internally and in [DOM Testing Library](https://github.com/testing-library/dom-testing-library/commits/master) (and other Testing Lib projects) and works like a charm.

Thoughts?

dobromir-hristov:
Well I can see where you are coming from here and you have a fair point.
Users could be a bit intimidated when they see a commit error, but then again, its a not rocket science to do this. Also helps us to see what happened in the commit.

I looked at Vue, Vue-cli and other similar libs and all use that pattern.

Other way of doing this would be to drop commit linting, but still allow them to use `yarn commit` to use Commitizen for an easy message format, if they desire.

afontcu:
> I looked at Vue, Vue-cli and other similar libs and all use that pattern.

Okay, that's a fair point.

Sure, let's go with this and provide the `commit` script to make it easier.


Thanks!

.github/CONTRIBUTING.md
```diff
@@ -34,6 +35,10 @@ Hi! I’m really excited that you are interested in contributing to Vue Test Uti
   - Provide detailed description of the bug in the PR. Live demo preferred.
   - Add appropriate test coverage if applicable.
 
+### Committing Changes
+
+Commit messages should follow the [commit message convention](./COMMIT_CONVENTION.md) so that changelogs can be automatically generated. Commit messages will be automatically validated upon commit. If you are not familiar with the commit message convention, you can use `npm run commit` pr `yarn commit` instead of `git commit`, which provides an interactive CLI for generating proper commit messages.
```
afontcu:
```suggestion
Commit messages should follow the [commit message convention](./COMMIT_CONVENTION.md) so that changelogs can be automatically generated. Commit messages will be automatically validated upon commit. If you are not familiar with the commit message convention, you can use `npm run commit` or `yarn commit` instead of `git commit`, which provides an interactive CLI for generating proper commit messages.
```

dobromir-hristov:
Should I change it so prettier and esling run a `fix` command, and we stage the changes? This way users dont need to figure out how to run them, it just happens for them?

afontcu:
> Should I change it so prettier and esling run a `fix` command, and we stage the changes? This way users dont need to figure out how to run them, it just happens for them?

I don't see why not!

dobromir-hristov:
Now it auto runs lint and prettier, then stages the changes when you commit. 
This should prevent issues with errors from eslint or prettier. Works similarly to how it is in Vue core.
Also set a default recommended Node version to 10

afontcu (approved):

Issue #1261, opened 2019-06-19
new slot syntax template can't be accessed in shallowMount


manonrichou:
### Version
1.0.0-beta.29

### Reproduction link
[https://codesandbox.io/embed/vue-template-dr6i7?fontsize=14&previewwindow=tests](https://codesandbox.io/embed/vue-template-dr6i7?fontsize=14&previewwindow=tests)

### Steps to reproduce
Run the two tests in the link. The one pointing to the old syntax's slot will pass, the one pointing to the new slot's syntax won't. 

### What is expected?
Template with v-slot syntax can be accessed in the test when using shallow mount.

### What is actually happening?
The test pointing to the template using the new slot syntax fails, whereas the one with the  soon-to-be deprecated syntax passes.

runarberg:
Are there any workarounds (other then using the deprecated syntax)?

kspackman:
Very interested in this issue, so that we can upgrade our syntax and still be able to test our components.

Stoom:
> Are there any workarounds (other then using the deprecated syntax)?

What I've found is you need to provide your own stub for the component with slots.  If you do that and add the named slots you care about then they will render in the shallowMount.

```typescript
wrapper = shallowMount(Appbar, {
    store, router, localVue, vuetify, stubs: {
        VTooltip: '<v-tooltip-stub><slot></slot><slot name="activator"></slot></v-tooltip-stub>',
    },
});
```

runarberg:
@Stoom That does not help if you need to test something that is inside a higher order component that uses scoped slots. E.g. in the following example, anything inside `<SomeHigherOrderComponent>` will be stubbed out:

```vue
<SomeHigherOrderComponent v-slot="scope">
  <p>
    We can use <code>scope</code> here,
    but <code>shallowMount</code> will stub this out.
    So we have to use <code>mount</code> to test it.
  </p>
</SomeHigherOrderComponent>

runarberg:
Perhaps this is relevant, but [vuese](https://github.com/vue-contrib/vuese) is experiencing a similar issue of not being able to access content inside slots that use the new `v-slot` syntax.

https://github.com/vue-contrib/vuese/issues/119

Stoom:
@runarberg When I was testing VHover from vuetify I just did the following and it worked enough for  me....

```typescript
        wrapper = shallowMount(Settings, {
            propsData, localVue, vuetify, store, stubs: {
                VHover: `<slot></slot>`,
            },
[...]

PR #1383, opened 2019-12-22
Support v-slot

lmiller1990:
Working on https://github.com/vuejs/vue-test-utils/issues/1261

- support the new `v-slot` syntax!
- need to bump to Vue v2.6
- likely a component combining the new and old scoped syntax will fail; we need to cover that case, it should be pretty easy, though.
- will need to rebase and fix my commits, the commit convention is wrong
- not sure if this is the best place to do this logic; likely we should have a simple util function to handle all the null checks.
- added two tests; one with SFC, one without. Likely we can nuke one... just wanted to be thorough.


JessicaSachs:
On the surface, this looks good to me. Like you said, we probably want to refactor it into a function to hide the nested checks... but aside from that 👍 

I'm curious if $parent will always be present -- can root elements have scopedSlots? I'm particularly thinking of components constructed with createElement.

lmiller1990:
@JessicaSachs will need to investigate more to answer you questions, but got around to doing a refactor.

My commit messages aren't styled property, will use the squash feature on Github to have a single nice commit message when we merge this. Might even be fine to merge as is and come back to address - thoughts? Maybe we don't cover every edge case, but should at least cover some.

packages/create-instance/create-component-stubs.js
```diff
@@ -75,6 +81,24 @@ function resolveOptions(component, _Vue) {
...
+  // We filter out the _normalized and $stable key
+  if (
+    ctx &&
+    ctx.$options &&
+    ctx.$options.parent._vnode &&
+    ctx.$options.parent._vnode.data &&
+    ctx.$options.parent._vnode.data.scopedSlots
+  ) {
+    const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
+    return keys(slotKeys).filter(x => x[0] !== '_' && !x.includes('$'))
```
afontcu: Thinking out loud here, and might be missing something, but:

If we're aiming to remove `_normalized` and `$stable`,  could we be more explicit about our intention?

```js
return keys(slotKeys).filter(x => x !== '_normalized' && x !== '$stable')
```

(or even `['_normalized', '$stable'].includes(x)`)

lmiller1990: Yep, sounds good. I don't think there should be any other special variables beginning with `_` or `$`.

afontcu (approved):

lmiller1990:
@JessicaSachs I tried a component like this:

```vue
<template>
  <slot name="blah" />
</template>
```

To see what would happen to `parent`, since `parent` should be `undefined`. Turns out you cannot do this, Vue will throw an error:

```
(Emitted value instead of an instance of Error)
    Error compiling template:

    <slot name="newSyntax" />

    - Cannot use <slot> as component root element because it may contain multiple nodes.
```

I added a check for good measure, just in case.

JessicaSachs:
@lmiller1990 yeah you can’t make slots the root element when you use template functions, but you can with createElement and a render function.

lmiller1990:
Hm, maybe I am missing something - does it make sense for the root element (eg `App.vue` most of the time) to have a `<slot />` in the first place? That would imply you are passing something from above the root to be rendered in the slot.

I couldn't figure out how to write a render function with slots - if you can write a basic example I would be happy to add a test and see what happens.

Thanks for the thorough review... slots feel like the most edge-casey of anything in VTU. We definitely need to get this right.

JessicaSachs (approved):
I think you're right. This looks good.

I spent a few hours trying to break this using render functions on root vue components and other use cases and couldn't get it to break. I do a bunch of edgecase-y things in my prod app and this seems to handle all of it.

dobromir-hristov (approved):

afontcu:
Great job mate! 🙌

lmiller1990:
Going to add one more test around `v-slot` then merge this up, should be done by the end of the week

PR #1387, opened 2020-01-02
fix: wait for the next tick

suits-at:
according to https://vue-test-utils.vuejs.org/guides/#using-nexttick we have to wait for the next tick for this test to pass

### DIFF
--- a/docs/guides/testing-single-file-components-with-mocha-webpack.md
+++ b/docs/guides/testing-single-file-components-with-mocha-webpack.md
@@ -148,7 +148,7 @@ Create a file in `src` named `Counter.vue`:
 And create a test file named `test/Counter.spec.js` with the following code:
 
 ```js
-import Vue from 'vue';
+import Vue from 'vue'
 import { shallowMount } from '@vue/test-utils'
 import Counter from '../src/Counter.vue'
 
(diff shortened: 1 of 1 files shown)


docs/guides/testing-single-file-components-with-mocha-webpack.md
```diff
@@ -148,13 +148,15 @@ Create a file in `src` named `Counter.vue`:
 And create a test file named `test/Counter.spec.js` with the following code:
 
 ```js
+import Vue from 'vue';
```
JessicaSachs: Thank you for the PR! Prettier is complaining about this semicolon. `yarn run format` will fix it.

JessicaSachs (changes requested):
Thank you for finding and fixing this! There is a small formatting change needed before merging. You can exec prettier via `yarn run format`.

We might run the linter as a githook to avoid this problem in the future.

dobromir-hristov:
I can add prettier to run auto on my commit lint branch.

docs/guides/testing-single-file-components-with-mocha-webpack.md
dobromir-hristov: You can also use `wrapper.vm.$nextTick()`, but this could be a bit more future proof, as nextTick will no longer be avail on the vm in V3...

lmiller1990 (commented):
LGTM once prettier is run

PR #1388, opened 2020-01-03
chore: remove gh-pages

JessicaSachs:
This PR removes GH pages, which are unneeded because we deploy docs via Netlify on any merge to the dev branch.

After merging this PR, we can delete the `gh-pages` branch.

One thing I'm a bit uncertain of is if `docs/assets/CNAME` and `docs/assets/circle.yml` are used in some other way. The code doesn't have any references after deleting `scripts/update-docs.sh`, but I'm still paranoid.

Thoughts?


dobromir-hristov:
The CNAME is for GHpages to allow domain binding. Netlify should not need it.
As for circleCI config, once we remove the branch, it should be fine.

dobromir-hristov (approved):

lmiller1990:
PRs that reduce total amount of code - my favorite kind.

lmiller1990 (approved):

PR #1390, opened 2020-01-03
Remove npm references from Contributing

### LINKED ISSUE #1235: Cannot run repo locally without yarn - BLOCKING CONTRIBUTIONS
### Version
`dev` branch

### Reproduction link
[https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md](https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md)

### Steps to reproduce
On a computer with the latest Node and npm versions installed, and **without** Yarn installed, attempt to run the repo locally following the Contributor Guideline instructions.

### What is expected?
You should be able to run the repo locally without having to globally install any non-standard tools.

Crashes looking for yarn. If you change the lerna json to not require Yarn, it will stop crashing on install, but cannot run unit tests because of missing dependencies.

---
The link to Lerna's website in the docs is wrong.

afontcu:
Closes #1235 (sorta...)

Right now, `npm` is not a valid option as the project won't run properly. Only `yarn` is. Thus, I updated the docs accordingly.

In an ideal world, running VTU scripts with `npm` would "just work", but I'm not experienced enough with Lerna to identify what's missing. That's why I thought that, at least, we should make the docs reflect the current state of the art.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -22,7 +22,7 @@ Hi! I’m really excited that you are interested in contributing to Vue Test Uti
 
 - It's OK to have multiple small commits as you work on the PR - we will let GitHub automatically squash it before merging.
 
-- Make sure `npm test` passes. (see [development setup](#development-setup))
+- Make sure `yarn test` passes. (see [development setup](#development-setup))
 
 - If adding new feature:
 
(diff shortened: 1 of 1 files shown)


lmiller1990 (approved):
Sounds good.

lmiller1990 (commented):
Seems will still reference npm in various places, including release scripts and package.json. We might want to move all those to `yarn` as well. What do you think?

afontcu:
> Seems will still reference npm in various places, including release scripts and package.json. We might want to move all those to `yarn` as well. What do you think?

Hm, I didn't want to change the release scripts as they are working fine (AFAIK). Same thing with package.json scripts as long as they are run using `yarn`. Should we update them anyway?

lmiller1990:
Your call - I think it makes sense to pick and stick w/ a single package manager. It's basically as easy as `s/npm run/yarn/g`, but we can do it in a separate PR if you are more comfortable with that too.

lmiller1990:
Since moving the scripts from `npm` to `yarn` is (maybe) a more significant change, let's chat about it and do that in another PR. This is fine for now, nice job.

PR #1392, opened 2020-01-06
Override component watcher with mounting options

### LINKED ISSUE #1391: `watch` mounting option should overwrite option on same component
### Version
1.0.0-beta.30

### Reproduction link
[https://i-included-a-repro.com](https://i-included-a-repro.com)

### Steps to reproduce
repro

```js
import { mount, shallowMount } from '@vue/test-utils'

const Comp = {
  template: '<div />',
  props: ['myProp'],
  watch: {
    myProp: {
      handler(val, oldVal) {
        console.log('Original watcher triggered!')
      },
      deep: true
    }
  }
}

test('Both watchers trigger', (done) => {
  const wrapper = mount(Comp, {
    propsData: {
      myProp: 'foo'
    },
    watch: {
      myProp: {
        handler(val, oldVal) {
          console.log('Overridden watcher triggered!')
        },
        deep: true
      }
    }
  })

  wrapper.setProps({ myProp: 'ok' })

  setTimeout(() => {
    done()
  }, 100)
})
```

### What is expected?
Only the mounting options watcher should do anything. 

Both watchers are triggered.

---
This is because of this line [here](https://github.com/vuejs/vue-test-utils/blob/e7b8d505d51d5987037646454d01d0076ef4d212/packages/create-instance/create-instance.js#L76) which uses Vue's watcher merge strategy, which is to create an array of watchers. That code is in Vue core, found [here](https://github.com/vuejs/vue-test-utils/blob/e7b8d505d51d5987037646454d01d0076ef4d212/packages/create-instance/create-instance.js#L76).

lmiller1990:
resolves #1391 . We should override a component's watcher if one is provided in the mounting options. The issue explains the bug a bit better and provides an example.

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -71,12 +71,6 @@ export default function createInstance(
   // used to identify extended component using constructor
   componentOptions.$_vueTestUtils_original = component
 
-  // make sure all extends are based on this instance
-
-  if (instanceOptions.watch) {
-    console.log(instanceOptions.watch)
-  }
-
   // watchers provided in mounting options should override preexisting ones
   if (componentOptions.watch && instanceOptions.watch) {
     const componentWatchers = Object.keys(componentOptions.watch)
@@ -91,6 +85,7 @@ export default function createInstance(
     }
   }
 
+  // make sure all extends are based on this instance
   const Constructor = _Vue.extend(componentOptions).extend(instanceOptions)
   componentOptions._Ctor = {}
   Constructor.options._base = _Vue
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -73,6 +73,24 @@ export default function createInstance(
 
   // make sure all extends are based on this instance
 
+  if (instanceOptions.watch) {
+    console.log(instanceOptions.watch)
+  }
+
+  // watchers provided in mounting options should override preexisting ones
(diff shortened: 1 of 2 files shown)


dobromir-hristov (approved):

JessicaSachs (commented):
I thought there might be alternate data structures for `watch`... Based on the examples [here](https://vuejs.org/v2/api/#watch), we need to support: Strings, Functions, Arrays, Objects.

Does `includes` work as desired with all of those data types?

dobromir-hristov:
Jess, the key is always a string :)

afontcu (approved):
LGTM!

Quick question: what would happen if the `instance` watcher is immediately invoked but the `component` one isn't (or the other way around)? Would this logic override the value from the `instance` watcher no matter what?

dobromir-hristov:
The one you pass to the mount should always override the entire watcher definition, imho. Its up to you, as a dev to take care if its immediate, deep or whatever, no?

JessicaSachs (approved):

afontcu:
> The one you pass to the mount should always override the entire watcher definition, imho. Its up to you, as a dev to take care if its immediate, deep or whatever, no?

Sure! I guess what I wanted to confirm is that the current implementation overrides the entire definion, not only the `handler` :) I'd assume so, but wanted to make sure 👍

PR #1396, opened 2020-01-07
Allow testing src directly

dobromir-hristov:
# Description
This PR allows testing the `src` files directly, instead of the old way, by building the package first.
The main advantages are speed and debugging. In-code breakpoints now should work properly.
`Watch` mode also works very fast. 

The main changes were:
1. Remove the `~vue/test-utils` alias, as it was causing more problems than it was solving.
2. Swap inside webpack the dist of both test-utils and server-utils, with their `src`, only when in dev testing mode.
3. Allow the old tests against the build to be run. This ensures the end user gets a fully working suite.
4. Change both `test-utils` and `server-test-utils` exports from a default exported object to a named export. This should not affect usage after building the package, as we are using cjs format in the end. Adjust where necessary.
5. Rename all references to the old aliases.
6. Add an extra script command to allow testing only in dev mode. This should not change in any way the existing flow.

# How to run

If you want to run the tests against the source:
```bash
yarn test:unit:only:dev
```

Thanks to Bogdan for helping  debug the initial Webpack config.


package.json
```diff
@@ -22,6 +22,7 @@
     "test:compat": "scripts/test-compat.sh",
     "test:unit": "npm run build:test && npm run test:unit:only",
     "test:unit:only": "mocha-webpack --webpack-config test/setup/webpack.test.config.js test/specs --recursive --require test/setup/mocha.setup.js",
+    "test:unit:only:dev": "cross-env TARGET=dev yarn run test:unit:only",
```
lmiller1990: The rest of the scripts use `npm` here, maybe stick with that (or change them all to `yarn`)?

dobromir-hristov: You are right. Want me to change them to `yarn` or we stick with npm?

lmiller1990 (commented):
I'll pull this down and try it out hopefully in the next day or two, but everything looks like it makes sense... awesome job, I'm really excited to use the watcher for the tests.

Just to confirm my suspicion, when the Karma tests run it still runs the "old" way which is to bundle the entire project and load the `dist` file?

dobromir-hristov:
Yep. We only run against the `src` if you run that command :)
I have tested both workflows and all pass + ci also passes.

lmiller1990:
Great... if it all works the same, sounds good to me. Just left one comment.

lmiller1990 (approved):

package.json
lmiller1990: Hm. I think it's safe to move everything to `yarn`. Up to you, from my point of view seems fine to do it here.

test/setup/webpack.test.config.js
```diff
@@ -15,19 +15,35 @@ const rules = [].concat(
   {
     test: /\.js$/,
     loader: 'babel-loader',
-    exclude: /node_modules\/(?!(shared|create-instance)\/).*/
+    exclude: /node_modules/
   }
 )
-
+const externals = nodeExternals({
+  // we need to whitelist both `create-instance` and files in `shared` package. Otherwise Webpack wont bundle them in the test dev env
```
afontcu:
```suggestion
  // we need to whitelist both `create-instance` and files in `shared` package. Otherwise webpack won't bundle them in the test dev env
```

afontcu (approved):
Looks great! Just added a minor nitpick suggestion.

👍

JessicaSachs (approved):

PR #1398, opened 2020-01-09
Allow find to work on both Pascal case and camel case 

### LINKED ISSUE #1232: vue-test-utils can not resolve PascalCase names of components
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/bootstrap-vue/bootstrap-vue/issues/3270](https://github.com/bootstrap-vue/bootstrap-vue/issues/3270)

### Steps to reproduce
Access a bootstrap-vue component by name: wrapper.find({ name: 'bFormRadioGroup' }).
bootstrap-vue version v2.0.0-rc.19

### What is expected?
wrapper of the element should be returned

empty wrapper is returned

[vue-test-utils]: find did not return Component, cannot call props() on empty Wrapper

---
Per the Vue style guide, since version 2.0.0-rc.12 bootstrap-vue team have named all components using PascalCase instead of camelCase

lmiller1990:
resolves #1232

I think it's fine to support finding a component named `TestComponent` with `test-component` and vice versa, or event `testComponent` - especially for third party packages, it can be hard to know the `name` of a component. Just makes `find` a bit more forgiving and improves the DX with no real downside.

I forked from https://github.com/vuejs/vue-test-utils/pull/1396, will retarget once that is merged. I love the new workflow (run tests against `src`).

### DIFF
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -6,12 +6,12 @@ import {
 import { isConstructor } from 'shared/validators'
 
 function vmMatchesName(vm, name) {
-  const normalize = (name = '') => name.replace(/-/gi, '').toLowerCase()
-  const normalizedName = normalize(name)
+  const lc = (name = '') => name.toLowerCase()
+  const lowerCaseName = lc(name)
   return (
     !!name &&
-    (normalize(vm.name) === normalizedName ||
-      (vm.$options && normalize(vm.$options.name) === normalizedName))
+    (lc(vm.name) === lowerCaseName ||
+      (vm.$options && lc(vm.$options.name) === lowerCaseName))
   )
 }
 
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -4,15 +4,19 @@ import {
   FUNCTIONAL_OPTIONS
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
-import { capitalize } from 'shared/util'
+import { capitalize, camelize } from 'shared/util'
 
 function vmMatchesName(vm, name) {
+  console.log(name)
+  console.log(vm.$options.name)
   return (
     !!name && (
-      vm.name === name || 
+      vm.name === name ||
       (vm.$options && vm.$options.name === name) ||
       vm.name === capitalize(name) ||
-      vm.$options && vm.$options.name === capitalize(name)
+      vm.$options && vm.$options.name === capitalize(name) ||
+      vm.$options && vm.$options.name && vm.$options.name === capitalize(camelize(name)) ||
+      vm.$options && vm.$options.name && capitalize(camelize(vm.$options.name)) === name
     )
   )
 }
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -7,14 +7,19 @@ import { isConstructor } from 'shared/validators'
 import { capitalize, camelize } from 'shared/util'
 
 function vmMatchesName(vm, name) {
-  const componentName = vm.$options && vm.$options.name || ''
+  // We want to mirror how Vue resolves component names in SFCs:
+  // For example, <test-component />, <TestComponent /> and `<testComponent />
+  // all resolve to the same component
+  const componentName = (vm.$options && vm.$options.name) || ''
   return (
-    !!name && (
-      componentName === name || 
+    !!name &&
+    (componentName === name ||
+      // testComponent -> TestComponent
       componentName === capitalize(name) ||
+      // testcomponent -> TestComponent
       componentName === capitalize(camelize(name)) ||
-      capitalize(camelize(componentName)) === name
-    )
+      // same match as above, but the component name vs query
+      capitalize(camelize(componentName)) === name)
   )
 }
 
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -16,7 +16,7 @@ function vmMatchesName(vm, name) {
     (componentName === name ||
       // testComponent -> TestComponent
       componentName === capitalize(name) ||
-      // testcomponent -> TestComponent
+      // test-component -> TestComponent
       componentName === capitalize(camelize(name)) ||
       // same match as above, but the component name vs query
       capitalize(camelize(componentName)) === name)
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -4,14 +4,16 @@ import {
   FUNCTIONAL_OPTIONS
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
+import { capitalize } from 'shared/util'
 
 function vmMatchesName(vm, name) {
-  const lc = (name = '') => name.toLowerCase()
-  const lowerCaseName = lc(name)
   return (
-    !!name &&
-    (lc(vm.name) === lowerCaseName ||
-      (vm.$options && lc(vm.$options.name) === lowerCaseName))
+    !!name && (
+      vm.name === name || 
+      (vm.$options && vm.$options.name === name) ||
+      vm.name === capitalize(name) ||
+      vm.$options && vm.$options.name === capitalize(name)
+    )
   )
 }
 
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -6,8 +6,12 @@ import {
 import { isConstructor } from 'shared/validators'
 
 export function vmMatchesName(vm, name) {
+  const normalize = (name = '') => name.replace(/-/gi, '').toLowerCase()
+  const normalizedName = normalize(name)
   return (
-    !!name && (vm.name === name || (vm.$options && vm.$options.name === name))
+    !!name &&
+    (normalize(vm.name) === normalizedName ||
+      (vm.$options && normalize(vm.$options.name) === normalizedName))
   )
 }
 
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -5,7 +5,7 @@ import {
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
 
-export function vmMatchesName(vm, name) {
+function vmMatchesName(vm, name) {
   const normalize = (name = '') => name.replace(/-/gi, '').toLowerCase()
   const normalizedName = normalize(name)
   return (
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -7,16 +7,13 @@ import { isConstructor } from 'shared/validators'
 import { capitalize, camelize } from 'shared/util'
 
 function vmMatchesName(vm, name) {
-  console.log(name)
-  console.log(vm.$options.name)
+  const componentName = vm.$options && vm.$options.name || ''
   return (
     !!name && (
-      vm.name === name ||
-      (vm.$options && vm.$options.name === name) ||
-      vm.name === capitalize(name) ||
-      vm.$options && vm.$options.name === capitalize(name) ||
-      vm.$options && vm.$options.name && vm.$options.name === capitalize(camelize(name)) ||
-      vm.$options && vm.$options.name && capitalize(camelize(vm.$options.name)) === name
+      componentName === name || 
+      componentName === capitalize(name) ||
+      componentName === capitalize(camelize(name)) ||
+      capitalize(camelize(componentName)) === name
     )
   )
 }
--- a/test/specs/wrapper/find.spec.js
+++ b/test/specs/wrapper/find.spec.js
@@ -431,6 +431,20 @@ describeWithShallowAndMount('find', mountingMethod => {
     )
   })
 
+  it('returns a Wrapper matching a component pascal case name in options object', () => {
+    const wrapper = mountingMethod(ComponentWithChild)
+    expect(wrapper.find({ name: 'TestComponent' }).name()).to.equal(
+      'test-component'
+    )
+  })
+
+  it('returns a Wrapper matching a component camel case name in options object', () => {
+    const wrapper = mountingMethod(ComponentWithChild)
+    expect(wrapper.find({ name: 'testComponent' }).name()).to.equal(
+      'test-component'
+    )
+  })
+
   it('returns Wrapper of Vue Component matching the ref in options object', () => {
     const wrapper = mountingMethod(ComponentWithChild)
     expect(wrapper.find({ ref: 'child' }).isVueInstance()).to.equal(true)
--- a/test/specs/wrapper/find.spec.js
+++ b/test/specs/wrapper/find.spec.js
@@ -468,7 +468,7 @@ describeWithShallowAndMount('find', mountingMethod => {
     const wrapper = mountingMethod(compiled)
     const a = wrapper.find('a')
     const message =
-      '[vue-test-utils]: $ref selectors can used on Vue component wrappers'
+      '[vue-test-utils]: $ref selectors can only be used on Vue component wrappers'
     const fn = () => a.find({ ref: 'foo' })
     expect(fn)
       .to.throw()
--- a/test/specs/wrapper/find.spec.js
+++ b/test/specs/wrapper/find.spec.js
@@ -431,14 +431,7 @@ describeWithShallowAndMount('find', mountingMethod => {
     )
   })
 
-  it('returns a Wrapper matching a component camel case name in options object', () => {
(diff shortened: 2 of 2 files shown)


packages/test-utils/src/matches.js
```diff
@@ -6,8 +6,12 @@ import {
 import { isConstructor } from 'shared/validators'
 
 export function vmMatchesName(vm, name) {
+  const normalize = (name = '') => name.replace(/-/gi, '').toLowerCase()
```
afontcu: Wondering if this should be a globally-available utility function..?

dobromir-hristov: Wont this match `TheNavigation` and `the-navigation` with `thenavigation` ?

lmiller1990: Hm, I can't see anywhere else we do something similar to this. Does anywhere come to mind?
Also just noticed this vmMatchesName does not need to be exported, going to remove that.

lmiller1990: @dobromir-hristov It will. Will this be a problem in practice? I guess the edge case is someone has two components, one is `TheNavigation` and one is `The-Navigation`. What do you think? I don't think that's too likely.

dobromir-hristov: `TheNavigation` and `The-Navigation` I think are the same for Vue, `thenavigation` is not I think.... We can give it a test.

But you are right, its kind of a niche case.

lmiller1990: Ok I agree - we should match whatever Vue does.

lmiller1990: @dobromir-hristov all the same in SFC: https://codesandbox.io/s/festive-star-dbr9n, in App.vue. I think we need to remove the `-` anyway if we want to allow `FooBar` to match against `foo-bar` (which we should, since from Vue's point of view either or is fine)

dobromir-hristov: `FooBar` is not `foobar`, according to your example. But am not sure if its a problem rly..

dobromir-hristov: We can also see how `vue-devtools` does its filtering?

lbogdan: For reference, this is how Vue does name resolving: https://github.com/vuejs/vue/blob/b07087deae060244fe5d9651b38dc5f1e79667a9/src/core/util/options.js#L454-L460 .

lbogdan:
For reference, this is how Vue does component name resolving: https://github.com/vuejs/vue/blob/b07087deae060244fe5d9651b38dc5f1e79667a9/src/core/util/options.js#L454-L460 .

packages/test-utils/src/matches.js
lbogdan: For reference, this is how Vue does component name resolving: https://github.com/vuejs/vue/blob/b07087deae060244fe5d9651b38dc5f1e79667a9/src/core/util/options.js#L454-L460 .

lmiller1990: Hm. In that case I guess `FooBar` should not match `foobar`. Although not technically correct, for code simplicity I personally don't mind - the goal is to let people easily `find` a component and write tests, and I think the edge case is unlikely enough to occur.

On the flip side, I see an argument for matching Vue as closely as possible, where `FooBar` is not `foobar`. Although in the SFC example linked above, it does not seem to make a difference.

Happy to go either decision. Anyone want to chime in? For me, DX and ease of use are the goal.

lbogdan: > Although in the SFC example linked above, it does not seem to make a difference.

I think it did? `<thenavigation>` doesn't match either `TheNavigation` or `The-Navigation` components.

lmiller1990: Oh, my bad. So to sum up, `find({ name: 'fooBar' })` should match:

- `name: 'FooBar'`
- `name: 'foobar`
- `name: 'fooBar'`

Basically we don't care about the case. But the `-`. is relevant. So, don't match `name: foo-bar`. I will make the change this evening! That should solve the original issue the OP pointed out, while keeping to way Vue matches names.

lbogdan: I've just tested, and `<fooBar>` only matches `FooBar` and `fooBar`, but not `foobar`.

So I'd say don't normalize at all and just check the three variations Vue checks.

lmiller1990: Yep, agreed, thanks!

JessicaSachs: So we'll wait for this change and then re-review? @lmiller1990

lmiller1990: yes I will ping you guys when I update it! Within 24h.

packages/test-utils/src/matches.js
```diff
@@ -5,9 +5,13 @@ import {
...
 
-export function vmMatchesName(vm, name) {
+function vmMatchesName(vm, name) {
+  const lc = (name = '') => name.toLowerCase()
+  const lowerCaseName = lc(name)
   return (
-    !!name && (vm.name === name || (vm.$options && vm.$options.name === name))
+    !!name &&
+    (lc(vm.name) === lowerCaseName ||
+      (vm.$options && lc(vm.$options.name) === lowerCaseName))
```
lbogdan: I still think we should mirror Vue behavior's here. This will match e.g. `<fooBar>` with `foobar`, `<foobar>` with `FooBar`, although it shouldn't.

packages/test-utils/src/matches.js
lmiller1990: ok, should be good now @JessicaSachs @dobromir-hristov @afontcu

packages/test-utils/src/matches.js
lmiller1990: Hm.. the original issue was that it `find` should not care about casing. I personally think it's fine to disregard casing for making a better DX.

If we decide to match Vue's system, I guess we just close this issue and PR? I don't have a super strong opinion, but I do like making the developer exp more simple.

lbogdan: The original issue was that `<bFormRadioGroup>` doesn't match `BFormRadioGroup`, which Vue does.

lmiller1990: No it's not, it has nothing to do with the tag in the markup. The issue is finding the component by the `name` field. He was doing `find({ name: 'bFormRadioGroup' })` which was working fine. Then the developer of the Bootstrap lib changed the `name` field to `BFormRadioGroup`, and his test broke, which I think isn't great since the actual behavior of the component was still fine. 

I think that the casing of the name field (and the hyphens too) should not be something the developer needs to think about when using `find`, especially when often the `name` tag is in a thid party library and it's not that easy to see what it is without going to the source code to have a look.

lbogdan: OK, last comment, as I don't want to become annoying 🙂: IMO ~`find[All]('name')`~ `find[All]({ name: 'name' })` should match exactly what Vue does when trying to find component `<name>` in a template - which is equivalent, I think, to calling `Vue.component('name')`. Case at hand, ~`find('bFormRadioGroup')`~ `find({ name: 'bFormRadioGroup'})` should find `BFormRadioGroup`, because Vue resolves `<bFormRadioGroup>` to `BFormRadioGroup`.

lmiller1990: I think discussion is good 👍 not annoying, I don't take it personally. 

I think I'm missing something obvious 🤔 . Did you mean to type `find({ name: 'bFormRadioGroup' })`? We don't have any support for `find` for custom tags, because there are some edge cases related to functional component we could not solve in a PR a while ago. So using `find(<string literal>)` only works on regular HTML elements. That's why we have the `name` option as well.
______
Are we in agreement that `find({ name: 'bFormRadioGroup' })` should find a component which has the `name` field set to `BFormRadioGroup`? If so, are you saying that I should update the code to only lowercase the first character? Again, I think I missed the point you were making, you definitely know how Vue resolves component names better than I do.

Opinions aside, let's get this to a point that improves the `find` function (if that means closing this PR and issue because changing the current functionality of `find` is correct, that is fine with me, too). My main goal at the moment is to either fix or close open issues, so only ones that are really problems are remaining.

lbogdan: Yes, sorry, I meant `find({ name: 'name' })`, indeed. Updated the comment.

lbogdan: I shared the link to how Vue resolves component names a few comments back, here it is again for convenience: https://github.com/vuejs/vue/blob/b07087deae060244fe5d9651b38dc5f1e79667a9/src/core/util/options.js#L454-L460 . It basically searches for `'name'`, `camelize('name')` and `capitalize(camelize('name'))`, is this order.

Case at hand
- `'bFormRadioGroup'` - not found
- `camelize('bFormRadioGroup') === 'bFormRadioGroup'` - not found
- `capitalize(camelize('bFormRadioGroup')) === 'BFormRadioGroup'` - **found**.

lmiller1990: Ok great, so what do you think is the best thing to do in this case? I'm pretty confused. Do we just lowercase the first character? The current PR lowercases all the characters. I thought Vue would resolve regardless of casing, I could be wrong on that one, but that's why I did `lc` on all the names.

lmiller1990: Ok, I get it now. I must have missed the fact it does `capitalize` at the end. I agree, let's just match this. Thanks for walking me through this. 👍

lmiller1990: done!

packages/test-utils/src/matches.js
```diff
@@ -4,10 +4,20 @@ import {
   FUNCTIONAL_OPTIONS
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
+import { capitalize, camelize } from 'shared/util'
 
-export function vmMatchesName(vm, name) {
+function vmMatchesName(vm, name) {
+  console.log(name)
```
afontcu: some debugging stuff over here! ;)

test/specs/wrapper/find.spec.js
```diff
@@ -441,7 +468,7 @@ describeWithShallowAndMount('find', mountingMethod => {
     const wrapper = mountingMethod(compiled)
     const a = wrapper.find('a')
     const message =
-      '[vue-test-utils]: $ref selectors can only be used on Vue component wrappers'
+      '[vue-test-utils]: $ref selectors can used on Vue component wrappers'
```
afontcu: Is this an expected modification?

packages/test-utils/src/matches.js
```diff
@@ -4,10 +4,20 @@ import {
...
   return (
-    !!name && (vm.name === name || (vm.$options && vm.$options.name === name))
+    !!name && (
+      vm.name === name ||
+      (vm.$options && vm.$options.name === name) ||
+      vm.name === capitalize(name) ||
+      vm.$options && vm.$options.name === capitalize(name) ||
+      vm.$options && vm.$options.name && vm.$options.name === capitalize(camelize(name)) ||
+      vm.$options && vm.$options.name && capitalize(camelize(vm.$options.name)) === name
+    )
```
dobromir-hristov: I have a proposal, lets save the vm.$options.name to a variable, something like 
```js
const componentName = vm.$options && vm.$options.name || ''
```
then we can easily reduce those those lines. Yes it will use a bit more memory, but for readability sake I think it makes sense?
Also that line 18 has an extra check others dont.
Unrelated:
You know whats the diff between `vm.name` and `vm.$options.name`? One is the one you set in its options and the other is the one that it is registered by?

packages/test-utils/src/matches.js
JessicaSachs: Do we want a lint rule for no console log? We do this at work. Same for debugger statements

packages/test-utils/src/matches.js
JessicaSachs: Can we add comments for each of the conditionals? Eg what use case each line is handling? A 7 line if statement is hard to grok easily.

JessicaSachs (changes requested):
Added comments. They’re mostly minor nits.

packages/test-utils/src/matches.js
dobromir-hristov: I think its OK to have one yeah :) With the new linting on commit, it will warn you.

test/specs/wrapper/find.spec.js
lmiller1990: I dont remember making this one, but it does improve the error message 🤔

packages/test-utils/src/matches.js
lmiller1990: Yep agreed, I'll make these changes

test/specs/wrapper/find.spec.js
lbogdan: Should probably read "can **be** used on" (missing "be"), though.

packages/test-utils/src/matches.js
lmiller1990: @dobromir-hristov I am not sure on the difference between `vm.name` and `vm.$options.name`. `vm.name` is undefined in every test... I removed it and everything is still passing. If it serves no purpose, we may as well omit that check. What do you think?

We should check the source code for Vue and see if we can figure out the difference.

Regarding your comment about the extra variable - I agree, that is more readable. I updated it the code. 

@JessicaSachs I added some comments explaining why we do each of the matches. What do you think? If you think you can make the comment clearer, you can add some commits or I can add some if you leave an updated comment here.

test/specs/wrapper/find.spec.js
lmiller1990: done, updated. (comment still show old version, but if you view "files changed" is it new updated one)

packages/test-utils/src/matches.js
dobromir-hristov: I also gave a quick search and did not see `vm.name` defined. Its not in the Vue types as well. 
Wonder if its something from "old days"? 
New code is much more readable, gj

dobromir-hristov (approved):

packages/test-utils/src/matches.js
```diff
@@ -4,10 +4,22 @@ import {
   FUNCTIONAL_OPTIONS
 } from 'shared/consts'
 import { isConstructor } from 'shared/validators'
+import { capitalize, camelize } from 'shared/util'
 
-export function vmMatchesName(vm, name) {
+function vmMatchesName(vm, name) {
+  // We want to mirror how Vue resolves component names in SFCs:
+  // For example, <test-component />, <TestComponent /> and `<testComponent />
```
afontcu, line 11:
```suggestion
  // For example, <test-component />, <TestComponent /> and <testComponent />
```

test/specs/wrapper/find.spec.js
```diff
@@ -431,6 +431,33 @@ describeWithShallowAndMount('find', mountingMethod => {
...
+  it('returns a Wrapper matching a kebab-case name option and a Pascal Case component name ', () => {
+    const component = {
+      name: 'CamelCase',
+      render: h => h('div')
+    }
+    const wrapper = mountingMethod(component)
+    expect(wrapper.find({ name: 'camel-case' }).name()).to.equal('CamelCase')
+  })
+
+  it('returns a Wrapper matching a Pascal Case name option and a kebab-casecomponent name ', () => {
```
afontcu, line 452:
```suggestion
  it('returns a Wrapper matching a PascalCase name option and a kebab-case component name', () => {
```

test/specs/wrapper/find.spec.js
```diff
@@ -431,6 +431,33 @@ describeWithShallowAndMount('find', mountingMethod => {
...
+  it('returns a Wrapper matching a camelCase name option and a Pascal Case component name ', () => {
+    const component = {
+      name: 'CamelCase',
+      render: h => h('div')
+    }
+    const wrapper = mountingMethod(component)
+    expect(wrapper.find({ name: 'camelCase' }).name()).to.equal('CamelCase')
+  })
+
+  it('returns a Wrapper matching a kebab-case name option and a Pascal Case component name ', () => {
```
afontcu, line 443:
```suggestion
  it('returns a Wrapper matching a kebab-case name option and a PascalCase component name', () => {
```

test/specs/wrapper/find.spec.js
```diff
@@ -431,6 +431,33 @@ describeWithShallowAndMount('find', mountingMethod => {
     )
   })
 
+  it('returns a Wrapper matching a camelCase name option and a Pascal Case component name ', () => {
```
afontcu, line 434:
```suggestion
  it('returns a Wrapper matching a camelCase name option and a PascalCase component name', () => {
```

afontcu (approved):

PR #1402, opened 2020-01-10
chore: fix missing rollup dependency

JessicaSachs:
Couldn't build the project with `yarn build` unless I added this dependency. Can someone validate that this is also happening for them?


dobromir-hristov:
I am building perfectly fine?

JessicaSachs:
Are you really? Do you have the dependency that I added?

JessicaSachs:
Ah, the dep is in the root and my environment was just messed up

Issue #1403, opened 2020-01-11
emitted() does not work with trigger in beta 29 , works with 28


samgre:
### Version
1.0.0-beta.30

### Reproduction link
[https://github.com/samgre/vue-test-utils-bug/](https://github.com/samgre/vue-test-utils-bug/)

### Steps to reproduce
To reproduce the issue, clone the repository and follow these steps:

1. Run `npm i && npm test` to run the test
2. The test fails.
3. Fix version of @vue/test-utils to older one, 1.0.0-beta.28
4. Re-run the test `npm i && npm test`
5. The test passes.

### What is expected?
The test should pass also on 1.0.0-beta.30. At least I did not see anything related to this in the change log.

### What is actually happening?
The test that used to pass, now fails.

---
I upgraded from beta 28 to beta 30.

dobromir-hristov:
Hey there. As of v30 the tests are no longer synchronous, so after each event trigger, if you want to assert if the dom got updated, you need to wait for the `$nextTick()`.
`await wrapper.vm.$nextTick()` should do the job, just after the trigger.

samgre:
That's it, thanks!

The issue is documented under section Common tips, but not in the api reference [1], [2]. It would be a nice improvement, if the code example in [1] could highlight this.

You may close this issue or make it a suggestion to improve docs.

[1] https://vue-test-utils.vuejs.org/api/wrapper/#trigger
[2] https://vue-test-utils.vuejs.org/api/wrapper/#emitted

lmiller1990:
@samgre Good idea. Would you like to make a PR updating the docs? Good chance for a first PR!

Happy to do it if you haven't got the time. We should update all the documentation to reflect the need for `await wrapper.vm.$nextTick`.

samgre:
Sure, would this do? https://github.com/vuejs/vue-test-utils/pull/1405

lmiller1990:
yep, nice job. looks good!

PR #1405, opened 2020-01-13
docs: add note about waiting for async trigger event to be handled

samgre:
Here's a documentation proposed documentation update related to issue #1403

### DIFF
--- a/docs/api/wrapper/emitted.md
+++ b/docs/api/wrapper/emitted.md
@@ -14,6 +14,8 @@ const wrapper = mount(Component)
 wrapper.vm.$emit('foo')
 wrapper.vm.$emit('foo', 123)
 
+await wrapper.vm.$nextTick()  // Wait until $emits have been handled
+
 /*
 wrapper.emitted() returns the following object:
 {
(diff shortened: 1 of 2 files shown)


docs/api/wrapper/emitted.md
```diff
@@ -9,26 +9,30 @@ Return an object containing custom events emitted by the `Wrapper` `vm`.
...
+  wrapper.vm.$emit('foo')
+  wrapper.vm.$emit('foo', 123)
 
-/*
-wrapper.emitted() returns the following object:
-{
-  foo: [[], [123]]
-}
-*/
+  await wrapper.vm.$nextTick() // Wait until $emits have been handled
```
afontcu, line 18: Not sure if it's too down-to-metal, but we're technically waiting for the next DOM update cycle, right? Should be mention that? I mean, `$nextTick` has nothing to do with `$emits` directly, and one could infer that from the comment above.

Just thinking out loud 😇

lmiller1990: What you say is true, although I actually think the update @samgre made is fine. I think the docs should just show you how to use the library, rather than explain the implementation details. As long as people do `await`, it will work. We should likely update all code examples to include this.

If you want to put another commit on top of this making the comment more accurate, that's fine too. Perhaps "await until the next cycle to ensure $emits have been handle" might be appropriate.

lmiller1990 (approved):

PR #1411, opened 2020-01-20
WIP: feat/fix: stub out transitions by default

### LINKED ISSUE #1384: v-if inside transition element not re rendering in beta30
### Version
1.0.0-beta.30

### Reproduction link
[https://github.com/aldarund/vue-test-utils-transition-bug](https://github.com/aldarund/vue-test-utils-transition-bug)

### Steps to reproduce
components/test.test.js 
When changing rendering v-if of element inside transition and do nextTick/flushPromises vue test utils still renders old version.
Tried with flush-promises too, same effect.
Relevant files only following 

components\TransitionTest.vue
components\test.test.js

all other just stub files from empty project create.

### What is expected?
test is passed

test is failing

---
If change version to v29 all works fine. Without wrapping into transition all works too

lmiller1990:
This PR resolves #1384 and probably a bunch of other issues. Since `<transition>` causes problems, and there isn't really anything to unit test, I just stub them out permanently.

Edit: probably need the same thing for transition group. I'll add that.

I built and tested this against the reproduction provided and it passed.

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -55,7 +55,7 @@ export default function createInstance(
   const stubComponentsObject = createStubsFromStubsObject(
     componentOptions.components,
     // $FlowIgnore
-    options.stubs,
+    { ...options.stubs, transition: true },
     _Vue
   )
 
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -55,7 +55,7 @@ export default function createInstance(
   const stubComponentsObject = createStubsFromStubsObject(
     componentOptions.components,
     // $FlowIgnore
-    { ...options.stubs, transition: true },
+    options.stubs,
     _Vue
   )
 
--- a/packages/test-utils/src/config.js
+++ b/packages/test-utils/src/config.js
@@ -1,5 +1,8 @@
 export default {
-  stubs: {},
+  stubs: {
+    transition: true,
+    'transition-group': true
+  },
   mocks: {},
   methods: {},
   provide: {},
--- a/test/resources/components/component-with-transitions.vue
+++ b/test/resources/components/component-with-transitions.vue
@@ -1,0 +1,20 @@
+<template>
+  <div>
+    <transition>
+      <span v-if="!expanded" data-testid="expanded">
+        Content
+      </span>
+    </transition>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'component-with-transitions',
+  data() {
+    return {
+      expanded: false
+    }
+  }
+}
+</script>
--- a/test/specs/config.spec.js
+++ b/test/specs/config.spec.js
@@ -1,6 +1,7 @@
 import { describeWithShallowAndMount } from '~resources/utils'
 import ComponentWithProps from '~resources/components/component-with-props.vue'
 import { config, createLocalVue } from '@vue/test-utils'
+import ComponentWithTransitions from '~resources/components/component-with-transitions.vue'
 
 describeWithShallowAndMount('config', mountingMethod => {
   const sandbox = sinon.createSandbox()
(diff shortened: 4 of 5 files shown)


packages/create-instance/create-instance.js
```diff
@@ -55,7 +55,7 @@ export default function createInstance(
   const stubComponentsObject = createStubsFromStubsObject(
     componentOptions.components,
     // $FlowIgnore
-    options.stubs,
+    { ...options.stubs, transition: true },
```
dobromir-hristov: Shouldnt we add this to the default stubs in options? Wont it be a bit cleaner?

dobromir-hristov:
Do you think ppl will want to have those transition hooks also working? I am not great with transitions, so I cant really say :/

packages/create-instance/create-instance.js
lmiller1990: Yeah this might not be the best place to add this - which line/file do you think we should add this?

dobromir-hristov: Didnt we have a default global `stubs` object?

lmiller1990:
@dobromir-hristov I forgot about the JS transition hooks. I don't think it makes a lot of sense to unit test those. I think the amount of people having problems with the transitions is much greater than those unit testing transition hooks, so it's probably a good trade off.

If someone has a complex transition hook they need to test, we can probably work with them to find an alternative strategy to testing those.

dobromir-hristov:
Yeah you got a point, but lots of Vue UI libs use transitions, and ppl try to test for example if Modals are shown, or whatever, and some of those use the transition hooks. 
I am just saying I am not exactly sure if this will be an easy thing for them to overcome.

lmiller1990:
Good point.

I'll run Vuetify's specs against this branch and see what happens. It looks like they already [stub transitions](https://github.com/vuetifyjs/vuetify/blob/d80e01f0c9ec9c54b066f155e642f0ca06021a17/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.ts#L101).

Vue Bootstrap does use the those hooks, [for example here](https://github.com/bootstrap-vue/bootstrap-vue/blob/c2d3be61ec33ddf27cd41e98153300ef257d0f3b/src/components/modal/modal.js#L602). which will likely break their builds if they upgrade to the latest version. We could allow passing 

```
stubs: {
  transition: false
}
```

for this case. Thoughts @dobromir-hristov ?

packages/create-instance/create-instance.js
lmiller1990: we do - updated the PR, good suggestion

lmiller1990:
Update; I tried to upgrade both Vuetify and Vue Bootstrap's suites to the latest VTU but they failed. Without a lot of work upgrading those suites, there is no way to tell if this is a large, breaking change or not. For now, you can use the old behavior with

```
const wrapper = mount(Comp, {
  stubs: {
   transition: false
  }
})
```

I can update the docs to include this default stub if everyone else is happy with this solution.

nachogarcia:
I'm stuck in v29 because we would have to rewrite our whole test suite to allow transitions etc. (Using bootstrap-vue). Which sucks because we have issues testing components that use code-splitting.

It would be really nice to have a built-in way of testing while using these libraries, although I understand how difficult is to integrate it. Just to express my point of view :P

JessicaSachs (approved):

JessicaSachs:
@nachogarcia I believe this PR will help you with the transitions issue.

Agreed. We’re thinking about how to best serve people who use component frameworks.

lmiller1990:
I started upgrading Vuetify's specs so I could get them on the latest VTU, then bump to this and see if there are any major breaking changes due to this. I'm not sure it's worth it, though:

1. upgrading Vuetify will take quite some time
2. not reflective of all ecosystem (Vue Bootstrap etc)

Need a good way to handle this moving forward. Hm.

lmiller1990:
Alright, it's time to head to v1. This bad boy is going in.

PR #1431, opened 2020-02-06
fix(types): emitted can return undefined

lucaswerkmeister:
I’m not sure how strict this project is about `undefined` types in general, but I found it confusing that the documentation advised to use `expect(…).toBeTruthy()` on a function (`emitted( 'foo' )`) that was typed to always return an array (i. e., always truthy).

It might be better to only add the `|undefined` to the overload with an event name, not to the overload that returns all events at once… not sure.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

~~I suppose in theory it’s a breaking change? People who previously assigned the result of `.emitted()` to a variable with an explicit type might now get a type error. The expected use, inline within `expect()`, should be fine, though.~~

Calls like `.emitted().foo[0]` or `.emitted('foo')[0]` may also need to be updated, depending on TypeScript config – in the tests, I’ve used the non-null assertion operator.

**Other information:**

### DIFF
--- a/packages/test-utils/types/index.d.ts
+++ b/packages/test-utils/types/index.d.ts
@@ -102,8 +102,8 @@ export interface Wrapper<V extends Vue | null> extends BaseWrapper {
   text (): string
   name (): string
 
-  emitted (): { [name: string]: Array<Array<any>> }
-  emitted (event: string): Array<any>
+  emitted (): { [name: string]: Array<Array<any>>|undefined }
+  emitted (event: string): Array<any>|undefined
   emittedByOrder (): Array<{ name: string, args: Array<any> }>
 }
 
--- a/packages/test-utils/types/test/wrapper.ts
+++ b/packages/test-utils/types/test/wrapper.ts
@@ -23,8 +23,8 @@ bool = wrapper.isVueInstance()
 
 wrapper.vm.$emit('hello')
 
-let n: number = wrapper.emitted().hello[0][0]
-let o: string = wrapper.emitted('hello')[0]
+let n: number = wrapper.emitted().hello![0][0]
+let o: string = wrapper.emitted('hello')![0]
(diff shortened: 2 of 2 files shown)


lucaswerkmeister:
After seeing the impact this has on `.emitted().foo[0]` and `.emitted('foo')[0]`, I’m no longer convinced this is worth it… up to y’all, feel free to close if you want.

lmiller1990:
Did not review the codebase yet, but I don't think `emitted `can be `undefined` - it's an empty object, `{}`, by default, and `emitted().foo` is an empty array `[]` by default, isn't it?

Either way, appreciate the effort to making a contribution. What do you think? I can look more into this if there is an `undefined` case (if it can be undefined, the types should reflect this).

lucaswerkmeister:
If `emitted().foo` was an empty array by default, then it would be a bug to use `expect( … ).toBeFalsy()` on it, as the documentation advises, because arrays are always truthy, even empty ones. But as far as I can tell, it starts out as `undefined`.

lucaswerkmeister:
(`emitted()` itself can never be `undefined`, that’s true, but I don’t think this pull request claims that either. `emitted(eventName)` can be `undefined`, though.)

lmiller1990:
Right, I get it now. I think this makes sense. Is this modified test code supposed to be there? I think just having the typings is probably fine. If you can remove that, I'd be happy to merge this.

lucaswerkmeister:
Without the changes to the test code, I got a CI failure because TypeScript complained about values being possibly `undefined`, I think.

lmiller1990:
Ok, I think this makes sense. Thanks!

PR #1441, opened 2020-02-20
stub globally registered components (fix #1272)

### LINKED ISSUE #1272: Some stubbed Vuetify components can't be found with `find` wrapper method
### Version
1.0.0-beta.29

### Reproduction link
[https://github.com/anthonygore/vuetify-test](https://github.com/anthonygore/vuetify-test)

### Steps to reproduce
npm i
npm run test:unit

### What is expected?
All tests should pass

The second test fails as the VTextField component can't be found.

* * * 

I don't know if this is a Vuetify issue or a Vue Test Utils issue, but some stubbed child components can't be found with the `find` wrapper method when the parent is shallow mounted.

It's not just `VTextField`, either, `VSelect` also has this problem, while others like `VIcon` work just fine.

```js
import { shallowMount } from "@vue/test-utils";
import Vuetify from "vuetify";
import { VBtn, VTextField } from "vuetify/lib";

Vue.use(Vuetify);

const component = {
  template: `
    <div>
      <v-btn />
      <v-text-field />
    </div>
  `
};

// Passes
it('should find v-btn', () => {
  const wrapper = shallowMount(component);
  expect(wrapper.find(VBtn).exists()).toBe(true);
});

// Fails
it('should find v-text-field', () => {
  const wrapper = shallowMount(component);
  expect(wrapper.find(VTextField).exists()).toBe(true);
});
```

### Update

I think the issue is something to do with how some global components are registered because this works:

```js
Vue.use(Vuetify);

const component = {
  template: `
    <div>
      <v-btn />
      <v-text-field />
    </div>
  `,
  components: {
    VTextField
  }
};
```

Note that I don't need to locally register `VBtn`, just `VTextField`...

thejcannon:
When specifying the components to stub, merge the globally registered components into the locally registered ones. (fix #1272)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/create-instance/create-instance.js
+++ b/packages/create-instance/create-instance.js
@@ -52,8 +52,14 @@ export default function createInstance(
   // root instance when it's instantiated
   const instanceOptions = extractInstanceOptions(options)
 
+  const globalComponents = _Vue.options.components || {}
+  const componentsToStub = Object.assign(
+    Object.create(globalComponents),
+    componentOptions.components
+  )
+
   const stubComponentsObject = createStubsFromStubsObject(
-    componentOptions.components,
+    componentsToStub,
     // $FlowIgnore
     options.stubs,
     _Vue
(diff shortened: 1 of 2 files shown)


lmiller1990 (commented):
Good catch. Can you remove the` dist` files? We don't commit them to the repository.

Other than that, looks good.

PR #1467, opened 2020-03-12
feat: support lazy modifier with setValue

### LINKED ISSUE #1465: setValue does not apply changes to model when lazy modifier is present
### Version
1.0.0-beta.32

### Reproduction link
[https://github.com/jaredmcateer/vtu-set-value](https://github.com/jaredmcateer/vtu-set-value)

### Steps to reproduce
Clone the minimal test case repo, run `npm install` once that is complete run `node test.js` The assert will fail on line 35 as the model has not updated. Commenting out line 35 will allow the remaining assert to run, which will pass because the change event has been triggered.

### What is expected?
When calling setValue the model should be updated to reflect that value

setValue triggers an "input" event on text input/textareas. A model with the lazy modifier does not update the model until the change event is triggered (see: https://vuejs.org/v2/guide/forms.html#lazy) 

---
If the desire is for setValue to mimic typing perhaps then the documentation should indicate that setValue must be accompanied by a triggering of the change event if the lazy modifier is present or an optional parameter could be passed to setValue to force a change event.

lmiller1990:
Fixes #1465

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -569,10 +569,19 @@ export default class Wrapper implements BaseWrapper {
       tagName === 'TEXTAREA' ||
       tagName === 'SELECT'
     ) {
-      const event = tagName === 'SELECT' ? 'change' : 'input'
       // $FlowIgnore
       this.element.value = value
-      this.trigger(event)
+
+      if (tagName === 'SELECT') {
+        this.trigger('change')
+      } else {
+        this.trigger('input')
+      }
+
+      // for v-model.lazy, we need to trigger a change event, too.
+      if (tagName === 'INPUT' && this.element._vModifiers) {
+        this.trigger('change')
+      }
     } else {
       throwError(`wrapper.setValue() cannot be called on this element`)
     }
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -581,7 +581,7 @@ export default class Wrapper implements BaseWrapper {
       // for v-model.lazy, we need to trigger a change event, too.
       // $FlowIgnore
       if (
-        tagName === 'INPUT' &&
+        (tagName === 'INPUT' || tagName === 'TEXTAREA') &&
         this.element._vModifiers &&
         this.element._vModifiers.lazy
       ) {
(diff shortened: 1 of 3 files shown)


packages/test-utils/src/wrapper.js
```diff
@@ -569,10 +569,19 @@ export default class Wrapper implements BaseWrapper {
...
-      this.trigger(event)
+
+      if (tagName === 'SELECT') {
+        this.trigger('change')
+      } else {
+        this.trigger('input')
+      }
+
+      // for v-model.lazy, we need to trigger a change event, too.
+      if (tagName === 'INPUT' && this.element._vModifiers) {
```
dobromir-hristov: shall we do `vModifiers.lazy` ?

dobromir-hristov: Lazy can be used on textarea too :)

jaredmcateer (commented):
Other than @dobromir-hristov's review points this is a satisfactory solution for me. Thanks for the quick turnaround!

packages/test-utils/src/wrapper.js
lmiller1990: Oops, I thought that was what I was doing. Yes, we definitely should, I'll fix this.

lmiller1990: Fixed

packages/test-utils/src/wrapper.js
```diff
@@ -569,10 +569,24 @@ export default class Wrapper implements BaseWrapper {
...
+      if (tagName === 'SELECT') {
+        this.trigger('change')
+      } else {
+        this.trigger('input')
+      }
+
+      // for v-model.lazy, we need to trigger a change event, too.
+      // $FlowIgnore
+      if (
+        tagName === 'INPUT' &&
```
dobromir-hristov: `.lazy` can also be used on `TEXTAREA`.

lmiller1990:
passes locally - why is CI failing? 🤔

packages/test-utils/src/wrapper.js
lmiller1990: oops, updated! all green locally. CI failing though 🤔

lmiller1990:
oh, I guess lazy isn't in old versions of Vue. I'll fix that.

lmiller1990:
oh, I guess lazy isn't in old versions of Vue. I'll fix that.

Issue #1461, opened 2020-03-09
Add a debuger function to check wrapper content  


SergiCL:
### What problem does this feature solve?
Actually, if you try to see the contents of your wrapper the result is a little bit confusing.
It could work when wrappers from simple elements, but if biger it tursn unreadable:

```javascript
console.log(simpleWrapper)
// output:  Wrapper { selector: 'p' }

console.log(bigerWrapper)
// output:  VueWrapper { isFunctionalComponent: undefined, emitted: [Object: null prototype] {},_emittedByOrder: [] }
```

One posible alternative could be to use `.html()` before log the content, but this function doesen't work with array wrappers.

My proposal is to add a function to the wrappers that allows to visualize its content in a clear and simple way, without worrying about what kind of wrapper it is. A well formated pretty print of the html of the wrapper element, who gives support to array wrappers too.

It can be very useful to understand what is really being tested, especially for people in their first steps who are still learning how the library works.

#### For simple wrapper:

```javascript
wrapper.debug()

// Output:
 Wrapper:
    <div>
      <p>1</p>
    </div> 
```

#### For array wrapper:

```javascript
wrapperArray.debug()

// Output:
Wrapper Array:
   At 0:
    <div>
      <p>1</p>
    </div> 

  At 1:
    <div>
      <p>2</p>
    </div> 

  At 2:
    <div>
      <p>3</p>
    </div> 
```

afontcu:
Hi! This is an interesting suggestion, we're currently focused on releasing `v1.0` but we'll keep it in mind for future releases :)

By the way, you could log your wrapped elements in a (arguably more) convoluted way:

```js
wrapper.findAll('p').wrappers.forEach(wrapper => console.log(wrapper.html()))
```

or even 

```js
console.log([...wrapper.findAll('p').wrappers.map(wrapper => wrapper.html())])
```

it is not nice, but we'll do the trick 😇

SergiCL:
> it is not nice, but we'll do the trick 😇

Is a bit tricky and hard to read, but it's for debugging so it's right for me.

>  we're currently focused on releasing v1.0 but we'll keep it in mind for future releases :)

Given that you're busy and since it doesn't seem like a very complex functionality. Is it okay if I try to implement it, or it's better to wait until v1.0 is released?

JessicaSachs:
I’d be happy to see this in a PR personally. Thank you for the idea and for
offering!

lmiller1990:
[...]

PR #1469, opened 2020-03-14
Add `debug` function

SergiCL:
This PR contains the changes discussed in #1461 

- [x] Feature

- [x] No

### DIFF
--- a/packages/test-utils/src/wrapper-array.js
+++ b/packages/test-utils/src/wrapper-array.js
@@ -101,6 +101,19 @@ export default class WrapperArray implements BaseWrapper {
     )
   }
 
+  /**
+   * Prety print element HTML content
+   */
+  debug(): void {
+    this.throwErrorIfWrappersIsEmpty('debug')
+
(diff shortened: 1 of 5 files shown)


lmiller1990:
Great work. I personally do not want to make the API even bigger at this point - I also think `debug` is the wrong name for this function, since it does not debug anything, it just prints html. `debug` suggests it will drop you into a kind of console debugger (like `pry` from perl/python/rub  or a JS debugger, like you get in IDEs such as WebStorm.

For this reason, I think this makes way more sense as an option to `html`. Eg: `console.log(wrapper.html({ pretty: true })`. We could also set this via a `config` variable to make this the default, is the users chooses to do so.

@afontcu @JessicaSachs @dobromir-hristov thoughts?

dobromir-hristov:
Yeah same.

Stoom:
It looks like the current `wrapper.html()` already pretty prints.  From the look at the code, this just does a `console.log` (which doesn't seem to be desirable) and loops through a `WrapperArray`.

SergiCL:
I'm going close this to avoid confusion and open a new one with the changes described in #1461 when done.

PR #1473, opened 2020-03-17
fix #1377 string stubs dropping props

### LINKED ISSUE #1377: Shallow mount drops attributes on stubbed components
### Version
1.0.0-beta.30

### Reproduction link
[https://github.com/Stoom/vuetify-test-utils-beta30-example](https://github.com/Stoom/vuetify-test-utils-beta30-example)

### Steps to reproduce
* Create new project
* Add vuetify
* Use a vuetify component like v-tooltip
* Stub out to get scoped slot
* Assert an attribute is on the stubbed element

### What is expected?
It to copy over the current attributes

custom prop attributes are lost

---
This works on beta28 but broke in beta29 and beta30

Stoom:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
This deprecates using a string stub.  It fixes a missing check on the compile function along with the dropping of the component name on `template` stubs.

### DIFF
--- a/docs/api/mount.md
+++ b/docs/api/mount.md
@@ -116,7 +116,7 @@ describe('Foo', () => {
   it('renders a div', () => {
     const wrapper = mount(Foo, {
       stubs: {
-        Bar: '<div class="stubbed" />',
+        Bar: '<div class="stubbed" />', // DEPRECATED
         BarFoo: true,
         FooBar: Faz
       }
--- a/docs/api/mount.md
+++ b/docs/api/mount.md
@@ -116,9 +116,9 @@ describe('Foo', () => {
   it('renders a div', () => {
     const wrapper = mount(Foo, {
       stubs: {
-        Bar: '<div class="stubbed" />', // DEPRECATED
         BarFoo: true,
-        FooBar: Faz
+        FooBar: Faz,
+        Bar: { template: '<div class="stubbed" />' }
       }
     })
     expect(wrapper.contains('.stubbed')).toBe(true)
@@ -126,5 +126,9 @@ describe('Foo', () => {
   })
 })
 ```
+
+**Deprecation Notice:**
+
+When stubbing components, supplying a string (`ComponentToStub: '<div class="stubbed" />`) is no longer supported.
 
 - **See also:** [Wrapper](wrapper/)
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -210,10 +210,11 @@ const wrapper = mount(WrapperComp).find(ComponentUnderTest)
 
 - type: `{ [name: string]: Component | string | boolean } | Array<string>`
 
-Stubs child components. Can be an Array of component names to stub, or an object. If `stubs` is an Array, every stub is `<${component name}-stub>`.
+Stubs child components can be an Array of component names to stub, or an object. If `stubs` is an Array, every stub is `<${component name}-stub>`.
 
-Depercation note:
-If declaring an object, string stubs are now deprecated and will be removed in a future version.
+**Deprecation Notice:**
+
+When stubbing components, supplying a string (`ComponentToStub: '<div class="stubbed" />`) is no longer supported.
 
 Example:
 
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -208,9 +208,12 @@ const wrapper = mount(WrapperComp).find(ComponentUnderTest)
 
 ## stubs
 
-- type: `{ [name: string]: Component | boolean } | Array<string>`
+- type: `{ [name: string]: Component | string | boolean } | Array<string>`
 
 Stubs child components. Can be an Array of component names to stub, or an object. If `stubs` is an Array, every stub is `<${component name}-stub>`.
+
+Depercation note:
+If declaring an object, string stubs are now deprecated and will be removed in a future version.
 
 Example:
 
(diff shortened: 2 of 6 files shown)


Stoom:
Any thoughts on why it would be failing with Karma?

docs/api/mount.md
```diff
@@ -116,7 +116,7 @@ describe('Foo', () => {
   it('renders a div', () => {
     const wrapper = mount(Foo, {
       stubs: {
-        Bar: '<div class="stubbed" />',
+        Bar: '<div class="stubbed" />', // DEPRECATED
```
lmiller1990: if it is deprecated should we just nuke it from the docs?

docs/api/options.md
```diff
@@ -208,10 +208,13 @@ const wrapper = mount(WrapperComp).find(ComponentUnderTest)
...
 ## stubs
 
-- type: `{ [name: string]: Component | boolean } | Array<string>`
+- type: `{ [name: string]: Component | string | boolean } | Array<string>`
 
 Stubs child components. Can be an Array of component names to stub, or an object. If `stubs` is an Array, every stub is `<${component name}-stub>`.
 
+Depercation note:
+If declaring an object, string stubs are now deprecated and will be removed in a future version.
+
```
lmiller1990, line 218: I think we can just drop it entirely

lmiller1990 (commented):
Not sure how to handle deprecations, we really need a process for this.

Also not sure why karma is failing, can take a look this week. It is failing in the `compat` step, which runs the tests old older version of Vue. You should be able to run these locally too, there is a script (with compat) in the name. Alternatively you can just install older versions of Vue and see which one failing, starting from 2.0. There is a helper somewhere to skips tests that don't make sense for older versions (eg mainly slots had some breaking changes along the way).

Great job on this!

docs/api/mount.md
Stoom: I'd make an argument against that as the typescript definitions show strings as a valid option.  It's very frustrating when you see something in the code, but missing in the documentation.  Is there a better place to indicate that strings are deprecated?

lmiller1990: Right, makes sense. Not sure.... strings aren't really deprecated, just ones that are valid HTML. Could we just add a check and throw and error if the string includes `<` or `>`, since you can still pass something like "Text". Or are we dropping that entirely too? In that case we would change the definitions and remove `string`?

Stoom: If you were to just pass text then the vue template compiler will complain about it not being a valid template.

```
  options.stub with mount
[vue-test-utils]: String stubs are deprecated and will be removed in future versions
[Vue warn]: Error compiling template:

Hello

- Component template requires a root element, rather than just text.
```

lmiller1990: Right, I wasn't thinking. You'd need to pass `h('text')`.

Let's just drop it, and remove `string` from the TS definitions if it's not longer valid? Or am I missing something 🤔

Stoom: I've seen projects use it in the past.  To provide a migration path I'd say we keep it in until 1.0 is official.  That way a major version change will help manage a good point to make a breaking change.  Thoughts on that?

lmiller1990: I think making breaking changes **before** 1.0 make a lot more sense - once we hit 1.0, deprecating things becomes a lot more difficult.

What do you think if we just deprecate it right now and add a check + warning "This is now deprecated"?

Either way let's make a decision and get this merged up! I'll leave it up to you to make the final call.

Stoom: Deprecated as in still works for maybe the next release or two.  then rip it out?  I'm good with that.  Really it's just to give people a chance to migrate without completely breaking any of these stubs.

Completely agree with after 1.0 it's a pain to get something out.  I'd suggest that we remove this completely before 1.0 final release.  Any idea on how many more beta's there will be?

docs/api/options.md
dobromir-hristov: Because we dont have versioned docs, I think it makes sense to leave it. We are deprecating, but have not yet removed it. (We saw what removing `sync` did without mentioning it in the docs any more).

Maybe we can add an example what is now deprecated?

docs/api/mount.md
dobromir-hristov: Just show the right methods here, and add a note below, of what used to work, but is now deprecated. 

New users get to read the valid syntax, and those that are searching will find why its erroring out when they update :)

dobromir-hristov (approved):

docs/api/mount.md
lmiller1990: Sounds good. Can we get this added and merge this @Stoom , thanks for sticking with this PR for so long

I have no idea about the number of betas, since I think are unclear on what constitutes a 1.0, I think we should get Vue 3 support working (it's WIP, hopefully public alpha soon) and call whatever we have at that point the 1.0 release.

Stoom:
@lmiller1990 @dobromir-hristov I've updated to docs.  Let me know what you think, or how you would change it 🙂

dobromir-hristov:
lgtm

docs/api/mount.md
JessicaSachs: Uh are we dropping string templates entirely...?

PR #1476, opened 2020-03-18
feat: support object class binding in stubbed functional components

### LINKED ISSUE #1474: Dynamic classes not added correctly to wrapper classes when using shallowMount
### Version
1.0.0-beta.32

### Reproduction link
[https://github.com/kierans/vue-test-utils-bug](https://github.com/kierans/vue-test-utils-bug)

### Steps to reproduce
1. Create a SFC  called `Card.vue`

```
<template>
  <v-card
    outlined
    flat
  >
    <v-list-item>
      <v-list-item-avatar
          color="grey"
      />
      <v-list-item-content>
        <v-list-item-title>{{ contents.title }}</v-list-item-title>
        <v-list-item-subtitle>
          {{ contents.subtitle }}
        </v-list-item-subtitle>
      </v-list-item-content>
    </v-list-item>

    <v-card-text
        class="subtitle-1"
        v-bind:class="{
        'negative-amount': contents.amount < 0
      }"
    >
      {{ contents.amount }}
    </v-card-text>
  </v-card>
</template>

<script>
export default {
  name: "Card",
  data() {
    return {
      contents: {
        title: "Title",
        subtitle: "Subtitle",
        amount: -10
      }
    };
  }
}
</script>
```

2. Write the following test code

```
wrapper = shallowMount(Card);

const text: Wrapper<Vue> = wrapper.find(VCardText as ComponentOptions<Vue>);
expect(text.classes("negative-amount"), "Amount not marked as negative").to.be.true;
```

### What is expected?
That the classes array/string will contain the value "negative-amount"

The classes array/string contains "[object Object]"

---

When running the app, the Card is correctly rendered with the "negative-amount" class being added.

lmiller1990:
fixes https://github.com/vuejs/vue-test-utils/issues/1474

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -68,7 +68,7 @@ function createClassString(staticClass, dynamicClass) {
   // see https://github.com/vuejs/vue-test-utils/issues/1474 for more context.
   if (typeof dynamicClass === 'object') {
     evaluatedDynamicClass = Object.keys(dynamicClass).reduce((acc, key) => {
-      if (dynamicClass[key] === true) {
+      if (dynamicClass[key]) {
         return acc + ' ' + key
       }
       return acc
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -60,10 +60,25 @@ function getCoreProperties(componentOptions: Component): Object {
 }
 
 function createClassString(staticClass, dynamicClass) {
-  if (staticClass && dynamicClass) {
-    return staticClass + ' ' + dynamicClass
+  // :class="someComputedObject" can return a string, object or undefined
+  // if it is a string, we don't need to do anything special.
+  let evalutedDynamicClass = dynamicClass
+
+  // if it is an object, eg { 'foo': true }, we need to evaluate it.
+  // see https://github.com/vuejs/vue-test-utils/issues/1474 for more context.
+  if (typeof dynamicClass === 'object') {
+    evalutedDynamicClass = Object.keys(dynamicClass).reduce((acc, key) => {
+      if (dynamicClass[key] === true) {
+        return acc + ' ' + key
+      }
+      return acc
+    }, '')
   }
-  return staticClass || dynamicClass
+
+  if (staticClass && evalutedDynamicClass) {
+    return staticClass + ' ' + evalutedDynamicClass
+  }
+  return staticClass || evalutedDynamicClass
 }
 
 function resolveOptions(component, _Vue) {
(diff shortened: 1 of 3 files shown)


kierans (approved):
Looks good to me! Well done @lmiller1990 👍

packages/create-instance/create-component-stubs.js
```diff
@@ -60,10 +60,25 @@ function getCoreProperties(componentOptions: Component): Object {
 }
 
 function createClassString(staticClass, dynamicClass) {
-  if (staticClass && dynamicClass) {
-    return staticClass + ' ' + dynamicClass
+  // :class="someComputedObject" can return a string, object or undefined
+  // if it is a string, we don't need to do anything special.
+  let evalutedDynamicClass = dynamicClass
```
afontcu:
```suggestion
  let evaluatedDynamicClass = dynamicClass
```

😇

afontcu (approved):

packages/create-instance/create-component-stubs.js
```diff
@@ -60,10 +60,25 @@ function getCoreProperties(componentOptions: Component): Object {
...
-    return staticClass + ' ' + dynamicClass
+  // :class="someComputedObject" can return a string, object or undefined
+  // if it is a string, we don't need to do anything special.
+  let evaluatedDynamicClass = dynamicClass
+
+  // if it is an object, eg { 'foo': true }, we need to evaluate it.
+  // see https://github.com/vuejs/vue-test-utils/issues/1474 for more context.
+  if (typeof dynamicClass === 'object') {
+    evaluatedDynamicClass = Object.keys(dynamicClass).reduce((acc, key) => {
+      if (dynamicClass[key] === true) {
```
dobromir-hristov: Should we check if this is a truthy value or like this, straight up true?

packages/create-instance/create-component-stubs.js
```diff
@@ -60,10 +60,25 @@ function getCoreProperties(componentOptions: Component): Object {
...
+      }
+      return acc
+    }, '')
+  }
+
+  if (staticClass && evaluatedDynamicClass) {
+    return staticClass + ' ' + evaluatedDynamicClass
   }
-  return staticClass || dynamicClass
+  return staticClass || evaluatedDynamicClass
```
dobromir-hristov, line 81: Wouldnt have pushing all to an array and using `join(' ')` at the end been a more cleaner way of doing this?

dobromir-hristov (approved):

packages/create-instance/create-component-stubs.js
lmiller1990: Not sure, I just went with this to be explicit. I haven't seen people use something like `:class={ 'my-class': 'some-truthy-val' }`. I guess we should just copy whatever Vue does - will try it out.

packages/create-instance/create-component-stubs.js
lmiller1990: I think both are fine - was having problems iterating over `Object.keys`, using `of` throws some warning. A few other places in the codebase do `Object.keys().reduce` so I just tried to follow that same style.

I like `reduce` but I don't mind either way.

dobromir-hristov:
gg

kierans:
What's the release process please?

PR #1481, opened 2020-03-22
fix #1477 Note exception to chaining find calls

### LINKED ISSUE #1477: find throws error when using refs
### Version
1.0.0-beta.30

### Reproduction link
[https://gitlab.com/unclejustin/test-utils-but](https://gitlab.com/unclejustin/test-utils-but)

### Steps to reproduce
In a test reference an element by ref. Then try to find anything in that element using anything other than a dom selector:

const button = wrapper.find({ ref: "testButton" });
expect(button.find(Icon).exists()).toBe(true);

This will throw the errror "[vue-test-utils]: cannot find a Vue instance on a DOM node. The node you are calling find on does not exist in the VDom. Are you adding the node as innerHTML?"

### What is expected?
That found elements have the same functionality as root wrappers.

Throwing an error.

---
I think this should work as described, but if I am wrong then maybe the docs just need to be updated to reflect the actual functionality?

unclejustin:
Make it clear that when chaining find calls, you can only use DOM selectors.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Docs update

- [ ] Yes
- [x] No

- [-] New/updated tests are included

- [-] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**


JessicaSachs:
Nice. Thank you. Can you run the formatter for the docs? `yarn format`

unclejustin:
Whoops! My bad, thanks @JessicaSachs formatted and pushed.

lmiller1990:
Why is this the case? I would have thought `find` returns a wrapper which should let you use `find` with any kind of selector. Interesting. Good catch. I wonder if we should totally disallow multiple `find` calls entirely, if it doesn't cover all the cases.

Can you remove the generated `dist` files? We don't commit those - those are generated when the next version of the package is published to npm (which will be soon with all the PRs that were merged lately).

Stoom:
I've used multiple `find` calls in the past to split things up in my testing:

```js
describe('SomeComponent.vue', () => {
  let button;
  beforeEach(() => {
    button = wrapper.find('button-stub');
  });

  it('Should have an icon', () => {
    const icon = button.find('icon-stub');
    expect(icon.exists()).to.be.true;
  });
});
```

So this is minimal, but if I had multiple tests on the button then it's cleaner for me to do that outside of each `it` block.

lmiller1990:
@Stoom good point. Would be good to figure out why this is the case and fix.

PR #1491, opened 2020-03-29
feat(test-utils): add 'overview' function

### LINKED ISSUE #1461: Add a debuger function to check wrapper content

SergiCL:
close #1461

- [x] Feature

- [x] No

**Other information:**
The main objective of this feature is "To give an overview of the state of the wrapper at the time the function is executed"

Example execution:
```javascript
wraper.overview()

// Output
Wrapper (Visible):

Html:
    <div class="test">
      <p>My name is Tess Ting</p>
    </div>

Data: {
    firstName: Tess,
    lastName: Ting
}

Computed: {
    fullName: Tess Ting'
}

Emitted: {',
    foo: [',
        0: [ hello, world ],
        1: [ bye, world ]'
    ],
    bar: [
        0: [ hey ]'
    ]
}
```

[More info](https://github.com/vuejs/vue-test-utils/issues/1461#issuecomment-602072301)

### DIFF
--- a/test/specs/wrapper/overview.spec.js
+++ b/test/specs/wrapper/overview.spec.js
@@ -1,0 +1,334 @@
+import { describeWithShallowAndMount } from '~resources/utils'
+
+describeWithShallowAndMount('overview', mountingMethod => {
+  const originalLog = console.log
+  let consoleOutput = []
+  const consoleLogMock = (...output) =>
+    consoleOutput.push(...output.join(' ').split('\n'))
+
+  beforeEach(() => {
+    consoleOutput = []
+    console.log = consoleLogMock
+  })
+  afterEach(() => (console.log = originalLog))
+
+  it('prints a simple overview of the Wrapper', () => {
+    const wrapper = mountingMethod({
+      template:
+        '<div class="test" style="{ color: red }"><p>My name is {{ firstName }} {{ lastName }}</p></div>',
+      data() {
+        return {
+          firstName: 'Tess',
+          lastName: 'Ting'
+        }
+      },
+      computed: {
+        onePlusOne: () => 1 + 1
+      }
+    })
+    wrapper.vm.$emit('foo', 'hello', 'world')
+    wrapper.vm.$emit('foo', 'bye', 'world')
+    wrapper.vm.$emit('bar', 'hey')
+
+    const expectedConsoleOutput = [
+      '',
+      'Wrapper (Visible):',
+      '',
+      'Html:',
+      '    <div class="test">',
+      '      <p>My name is Tess Ting</p>',
+      '    </div>',
+      '',
+      'Data: {',
+      '    firstName: Tess,',
+      '    lastName: Ting',
+      '}',
+      '',
+      'Computed: {',
+      '    onePlusOne: 2',
+      '}',
+      '',
+      'Emitted: {',
+      '    foo: [',
+      '        0: [ hello, world ],',
+      '        1: [ bye, world ]',
+      '    ],',
+      '    bar: [',
+      '        0: [ hey ]',
+      '    ]',
+      '}',
+      ''
+    ]
+    wrapper.overview()
+    expect(consoleOutput).to.have.ordered.members(expectedConsoleOutput)
+  })
+
+  it('throws error when called on non VueWrapper', () => {
+    const wrapper = mountingMethod({ template: '<div><p /></div>' })
+    const nonVueWrapper = wrapper.find('p')
+    const message =
+      '[vue-test-utils]: wrapper.overview() can only be called on a Vue instance'
+
+    expect(() => nonVueWrapper.overview())
+      .to.throw()
+      .with.property('message', message)
+  })
+
+  describe('vibility', () => {
+    it('prints "Visible" when the wrapper is visible', () => {
+      const wrapper = mountingMethod({ template: '<div class="test"></div>' })
+
+      const expectedConsoleOutput = [
+        '',
+        'Wrapper (Visible):',
+        '',
+        'Html:',
+        '    <div class="test"></div>',
+        '',
+        'Data: {}',
+        '',
+        'Computed: {}',
+        '',
+        'Emitted: {}',
+        ''
+      ]
+
+      wrapper.isVisible = () => true
+      wrapper.overview()
+      expect(consoleOutput).to.have.ordered.members(expectedConsoleOutput)
+    })
+
+    it('prints "Not Visible" when the wrapper is not visible', () => {
+      const wrapper = mountingMethod({ template: '<div class="test"></div>' })
+
+      const expectedConsoleOutput = [
+        '',
+        'Wrapper (Not visible):',
+        '',
+        'Html:',
+        '    <div class="test"></div>',
+        '',
+        'Data: {}',
+        '',
+        'Computed: {}',
+        '',
+        'Emitted: {}',
+        ''
+      ]
+
+      wrapper.isVisible = () => false
+      wrapper.overview()
+      expect(consoleOutput).to.have.ordered.members(expectedConsoleOutput)
+    })
+  })
+
+  describe('html', () => {
+    it('prints Html as empty when html is not defined', () => {
+      const wrapper = mountingMethod({ template: '' })
+
+      const expectedConsoleOutput = [
+        '',
+        'Wrapper (Visible):',
+        '',
+        'Html:',
+        '',
+        'Data: {}',
+        '',
+        'Computed: {}',
+        '',
+        'Emitted: {}',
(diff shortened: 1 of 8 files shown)


lmiller1990:
Holy moly what a PR! I like how in depth your tests go.

Please give me some time to review this.

SergiCL:
> Holy moly what a PR! I like how in depth your tests go.
>
> Please give me some time to review this.

Thank you. No problem, take as much time as you need.

test/specs/wrapper/overview.spec.js
```diff
@@ -0,0 +1,336 @@
...
+          ''
+        ]
+
+        wrapper.isVisible = () => true
+        wrapper.overview()
+        expect(consoleOutput).to.have.ordered.members(expectedConsoleOutput)
+      })
+
+      it('does not print child component data or computed', () => {
+        if (mountingMethod.name === 'shallowMount') return
```
lmiller1990: I might be missing something - why are we skipping `shallowMount`?

test/specs/wrapper/overview.spec.js
```diff
@@ -0,0 +1,336 @@
...
+          '',
+          'Html:',
+          '    <div class="test"></div>',
+          '',
+          'Data: {}',
+          '',
+          'Computed: {}',
+          '',
+          'Emitted: {',
+          '    foo: [',
```
lmiller1990, line 211: what is displayed if objects are emitted? Probably need a test for this

test/specs/wrapper/overview.spec.js
```diff
@@ -0,0 +1,336 @@
...
+    const nonVueWrapper = wrapper.find('p')
+    const message =
+      '[vue-test-utils]: wrapper.overview() can only be called on a Vue instance'
+
+    expect(() => nonVueWrapper.overview())
+      .to.throw()
+      .with.property('message', message)
+  })
+
+  if (vueVersion > 2) {
```
lmiller1990, line 26: how come we need this check? This lib only supports Vue 2.x - doe s something weird happen if you remove this?

lmiller1990 (commented):
I left some comments!

test/specs/wrapper/overview.spec.js
SergiCL: The test checks if the HTML shown in the overview includes the child component. 

With mount child component is:
`<div class="tester">test</div>`

But with shallow mount is something like this:
`<tester-stub></tester-stub>`

I don't know if it's really necessary to test this with shallow mount, but to make the test clearer I can add a diferent expect for every mounting method. 

For example:

```javascript
const expectedConsoleOutput = [
	  '',
	  'Wrapper (Visible):',
	  '',
	  'Html:',
	  mountingMethod.name === 'shallowMount'
	    ? '    <div>1<tester-stub></tester-stub>'
	    : '    <div>1<div class="tester">test</div>',
	  '    </div>',
	  '',
	  'Data: {}',
	  '',
	  'Computed: {}',
	  '',
	  'Emitted: {}',
	  ''
	]
```

SergiCL: > it ('does not print child component data or computed')

I got the test mixed up. You're talking about the one for computed and data. In this case it makes more sense, anyways the solution would be the same.

test/specs/wrapper/overview.spec.js
SergiCL: I totally agree. I'll get right on it.

Edit: I will add a test to check when the object is emitted, and when the emit is empty.

test/specs/wrapper/overview.spec.js
SergiCL: The test of `test:compat` fails at Vue 2.0.
This is because hooks are added to the emitted:

```javascript
"Emitted: {"
	" hook:beforeCreate: ["
	" 0: [ ]"
	" ],"
	" hook:created: ["
	" 0: [ ]"
	" ],"
	" hook:beforeMount: ["
	" 0: [ ]"
	" ],"
	" hook:mounted: ["
	" 0: [ ]"
	" ]"
 "}"
```

I don't know exactly how many people use version 2.0. If we want to test it, I can think of the following options:

1. **Duplicate the tests (or a part of them) in an `else`**: We repeat a lot of code but gain in readability.

2.  **Extract the expectedConsoleOutput to a function (at least the emitted)**: The tests are less clear, but it's better than adding a condition to each test.

SergiCL: I think I have a middle ground solution. 
Right now we're checking the outcome with this:

```javascript
expect(consoleOutput).to.have.ordered.members(expectedConsoleOutput)
```

If we replace the `have.ordered` with an `include`  it should work for Vue 2.0:
```
expect(consoleOutput).to.include.members(expectedConsoleOutput)
```

The only problem we have is when nothing has been emitted. In that case the expected emit will be `Emitted: {}` and in 2.0 it will show:

```javascript
"Emitted: {"
" hook:beforeCreate: ["
" 0: [ ]"
" ],"
" hook:created: ["
" 0: [ ]"
" ],"
" hook:beforeMount: ["
" 0: [ ]"
" ],"
" hook:mounted: ["
" 0: [ ]"
" ]"
"}"
```

My proposal is to create a simple function that takes care of this without making the tests dirty.

Something like:
```javascript
expect(compareConsoleOutput(consoleOutput, expected)).to.be.true
```
or even a function that includes the expect inside:
```javascript
expectToHaveConsoleOutputs(consoleOutput, expectedConsoleOutputs)
```

test/specs/wrapper/overview.spec.js
lmiller1990: sounds good

test/specs/wrapper/overview.spec.js
lmiller1990: I think it's fine to show the stubbed html for the shallow mounted components. shallowMount's output is basically useless, but I think it's better to be consistent and show it anyway. At least from my point of view, I would expect overview to show the same information for both rendering APIs.

test/specs/wrapper/overview.spec.js
lmiller1990: I see - I think it's fine to leave the check in. I doubt many people are on 2.0 now anyway. No problem 👍

lmiller1990 (approved):
looks like a solid PR.

this is the last API we will be adding before v1! This will go out with v1, should be in a week or two.

julisch94:
Hey folks, I'm unsure if this is the right place to ask. I would like such an `overview()` function but I can't find it anywhere even though I can see it was added by you. 

It's still in the code:
https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/wrapper.js#L450-L455C21
(even though deprecated, why?)

There's also nothing about it in the docs:
https://v1.test-utils.vuejs.org/api/wrapper

So how can I print the DOM tree of a component with vue test utils like I can do with react testing library using `screen.debug()`?

Anyone can help me understand this a bit better?

lmiller1990:
You can print the HTML with `wrapper.html()` https://test-utils.vuejs.org/api/#html

I am not sure about `overview()`, I completely forgot about it. I think the confusion is the `overview()` function was added to v1 of this lib (which works with Vue 2) but you might be on v2 of this lib (which works with Vue 3).

julisch94:
@lmiller1990 oh right! Thanks! I was looking at the v1 of the docs. Didn't notice that Google sent me to v1 instead of v2. And also thanks for pointing me to `wrapper.html()`. It served my purpose. :)

PR #1502, opened 2020-04-09
Adding missing documentation

vegerot:
When reading through the documentation I was very confused why it randomly started talking about `setTimeout`, when I realized it was referring to a hypothetical `setTimeout` example that wasn't there.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x ] Other, please describe: Documentation

- [ ] Yes
- [x ] No

- [x ] It's submitted to the `dev` branch.
- [ x] All tests are passing: https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md#development-setup

**Other information:**

### DIFF
--- a/docs/guides/testing-async-components.md
+++ b/docs/guides/testing-async-components.md
@@ -84,7 +84,7 @@ it('fetches async when a button is clicked', () => {
 })
 ```
 
-This test currently fails because the assertion is called before the promise in `fetchResults` resolves. Most unit test libraries provide a callback to let the runner know when the test is complete. Jest and Mocha both use `done`. We can use `done` in combination with `$nextTick` or `setTimeout` to ensure any promises are settled before the assertion is made.
+This test currently fails because the assertion is called before the promise in `fetchResults` resolves. Most unit test libraries provide a callback to let the runner know when the test is complete. Jest and Mocha both use `done`. We can use `done` in combination with `$nextTick` or `setTimeout`/`setImmediate` to ensure any promises are settled before the assertion is made.
 
 ```js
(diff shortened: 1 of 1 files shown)


lmiller1990:
Thanks for the contribution. I agree the docs are not very good.

I'm not sure our docs should go this in depth anyway. Should we just delete the `setTimeout` example, and all mentions to it? I don't think anyone is actually writing tests like this, it bleeds implementations details into your specs.

Can we just simplify this to explain to use `await flushPromises` to ensure the DOM is updated or something?

We will redo docs entirely for Vue 3 integration main for this reason - they are way too complicated.

lmiller1990:
@vegerot still interested to get this merged? I think we should try and make it a bit more concise.

vegerot:
I’ve always been interested in merging this 🙂

lmiller1990:
@vegerot can we simply the explanation? I don't think the docs for a test runner is the right place to explain the complexities of the event loop. It think we already have way too much information that is not about testing. 

We could include something more simple like "This problem occurs because Vue updates the DOM asynchronously, and the test runner executes assertions synchronously. Work around this by running `await wrapper.vm.$nextTick` to ensure the DOM is updated before the assertion runs. See <this external link> for more information".

I don't think `setTimeout` is something we should encourage or mention. While this does work, it's not very explicit or clear why it does, unless you know the event loop very well. `nextTick` is definitely more clear. It would be great if we could just have "one way" of doing things.

vegerot:
Yeah, that makes sense.  I'll try that out

Issue #1484, opened 2020-03-22
Failed to mount when returning render function from setup()


LeBenLeBen:
### Version
1.0.0-beta.32

### Reproduction link
[https://github.com/sduduzog/vue-composition-api-tsx-example](https://github.com/sduduzog/vue-composition-api-tsx-example)

### Steps to reproduce
1. Clone the repo
2. Update the package.json to have the latest @vue/test-utils version, current is 1.0.0-beta.32
3. Install dependencies
4. Run `npm run test:unit`

### What is expected?
The component "HelloWorld" should be properly mounted.

### What is actually happening?
The component doesn't mount with the following error:

> Failed to mount component: template or render function not defined.

---
This is actually a cross-post from https://github.com/vuejs/composition-api/issues/151. I’m looking for help to resolve this issue where render functions returned from the composition api `setup` functions are not properly set while mounted through vue-test-utils.

At this point I’m wondering if vue-test-utils is the problem here or if it’s totally unrelated. And if it is, I would appreciate some guidance about where to look next to debug it.

lmiller1990:
I can't even get returning a render function from `setup` to work in a regular app. Is this even supported? It doesn't say it is here: https://github.com/vuejs/composition-api

I'll try pulling that repo. We should repro without TS; the less dependencies, the better.

Edit: I cloned that repo and ran `yarn` and got:

```
warning ../package.json: No license field
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "https://npm.codeo.co.za/@vue%2fcli-plugin-pwa/-/cli-plugin-pwa-4.2.2.tgz: Request failed \"401 Unauthorized\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/lachlan/code/dump/vue-composition-api-tsx-example/yarn-error.log".
```

🤔

dobromir-hristov:
According to https://github.com/vuejs/composition-api#template-refs it doesnt. That will work for Vue 3 i think.

lmiller1990:
This is my understanding too - unless something has changed. It looks like some people have it working? Maybe the readme is outdated. From what I can tell Vue 2 does not expose `h` or `createElement` in the same way Vue 3 does. This definitely works in Vue 3, at least using regular `h` (which is basically what jsx is).

LeBenLeBen:
@dobromir-hristov @lmiller1990 I think that the docs you mention is specific to refs, no?

[...]

PR #1512, opened 2020-04-23
WIP feat(test-utils): support returning render function from setup

lmiller1990:
https://github.com/vuejs/composition-api/issues/151
#1484 

4 tests still failing but a good proof of concept.

### DIFF
--- a/packages/test-utils/src/mount.js
+++ b/packages/test-utils/src/mount.js
@@ -22,6 +22,10 @@ export default function mount(component, options = {}) {
 
   const _Vue = createLocalVue(options.localVue)
 
+  if (component.setup) {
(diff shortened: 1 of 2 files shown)


lmiller1990:
This is probably impossible w/o changing how we do stubs.

lmiller1990:
Will be fixed in vuejs/composition-api#383

packages/test-utils/src/mount.js
```diff
@@ -22,6 +22,10 @@ export default function mount(component, options = {}) {
 
   const _Vue = createLocalVue(options.localVue)
 
+  if (component.setup) {
```
pikax, line 25: This shouldn't be required

PR #1523, opened 2020-04-30
Release: 1.0.0

lmiller1990:
A somewhat historic moment after years of waiting, endless bugfixes, debates about `shallowMount`, blood, sweat and tears is upon us - the greatly anticipated 1.0.0.

Some bugs will likely never be fixed due to flaws in the library design - we learned a lot and are using the lessons to improve with [VTU next](github.com/vuejs/vue-test-utils-next/), which targets Vue 3.

Thanks to the [217 contributors](https://github.com/vuejs/vue-test-utils/graphs/contributors) who helped build VTU, and the countless others who reported bugs, wrote articles, or contributed in some other form to the library! Special thanks must go to @eddyerburgh for writing the original library and acting as the lead maintainer for most of its existence.

The goal of this PR is to ensure we cover bases before doing a 1.0.0, sometime in the next week or two. This means API changes will be less likely to happen, and if they do, happen via the RFC process. It doesn't mean development will stop - as long as people are using the library, we (all the OSS contributors) will continue to improve the library, fix bugs, and update the documentation.

TODO before 1.0.0

- [ ] ensure all deprecation messages are up to date and accurate
- [ ] important features/changes are backported from VTU next
- [ ] write clear and concise release notes, detailing the "final" API and notable behaviors
- [ ] clean up docs a little - make sure any references to old APIs or behaviors are gone, remove some un-needed info.
- [ ] ??

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,5 +5,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "1.0.0-beta.33"
+  "version": "1.0.0"
 }
--- a/packages/server-test-utils/dist/vue-server-test-utils.js
+++ b/packages/server-test-utils/dist/vue-server-test-utils.js
@@ -6,9 +6,9 @@ function _interopDefault (ex) { return (ex && (typeof ex === 'object') && 'defau
 
 var Vue = _interopDefault(require('vue'));
 var vueTemplateCompiler = require('vue-template-compiler');
-var vueServerRenderer = require('vue-server-renderer');
 var testUtils = require('@vue/test-utils');
 var testUtils__default = _interopDefault(testUtils);
+var vueServerRenderer = require('vue-server-renderer');
 var cheerio = _interopDefault(require('cheerio'));
 
 // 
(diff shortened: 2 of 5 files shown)


dobromir-hristov:
### Deprecation

I checked the deprecation messages, we have covered what makes sense. We do not offer the new mount option structure, so it does not make sense to emit about those. The one for `findComponent` is not correct, as it can also take `querySelector` strings, but it would make refactor more annoying. Then again, I am sure people would like to be able to target by it. Maybe we can ensure it only returns Component instances and never DOM nodes, and leave out the string check? `find` is fine as it is.

### Features

I think with the addition of `findComponent`and returning promise from `trigger` we cover the most important features.

### Release Notes

What do we want to mention here? What has changes since last release or? If so, that should be straight forward, though release notes are not a guide :/

### About docs:

1. I tried to refactor the `async` related stuff, was a bit of a mess. The `Guides` section is a mess, things getting repeated over and over, without a clear path to follow. 
1. It makes sense to keep 90% of that info, just move it around, make it consistent. 
1. Also that hack with injecting sub-pages into `guides` page, causes some pages to lead to dead ends, but I dont have a better idea atm. Vuepress refuses to add none-renderable pages - https://github.com/vuejs/vuepress/issues/2101
1. I think we should have a mention about the old "sync"

lmiller1990:
Regarding`findComponent` taking a string - it's pretty annoying to refactor, so I think we just leave it as-is. We can remove the fact it works with querySelector from the docs.

I guess release notes should be easy - you are right.

Improving docs is a big task... I guess we do it incrementally. A lot of info is repeated like you mentioned. I think we need to be a bit brutal and cut a lot of it, but this probably can be done incrementally and not block 1.0.

PR #1538, opened 2020-05-07
Sync docs with v1.0

### LINKED ISSUE #1536: Marking deprecated methods in docs
Misunderstood in docs

--

Is it worth to mark deprecated methods with some warning? And maybe worth to mention about using `@testing-library/jest-dom` as replace? Since in cli we have deprecation message, in docs we don't have such info.

Don't know why but it took me a minute to figure out that I need to use `.element` in my assert method to use i.e. `toBeEmpty()`. Maybe it's also worth to mention in docs.

so it looks like this now
```
expect(wrapper.find('.empty-el').element).toBeEmpty()
```

instead of:
```
expect(wrapper.find('.empty-el').isEmpty()).toBe(true)
```

afontcu:
Closes #1536.

This PR aims to sync docs with `v1.0` release.


lmiller1990 (approved):
Thanks!

JessicaSachs:
Love it.

JessicaSachs (approved):

hiendv:
I've been testing with **find** and realize that `find({string} selector)` would also yield an object. Hence, we should also encourage using `.element` when querying with selectors, not only `findComponent`.

PR #1548, opened 2020-05-14
fix #1541: Improve deprecation messages

### LINKED ISSUE #1541: Explain what the replacement for deprecations are
The documentation for for example [setMethods](https://vue-test-utils.vuejs.org/api/wrapper-array/#setmethods) is unclear. It simply states: "setMethods is deprecated and will be removed in future releases.". Same goes for [isVueInstance](https://vue-test-utils.vuejs.org/api/wrapper-array/#isvueinstance).

This does not explain what the replacement (if any) should be, or why this was deprecated.

These types of warnings also happens when running the tests:
"[vue-test-utils]: overwriting methods via the `methods` property is deprecated and will removed in the next major version"

Here it's also unclear what the fix should be or why this was deprecated.

Not applicable.

afontcu:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [X] Refactor (I guess? xD)
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

I'm open to suggestions with regard the specifics of the deprecation message – not having a clear migration path for `setMethods` and `options.methods` isn't ideal, but I don't think we will ever find one.

Closes #1541

### DIFF
--- a/docs/api/wrapper-array/isVueInstance.md
+++ b/docs/api/wrapper-array/isVueInstance.md
@@ -1,6 +1,6 @@
 ## isVueInstance
 
-::: warning
+::: warning Deprecation warning
 `isVueInstance` is deprecated and will be removed in future releases.
 :::
 
--- a/docs/api/wrapper-array/isVueInstance.md
+++ b/docs/api/wrapper-array/isVueInstance.md
@@ -2,6 +2,10 @@
 
 ::: warning Deprecation warning
 `isVueInstance` is deprecated and will be removed in future releases.
+
+Most of the times, tests relying on `isVueInstance` provide little to no-value so you could simply remove them.
+
+If you want to keep these tests, a valid replacement is to assert that `wrapper.find(...).vm` exists.
 :::
 
 Assert every `Wrapper` in `WrapperArray` is Vue instance.
--- a/docs/api/wrapper-array/isVueInstance.md
+++ b/docs/api/wrapper-array/isVueInstance.md
@@ -3,9 +3,9 @@
 ::: warning Deprecation warning
 `isVueInstance` is deprecated and will be removed in future releases.
 
-Most of the times, tests relying on `isVueInstance` provide little to no-value so you could simply remove them.
+Tests relying on the `isVueInstance` assertion provide little to no value. We suggest replacing them with purposeful assertions.
 
-If you want to keep these tests, a valid replacement is to assert that `wrapper.find(...).vm` exists.
+To keep these tests, a valid replacement for `isVueInstance()` is a truthy assertion of `wrapper.find(...).vm`.
 :::
 
 Assert every `Wrapper` in `WrapperArray` is Vue instance.
--- a/docs/api/wrapper-array/setMethods.md
+++ b/docs/api/wrapper-array/setMethods.md
@@ -5,9 +5,9 @@
 
 There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods).
 
-Thus, we suggest to to rethink those tests using the tools Vue Test Utils provides.
+We suggest rethinking those tests using the tools Vue Test Utils provides.
 
-If you need to stub out a complex method, extract it out from the component and test it in isolation. On the other hand, if you want to assert that a method is called, use your test runner to spy on it.
+To stub a complex method extract it from the component and test it in isolation. To assert that a method is called, use your test runner to spy on it.
 :::
 
 Sets `Wrapper` `vm` methods and forces update on each `Wrapper` in `WrapperArray`.
--- a/docs/api/wrapper-array/setMethods.md
+++ b/docs/api/wrapper-array/setMethods.md
@@ -3,7 +3,7 @@
 ::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
 
-There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](<(https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods)>).
+There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods).
 
 Thus, we suggest to to rethink those tests using the tools Vue Test Utils provides.
 :::
--- a/docs/api/wrapper-array/setMethods.md
+++ b/docs/api/wrapper-array/setMethods.md
@@ -2,6 +2,10 @@
 
 ::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
+
+There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](<(https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods)>).
+
+Thus, we suggest to to rethink those tests using the tools Vue Test Utils provides.
 :::
 
 Sets `Wrapper` `vm` methods and forces update on each `Wrapper` in `WrapperArray`.
--- a/docs/api/wrapper-array/setMethods.md
+++ b/docs/api/wrapper-array/setMethods.md
@@ -6,6 +6,8 @@
 There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods).
 
 Thus, we suggest to to rethink those tests using the tools Vue Test Utils provides.
+
+If you need to stub out a complex method, extract it out from the component and test it in isolation. On the other hand, if you want to assert that a method is called, use your test runner to spy on it.
 :::
 
 Sets `Wrapper` `vm` methods and forces update on each `Wrapper` in `WrapperArray`.
--- a/docs/api/wrapper-array/setMethods.md
+++ b/docs/api/wrapper-array/setMethods.md
@@ -1,6 +1,6 @@
 ## setMethods
 
-::: warning
+::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
 :::
 
--- a/docs/api/wrapper-array/setMethods.md
+++ b/docs/api/wrapper-array/setMethods.md
@@ -5,7 +5,7 @@
 
 There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods).
 
-We suggest rethinking those tests using the tools Vue Test Utils provides.
+We suggest rethinking those tests.
 
 To stub a complex method extract it from the component and test it in isolation. To assert that a method is called, use your test runner to spy on it.
 :::
--- a/docs/api/wrapper/isVueInstance.md
+++ b/docs/api/wrapper/isVueInstance.md
@@ -1,6 +1,6 @@
 ## isVueInstance
 
-::: warning
+::: warning Deprecation warning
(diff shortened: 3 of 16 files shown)


lmiller1990 (approved):
excellent, thanks. will see if anyone else has any comments, otherwise merge and do a release in the next day or two.

docs/api/wrapper-array/setMethods.md
```diff
@@ -1,7 +1,11 @@
 ## setMethods
 
-::: warning
+::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
+
+There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](<(https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods)>).
```
dobromir-hristov: is this link properly formatted?

docs/api/wrapper/isVueInstance.md
```diff
@@ -1,7 +1,11 @@
 ## isVueInstance
 
-::: warning
+::: warning Deprecation warning
 `isVueInstance` is deprecated and will be removed in future releases.
+
+Most of the times, tests relying on `isVueInstance` provide little to no-value so you could simply remove them.
```
dobromir-hristov: Technically those are only valid in shallowMount environments, to assert dynamic components via `<component :is=""`, nothing much else.... I had to use it probably once or twice.

The proposed solution does the job :)

docs/api/wrapper/setMethods.md
```diff
@@ -1,7 +1,11 @@
 ## setMethods
 
-::: warning
+::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
+
+There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](<(https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods)>).
```
dobromir-hristov: We can suggest users to use `jest.spyOn(Component.methods, 'method')`, which works most of the times. Ofc this will not work with the new composition api 😆 but if you are using that, you can figure it out.

dobromir-hristov (approved):

docs/api/wrapper/setMethods.md
lmiller1990: yep, assume composition API ppl are advanced users and can figure it out

docs/api/wrapper-array/setMethods.md
afontcu: oops! copy paste went wrong. thanks! fixed

docs/api/wrapper/setMethods.md
afontcu: So you think that mentioning the ability to spy on methods is worth adding?

something along the lines of…

`
If you need to stub out a complex method, extract it out and test it.
On the other hand, if you want to assert a method is called, you can use your test runner to spy on it.
`

janosrusiczki (approved):

docs/api/wrapper/setMethods.md
lmiller1990: I like that! but either way is fine, I think.

afontcu: done!

docs/api/wrapper-array/isVueInstance.md
```diff
@@ -1,7 +1,11 @@
 ## isVueInstance
 
-::: warning
+::: warning Deprecation warning
 `isVueInstance` is deprecated and will be removed in future releases.
+
+Most of the times, tests relying on `isVueInstance` provide little to no-value so you could simply remove them.
```
JessicaSachs:
```suggestion
Tests relying on the `isVueInstance` assertion provide little to no value. We suggest replacing them with purposeful assertions.
```

docs/api/wrapper-array/setMethods.md
```diff
@@ -1,7 +1,13 @@
 ## setMethods
 
-::: warning
+::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
+
+There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods).
+
+Thus, we suggest to to rethink those tests using the tools Vue Test Utils provides.
```
JessicaSachs:
```suggestion
We suggest rethinking those tests.
```

Double "to" and "using the tools Vue Test Utils provides" is implicit.

docs/api/wrapper-array/setMethods.md
```diff
@@ -1,7 +1,13 @@
...
+
+If you need to stub out a complex method, extract it out from the component and test it in isolation. On the other hand, if you want to assert that a method is called, use your test runner to spy on it.
```
JessicaSachs:
```suggestion
To stub a complex method extract it from the component and test it in isolation. To assert that a method is called, use your test runner to spy on it.
```

More concise wording

docs/api/wrapper/isVueInstance.md
JessicaSachs:
```suggestion
Tests relying on the `isVueInstance` assertion provide little to no value. We suggest replacing them with better assertions.
```

Active voice

docs/api/wrapper/isVueInstance.md
```diff
@@ -1,7 +1,11 @@
...
+
+If you want to keep these tests, a valid replacement is to assert that `wrapper.find(...).vm` exists.
```
JessicaSachs:
```suggestion
To keep these tests, a valid replacement for `isVueInstance()` is a truthy assertion of `wrapper.find(...).vm` 
```

docs/api/wrapper/setMethods.md
```diff
@@ -1,7 +1,13 @@
 ## setMethods
 
-::: warning
+::: warning Deprecation warning
 `setMethods` is deprecated and will be removed in future releases.
+
+There's no clear path to replace `setMethods`, because it really depends on your previous usage. It easily leads to flaky tests that rely on implementation details, which [is discouraged](https://github.com/vuejs/rfcs/blob/668866fa71d70322f6a7689e88554ab27d349f9c/active-rfcs/0000-vtu-api.md#setmethods).
+
+Thus, we suggest to to rethink those tests using the tools Vue Test Utils provides.
```
JessicaSachs:
```suggestion
We suggest rethinking those tests.
```

docs/api/wrapper/setMethods.md
```diff
@@ -1,7 +1,13 @@
...
+
+If you need to stub out a complex method, extract it out from the component and test it in isolation. On the other hand, if you want to assert that a method is called, use your test runner to spy on it.
```
JessicaSachs:
```suggestion
To stub a complex method extract it from the component and test it in isolation. To assert that a method is called, use your test runner to spy on it.
```

docs/api/wrapper/isVueInstance.md
JessicaSachs: Even if you don't take the suggestion, "Most of the times" should be "Most of the time"

JessicaSachs (commented):
I added some rewording for active voice, two-three minor typos, and conciseness. Feel free to take them or leave them.

docs/api/wrapper-array/isVueInstance.md
afontcu: Thanks for the suggestions, @JessicaSachs! very much appreciated 🤗

dcb99:
Improved deprecation messages is nice, but you're still deprecating a feature that we have implemented in hundreds of tests. If I want to mock a function, and then assert that a button click called that function, it used to be very straightforward, and is extremely useful for verifying the code (and the UI) behaves as expected. Tests are supposed to prove that your code does what it says it does. Now we get hundreds of deprecation warnings. 
I get that we don't want tests to rely on implementation details. But the blanket suggest to "...rethink those tests" is both short-sighted and condescending. I don't think all the use cases were properly considered here, just extracting a function and testing it doesn't solve the base test case I mentioned, which is that a button click calls the right function (which we can then use to make sure other behaviors happened as expected), or that a particular function was called n times for n clicks.
If you're going to deprecate something and not provide anything other than a nebulous deprecation warning, you're not doing developers any favors. Mocking/stubbing functions isn't an antipattern, it's a pretty important part of verifying inputs and outputs.
Yes, it CAN make tests brittle, yes, it CAN be misused, but removing the ability to be productive for a use case that seems pretty common is more of an antipattern that mocking functions. Better deprecation warnings are not a solution to not being able to easily and succinctly prove that your code does what it says it does.

afontcu:
> Now we get hundreds of deprecation warnings

You can easily fix that by setting `config.showDeprecationWarnings = false` ([source](https://vue-test-utils.vuejs.org/api/#config))

> If I want to mock a function, and then assert that a button click called that function, it used to be very straightforward, and is extremely useful for verifying the code (and the UI) behaves as expected

I'm not sure I fully agree with this statement. Clicking a button is not the important part of a component. The _consequences_ of that click are – HTTP requests, `$emit`, creating a Cookie, or whatever. And that's something totally doable without `setMethods`.


> Mocking/stubbing functions isn't an antipattern, it's a pretty important part of verifying inputs and outputs

We haven't said a word against mocking/stubbing – we actually suggest in docs to extract any complex method and spy on it.


> But the blanket suggest to "...rethink those tests" is both short-sighted and condescending

I'm sorry we made you feel this way. Truth is, however, that there's no clear migration path. That being said, we're totally open to discuss further improvements on deprecation messages 👍

dobromir-hristov:
Because I can feel this will get hairy, as people wont change their mindsets, especially as we just wont support this in Vue 3.

If you really really want to override your methods/computeds/what have you, you can do so just before you mount. The component is just an object any way.

```js
const merged = merge(Component, { methods: { myMethod: jest.fn() } })
mount(merged, options)
```

MilenaMalysh:
Even though you say that you understand importance of mocking and stubbing in the tests, I don't think you actually do. Without providing a clear way to do it you make vue-test-utils only applicable for small projects without much of frontend business logic. And in reality those projects don't require unit tests. It's really hard to screw up the "click on the button -> send an API request" functionality. You provide a lot of examples on how to test this kind of flow, but how to deal with a bunch of complex, tangled methods calling each other? Tests are really needed in the large and complex projects. These projects are usually hard to mantain and almost impossible to refactor. And the way they are structured strongly depends on the domain logic. So "to extract complex methods from the component and test them in isolation" isn't only hard to do but also absolutely unreasonable. And what if all of the methods in a component are complex? Extract all of them?

If we agree that to refactor the code in order to make the tests work isn't a good idea, let me provide an example of when stubbing can be absolutely required. Let's say we have a image file input field, a button and a file validation function which is called when user clicks on the button.  The validation function checks million different things like the size of the files array, files extensions and so on. If everything is ok it calls a function that loads the image file and show it to the user. If we won't be allowed to stub the file loading method in order to test every possible validation case we would have to call the file loading method over and over again. File loading takes some time and if we have thousands of validation cases we would absolutely unnecessary increase tests running time greatly. Especially considering that we don't really need to test js FileLoader...

Vue-test-utils claims to be a unit testing library. And even though I understand that you consider a component as a unit in reality sometimes we have to think about separate methods. The topic is quite controversial but if you could continue supporting the 'setMethods' function and the 'methods' mount option it would keep developers from re-writting their whole unit-tests code base.

lmiller1990:
VTU is utilities for testing Vue components. You can use it write unit tests, integration tests, whatever you like.

> The validation function checks million different things like the size of the files array, files extensions
> increase tests running time greatly

If you have 1 million different scenarios, you now need 1 million tests, each rendering a Vue component. This will be slow. Wouldn't it make a lot more sense to put that into a function, then test that function 1 million times?

Eg:

```
function validateFile(...lotsOfComplexArgs) {
  // if statements
  // API calls
  // whatever else
}
```

Then in your component, you would just call `validateFile(manyArguments)`. It will be much faster, too, since you don't have the overhead of a DOM, Vue, etc.

This would require a refactor. If this refactor is too significant and does't make much sense from a business point of view, just don't do it - continue using `setMethods`. It won't be removed from V1 (**Vue 2** support). Alternatively, if you don't want to use a deprecated API, you can use the user-land implementation posted above. 

The deprecation warning is there because the method will not be in **VTU v2** (which is for **Vue 3**). For those interested, it's found [here](https://github.com/vuejs/vue-test-utils-next/). It is a rewrite since the Vue internals changed significantly for Vue 3.

_______

For some context, I believe I do understand the scenario you are describing. Many of my clients have this problem - one gigantic `Component.vue`, written by some other developer they consulted to write their Vue app, whom did not write any tests. The client then realizes they have a huge, unmaintainable mess and reach out.

They want to write some tests before they continue development, since that's the only way they can do so without fear of breaking their product - makes perfect sense. Using an elaborate `mount` method with `methods`, `computed` etc is not the way to go, though. Rather than add **more** complexity, it's far better to just simplify the problem by extracting your complexity into simple functions, and testing those individually.

This is still just my opinion, derived from my experience writing, testing and maintaining Vue.js apps. If you are not satisfied with this recommendation, I have outlined a few alternatives above (continue using `setMethods`, make use of the user-land work-around).

PR #1557, opened 2020-05-24
Docs: deprecations

lmiller1990:
We should have done this for V1. Anyway, better late than ever. Let's provide a solid "this is how you migrate" docs - some of the deprecation messages are unclear or difficult to understand.

### DIFF
--- a/docs/upgrading-to-v1/README.md
+++ b/docs/upgrading-to-v1/README.md
@@ -40,7 +40,7 @@ See `isEmpty` above. Consider using [toBeVisible](https://github.com/testing-lib
 
 ### `name`
 
-Asserting against `name` encourages testing implementation details, which is a bad practice. Avoid this is possible. If you need this feature, though, you can use `vm.$options.name` for Vue components or `element.tagName` for DOM nodes. Again, consider if you really need this test - it's likely you don't.
+Asserting against `name` encourages testing implementation details, which is a bad practice. If you need this feature, though, you can use `vm.$options.name` for Vue components or `element.tagName` for DOM nodes. Again, consider if you really need this test - it's likely you don't.
 
 ### `setMethods` and `mountingOptions.methods`
 
--- a/docs/upgrading-to-v1/README.md
+++ b/docs/upgrading-to-v1/README.md
@@ -6,7 +6,7 @@ You can read the release notes for V1 [here](https://github.com/vuejs/vue-test-u
 
 ### `find`
 
-In the beta,`find` could be used to find both DOm nodes using the `querySelector` syntax, or a component via a component reference, a `ref` or a `name` option. This behavior is now split into two methods: `find` and `findComponent`.
+In beta,`find` could be used to find both DOM nodes (using the `querySelector` syntax) or a component (via a component reference, a `ref` or a `name` option). This behavior is now split into two methods: `find` and `findComponent`.
 
 If you were using this syntax:
 
--- a/docs/upgrading-to-v1/README.md
+++ b/docs/upgrading-to-v1/README.md
@@ -24,7 +24,7 @@ This method was deprecated because it tends to encourage testing implementation 
 
 ### `contains`
 
-Tests using `contains` can be replaced with `find` or `findComponent` or `get`. For example, `expect(wrapper.contains('#el')).toBe(true)` may be written as `expect(wrapper.find('#el')).toBe(true)`.
+Tests using `contains` can be replaced with `find` or `findComponent` or `get`. For example, `expect(wrapper.contains('#el')).toBe(true)` may be written as `wrapper.get('#el')`, which will throw an error if the selector is not matched. Another way to write this using `find` is `expect(wrapper.find('#el').element).toBeTruthy()`.
 
 ### `is`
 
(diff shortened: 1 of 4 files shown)


packages/shared/util.js
```diff
@@ -2,7 +2,7 @@
 import Vue from 'vue'
 import semver from 'semver'
 import { VUE_VERSION } from './consts'
-import { config } from '@vue/test-utils'
+import { config } from '../test-utils/src'
```
lmiller1990: I thought this was merged 🤔

docs/upgrading-to-v1/README.md
```diff
@@ -0,0 +1,72 @@
+### Upgrading to V1.0
+
+After a long time in beta, Vue Test Utils finally released v1.0! Some APIs that made sense during beta but turned out to be not such a good idea were deprecated. This guide will help you migrate away from the deprecation warnings you may be seeing if you were using the beta heavily.
+
+You can read the release notes for V1 [here](https://github.com/vuejs/vue-test-utils/releases) or the discussion around the deprecations [here](https://github.com/vuejs/rfcs/pull/161).
+
+### `find`
+
+In the beta,`find` could be used to find both DOm nodes using the `querySelector` syntax, or a component via a component reference, a `ref` or a `name` option. This behavior is now split into two methods: `find` and `findComponent`.
```
afontcu:
```suggestion
In beta,`find` could be used to find both DOM nodes (using the `querySelector` syntax) or a component (via a component reference, a `ref` or a `name` option). This behavior is now split into two methods: `find` and `findComponent`.
```

docs/upgrading-to-v1/README.md
```diff
@@ -0,0 +1,72 @@
+### Upgrading to V1.0
+
+After a long time in beta, Vue Test Utils finally released v1.0! Some APIs that made sense during beta but turned out to be not such a good idea were deprecated. This guide will help you migrate away from the deprecation warnings you may be seeing if you were using the beta heavily.
```
afontcu: I'd suggest making this terser

```suggestion
After a long time in Beta, Vue Test Utils finally released v1.0! We deprecated some methods that were not helpful, so you might see several deprecation warnings after upgrading. This guide will help you migrate away from them.
```

docs/upgrading-to-v1/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Finding out whether a DOM node is empty is not a Vue specific feature, and it is something that is difficult to get right. Rather than re-invent the wheel, we have decided it's better to delegate to an existing, well tested solution by default. Consider the excellent `toBeEmpty` matchers from [jest-dom](https://github.com/testing-library/jest-dom#tobeempty), for example, if you are using Jest.
+
+### `isVisible`
+
+See `isEmpty` above. Consider using [toBeVisible](https://github.com/testing-library/jest-dom#tobevisible) from `jest-dom` if you are using Jest.
+
+### `name`
+
+Asserting against `name` encourages testing implementation details, which is a bad practice. Avoid this is possible. If you need this feature, though, you can use `vm.$options.name` for Vue components or `element.tagName` for DOM nodes. Again, consider if you really need this test - it's likely you don't.
```
afontcu:
```suggestion
Asserting against `name` encourages testing implementation details, which is a bad practice. If you need this feature, though, you can use `vm.$options.name` for Vue components or `element.tagName` for DOM nodes. Again, consider if you really need this test - it's likely you don't.
```

docs/upgrading-to-v1/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+### `setMethods` and `mountingOptions.methods`
+
+By using `setMethods`, you are mutating the Vue instance - this is nothing some Vue supports, and can often hide tests that would otherwise fail. There is no straight forward replacement for this, it depends on you use case. If you have a comlex method you would like to stub out, consider moving it another file and using your test runnner's stub or mock functionality. For example, you may want to avoid an API call:
```
afontcu:
```suggestion
By using `setMethods`, you are mutating the Vue instance - this is not something Vue supports, and may lead to coupled, flaky tests.

There is no straight forward replacement for this. If you have a complex method you would like to stub out, consider moving it another file and using your test runner's stub or mock functionality.

For example, you may want to avoid an API call:
```

docs/upgrading-to-v1/README.md
```diff
@@ -0,0 +1,72 @@
...
+mount(Foo, {
+  methods: {
+    getData: () => {}
+  }
+}
+```
+
+Mock out the `axios` dependency. In Jest, for example, you can do `jest.mock('axios')`. This will prevent the API call, without mutating your Vue component.
+
+If you need more help migrating, you can join the Discord server VueLand. Alternatively, open an issue on this project's GitHub repository and someone will do their best to help you.
```
afontcu: I'd def suggest people to go to Discord instead of using GH issues as a helping tool:

```suggestion
If you need more help migrating, you can join the [VueLand server](https://chat.vuejs.org/) on Discord.
```

docs/upgrading-to-v1/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+In the beta,`find` could be used to find both DOm nodes using the `querySelector` syntax, or a component via a component reference, a `ref` or a `name` option. This behavior is now split into two methods: `find` and `findComponent`.
+
+If you were using this syntax:
+
+- `find(Foo)`
+- `find({ name: 'foo' })`
+- `find({ ref: 'my-ref' })`
+
+You should change those instances to be `findComponent`. You may continue using `find` on DOM nodes using the `querySelector` syntax.
```
afontcu:
```suggestion
Change them to be `findComponent`.

You may continue using `find` on DOM nodes using the `querySelector` syntax.
```

afontcu (commented):
This was def needed! Good point.

Added a few suggestions here and there, but looks good overall!

lmiller1990:
@afontcu thanks for this (you can see I didn't really proofread it, just threw something together).

I'll clean it up based on your suggestions today (after work).

lmiller1990:
Ok @afontcu I think we should be good?

afontcu:
> Ok @afontcu I think we should be good?

Sorry it took me a while to get back to you :) This looks cool!

afontcu (approved):

Issue #1578, opened 2020-06-12
Suggestion: un-deprecate attachToDocument


alexmiddeleer:
I feel the most common use case of `attachTo` is still `attachToDocument`, since that is necessary for testing any Vue code that utilizes `document.querySelector` and similar methods, even if run in total isolation. By un-deprecating attachToDocument, we prevent developers who are new to the library from having to hunt for the following code.
```
    const elem = document.createElement('div')
    if (document.body) {
      document.body.appendChild(elem)
    }
   wrapper = mount(Component, {
    attachTo: elem
   })
```

afontcu:
Hi! 👋  Thanks for bringing this up. While it is true that attaching to `document` is probably the most common scenario, `attachTo` provides more flexibility to cover other cases, so we'd want to stick with it.

OTOH, it would be great to add this snippet in docs. Would you like to do so? 😄

alexmiddeleer:
Sure, I can do that. Thanks for considering.

just-boris:
+1 to the undeprecation suggestion.

What is the reason for all users to write 3 lines of boilerplate, if this can be an extra built-in option? What are the risks of having both `attachTo` and `attachToDocument`?

Jeff-Duke:
+1 to the un-deprecation suggestion.
why unnecessarily increase boilerplate? why not have both methods?

lmiller1990:
Supporting both of these seems pretty reasonable. I think the original PR was to make it more robust, but we did not consider the additional boilerplate that would be introduced.

I wonder if we can have the best of both worlds? 

```js
mount(Foo, {
  attachTo: someEl
})

// what about adding support for?

mount(Foo, {
  attachTo: document
})
```

Thoughts? We get the best of both words, and it reads pretty nicely too. We could just check if value of the `attachTo` property is a `document`, and if it is, create a new element, append it to `document.body` and mount there.

@Jeff-Duke, @just-boris, @afontcu?

just-boris:
As far as I understand how `attachTo` currently works, it is replacing the DOM node with the content from Vue. I tried using it with `mount(Foo, {  attachTo: document.body})` and it broke, because `document.body` becomes null for subsequent tests.

Having a separate explicit method is preferred because it is more clear than the difference between ` attachTo: document.body` and ` attachTo: htmlElement` behavior.

afontcu:
[...]

PR #1580, opened 2020-06-14
docs update for #1578

### LINKED ISSUE #1584: attachToDocument deprecated and no documented substitute?
### Subject of the issue

attachToDocument was deprecated, in favor of attachTo, which is supposed to be more flexible, by allowing any css selector as the mount point.

But what is the actual replacement for attachToDocument: true? attachTo: document.documentElement yields HierarchyRequestError: The operation would yield an incorrect node tree.

Seems like the functionality of attachToDocument was deprecated in favor of something that does not substitute it, but only provides complementary behaviour.

Note that, besides the hierarchy problem, attachTo should be acompanied by wrapper.destroy() at the end. The thing is that this DESTROYS THE NODE to which the wrapper was attached. So, I can attach the element to the body, but the body is going to be destroyed in the call.

What I'm currently doing is adding a dumb routine to each test suite that needs the wrapper to be attached to the dom:
```

    var node = document.createElement("div");
    node.setAttribute('id', 'app')
    document.body.appendChild(node);

    const wrapper = mount(List, {
      attachTo: '#app'
    }

    ...

    wrapper.destroy() 
```

Needless to say, unless I'm missing something, it would be way easier to not have attachToDocument deprecated.

alexmiddeleer:
Closes #1584

- [ X ] It's submitted to the `dev` branch.
- [ X ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [  ] All tests are passing: https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md#development-setup
- [  ] New/updated tests are included

Note that prettier found some issues in:
docs/zh/guides/common-tips.md
docs/zh/guides/getting-started.md

But I did not touch these files.

image

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -323,7 +323,7 @@ wrapper.destroy()
 - default: `false`
 
 ::: warning Deprecation warning
-`attachToDocument` is deprecated and will be removed in future releases. Use [`attachTo`](#attachto) instead. Example:
+`attachToDocument` is deprecated and will be removed in future releases. Use [`attachTo`](#attachto) instead. For example, if you need to attach the component to the document.body:
 
 ```js
 const elem = document.createElement('div')
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -323,7 +323,18 @@ wrapper.destroy()
 - default: `false`
 
 ::: warning Deprecation warning
-`attachToDocument` is deprecated and will be removed in future releases. Use [`attachTo`](#attachto) instead.
+`attachToDocument` is deprecated and will be removed in future releases. Use [`attachTo`](#attachto) instead. Example:
+
+```js
+const elem = document.createElement('div')
+if (document.body) {
+  document.body.appendChild(elem)
+}
+wrapper = mount(Component, {
+  attachTo: elem
+})
+```
+
 :::
 
 Like [`attachTo`](#attachto), but automatically creates a new `div` element for you and inserts it into the body.


alexmiddeleer:
I attempted to fix the pre-existing prettier errors in https://github.com/vuejs/vue-test-utils/pull/1581

docs/api/options.md
```diff
@@ -323,7 +323,18 @@ wrapper.destroy()
 - default: `false`
 
 ::: warning Deprecation warning
-`attachToDocument` is deprecated and will be removed in future releases. Use [`attachTo`](#attachto) instead.
+`attachToDocument` is deprecated and will be removed in future releases. Use [`attachTo`](#attachto) instead. Example:
```
afontcu: Could we add a very simple explanation of what does the example do? Say, "For example, if you need to attach the component to the `document.body`" or similar

alexmiddeleer:
Please let me know if there's anything you're still waiting on me for.

afontcu:
Looks good! There are some issues with Prettier. Would you mind running `yarn format` and push again? (I think they might be unrelated to your changes)

alexmiddeleer:
@afontcu I ran `yarn format`, but it resulted in the same code changes that I addressed in https://github.com/vuejs/vue-test-utils/pull/1581. Those changes broke a desired convention in the /zh language docs.

afontcu:
Hm, weird. It def looks like it is related to `zh` docs. Will take a look, but otherwise your PR looks fine 👍  gonna try to find out what's going on and then merge it up.

thanks!! 🤗

afontcu:
Closing in favor of #1611! thanks for this :)

Issue #1597, opened 2020-06-28
Composition API provide/inject with symbols are not working


Babbafett:
I am not able to provide/inject my composition function in my component during a test.

I have the following provide function to provide my function to my main component:
```
export interface AuthContext {
  redirectToCachedRoute(): Promise<void>;
  login(): Promise<void>;
  logout(): Promise<void>;
  getAccessToken(): Promise<CognitoAccessToken>;
  getUserInfo(): Promise<CognitoUserInfo | any>;
  getPermissions(): Promise<string[]>;
  cacheRoute(route: Route | string): Promise<void>;
}

const AuthContextKey: InjectionKey<AuthContext> = Symbol('auth');
export function provideAuth() {
.......
  const auth: AuthContext = {
    redirectToCachedRoute,
    login,
    logout,
    getAccessToken,
    getUserInfo,
    getPermissions,
    cacheRoute,
  };

  provide<AuthContext>(AuthContextKey, auth);

  return auth;
}
```
I have the following function to inject my function to my child components:
```
export function useAuth() {
  const context = inject<AuthContext>(AuthContextKey);

  if (context == null) {
    throw new Error('provideAuth call missing in main.ts');
  }

  return context;
}
```

Both functions are in the same file and have access to my injection key.
The provideAuth function will be called within the setup function of my main.ts:
```
new Vue({
  setup() {
    const auth = provideAuth();
    return {};
  },
  render: (h) => h(App),
}).$mount('#app');
```

### Subject of the issue

Composition API: Provide/Inject with InjectionKeys are not working

### Steps to reproduce

1. Create a new composition function with a provide and a use function following the recommended instructions of the RFC api reference --> https://composition-api.vuejs.org/api.html#dependency-injection

2. Use the use function at your component and try to mount the component.

### Expected behaviour

Child component can inject the object from the use function.

### Actual behaviour

inject command failed because it can't find the symbol injection key at the child component.
Exception "throw new Error('provideAuth call missing in main.ts');" of my inject function is raised.

### Possible Solution

I tried to provide my function with the setup mounting options and also tried to provide it via the provide options, but both are not working:
```
describe('general-content-page.vue', () => {
  const localVue = createLocalVue();
  localVue.use(CompositionApi);
  const key: InjectionKey<AuthContext> = Symbol('auth');
  const mock: AuthContex t= {
     redirectToCachedRoute: jest.fn(),
     login: jest.fn(),
     logout: jest.fn(),
[...]

PR #1598, opened 2020-06-30
chore: repro problem with provide/inject

lmiller1990:
Fail to reproduce error for #1597

### DIFF
--- a/test/specs/mounting-options/provide.spec.js
+++ b/test/specs/mounting-options/provide.spec.js
@@ -1,11 +1,12 @@
-import { config } from '@vue/test-utils'
+import Vue from 'vue'
+import { config, mount } from '@vue/test-utils'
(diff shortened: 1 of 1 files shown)


Stoom (approved):
Looks good 👍

test/specs/mounting-options/provide.spec.js
```diff
@@ -201,6 +202,33 @@ describeWithShallowAndMount('options.provide', mountingMethod => {
     }
   )
 
+  it('provides with a symbol', () => {
```
Stoom, line 205: You'll want to wrap this in the following for compact:
```ts
 itDoNotRunIf(
    !injectSupported,
```

lmiller1990:
This was just demo-ing a bug (that isn't a bug, I don't think

PR #1610, opened 2020-07-07
Docs: Create "Installation" section out of Guide

### LINKED ISSUE #1606: Unclear why documentation uses wrapper.element for snapshots
### Documentation Feedback

**Provide a brief summary of what you would like to see changed in our documentation.**
Naively, for as long as I've been using this library I've been doing snapshot testing via
```javascript
expect(wrapper).toMatchSnapshot()
```
however, the documentation without reasons shows
```javascript
expect(wrapper.element).toMatchSnapshot()
```
[while vuetify shows](https://vuetifyjs.com/en/getting-started/unit-testing/#mocking-vuetify)
```javascript
expect(wrapper.html()).toMatchSnapshot()
```

It's not clear to me what the exact differences between the three are.

**Feel free to provide any suggestions of content or examples you’d like us to include.**
It would be nice to explain either the pros and cons of the different methods, or to just endorse one and say why explicitly. 

**Affected documentation page:** Insert a link to the affected page
https://vue-test-utils.vuejs.org/guides/#testing-single-file-components-with-jest

afontcu:
This PR adds an Installation section in docs:

image

---

There are a lot of low-hanging fruits in VTU v1 docs, and I plan on tackling them in the future. Meanwhile, this PR:

* Creates a Installation section.
* Reduces the amount of content in the section. As Lachlan [said](https://github.com/vuejs/vue-test-utils/pull/1610#issuecomment-655173630), keeping these kind of docs up to date is a burden and none of the other Vue projects contain this kind of low level tooling info.
* Recommends Jest as the default way to run VTU.
* Removes snapshot testing section. This is opinionated, so I'm happy to discuss it further. Component snapshot testing creates a totally coupled tests that break on every iteration. Thus, snapshot testing should not be highlighted that much in VTU docs.
* Adds Semantic Versioning section, just as Vue docs [does](https://vuejs.org/v2/guide/installation.html#Semantic-Versioning).

If we agree on these changes, then we'd need to update docs in `ru`, `zh`, and `ja`. I can do my best by copying and pasting, but I'd need someone to review changes! 😇 

---

Also this PR closes #1606


lmiller1990:
Thanks, we should definitely keep this section lean and focused. I can fix the Japanese docs this week sometime.

Potentially unpopular opinion, but more docs (especially around tooling) means more maintenance required. I think we should straight up delete:

- Testing Single-File Components with Mocha + webpack (just recommend jest instead)
- Testing Single-File Components with Karma ( 🤷‍♂️  idk who uses this, they can figure it out)
- Testing in Node.js Without a Build Step (not that common - "advanced" use case. all you need to do is import a few libraries anyway. Could include a paragraph instead of an entire page, something like "ensure jsdom is available before import vue test utils)

Maintaining these is hard, dependencies constantly changed. None of the other Vue projects contain this kind of low level tooling info - at best, they give you a paragraph or two.

dobromir-hristov:
Looks good to me

afontcu:
> Potentially unpopular opinion, but more docs (especially around tooling) means more maintenance required. I think we should straight up delete:

Not that unpopular over here, hahaha. I'm up for it. Actually, while moving things around I thought we should have a single page _à la_ vuejs.org (https://vuejs.org/v2/guide/installation.html), with some recommendations (Jest).

lmiller1990:
Great!

lmiller1990 (approved):

lmiller1990:
A little worried out dated translations... 🤔 feel free to hit merge, don't think the translations should be a blocker.

afontcu:
> A little worried out dated translations

Me too… we can always point to this PR if someone is willing to update them!

Jinjiang:
LGTM about the zh part. Thanks :-)

PR #1611, opened 2020-07-08
docs: add suggestion to deprecated attachToDocument

### LINKED ISSUE #1584: attachToDocument deprecated and no documented substitute?

afontcu:
Closes #1584

Just a mirror of #1584, which fails on Prettier unexpectedly (AFAIK).

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -9,20 +9,22 @@ These options will be merged with the component's existing options when mounted 
 [See other options for examples](#other-options)
 :::
 
-- [`context`](#context)
-- [`data`](#data)
-- [`slots`](#slots)
-- [`scopedSlots`](#scopedslots)
-- [`stubs`](#stubs)
-- [`mocks`](#mocks)
-- [`localVue`](#localvue)
-- [`attachTo`](#attachto)
-- [`attachToDocument`](#attachtodocument)
-- [`propsData`](#propsdata)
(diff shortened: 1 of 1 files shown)


afontcu:
Looks like build is passing. I'm going to merge this.

All credit goes to @alexmiddeleer ;)

PR #1613, opened 2020-07-13
Revamp Getting Started

afontcu:
Revamp the `Getting Started` section to match docs for VTU2.

Besides that, this PR:
- Adds reference to Vue Testing Lib.
- Reduces the prominence of Shallow Mounting. 
- Fixes some headings and styling here and there.

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -21,7 +21,7 @@ const MessageComponent = {
 test('displays message', () => {
   // mount() returns a wrapped Vue component we can interact with
   const wrapper = mount(MessageComponent, {
-    props: {
+    propsData: {
       msg: 'Hello world'
     }
   })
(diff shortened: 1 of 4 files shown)


docs/guides/getting-started.md
```diff
@@ -2,141 +2,75 @@
...
+// The component to test
+const MessageComponent = {
+  template: '<p>{{ msg }}</p>',
+  props: ['msg']
+}
+
+test('displays message', () => {
+  // mount() returns a wrapped Vue component we can interact with
+  const wrapper = mount(MessageComponent, {
+    props: {
```
lmiller1990: I think this is `propsData` in VTU V1 🤔 

We should backport `props` perhaps.

lmiller1990 (approved):
One small comment, great improvements here.

I wonder if we can share some of the guides in VTU and VTU next? This would be ideal. 🤔  At least the high level ones should be very similar (if not identical?)

docs/guides/getting-started.md
afontcu: dang, you're right. fixed.

and yeah, maybe porting back `props` to match VTU 2 is a sensible idea 🤔

PR #1634, opened 2020-08-04
Fix #1633 – feat(array wrapper): return ErrorWrapper when index not found with .at()

### LINKED ISSUE #1633: at() should return an ErrorWrapper when wrappers is empty
**Situation**

The component (`Component`) has children components (`Child`), one and one only of type `X` that is conditionally rendered.
I need to test this component in severaI cases and so it's common to retrieve  the component like this:

```javascript
const findX = wrapper.findAll(Child).filter(isX).at(0);
```

**Complication**

I know the element is rendered conditionally, so I want to test the case when this element does not exist. But when I try:

```javascript
expect(findX().exists()).toBe(false);
```
surprisingly enough, I get the following error `no item exists at 0`.

I know the option would be not to add `at(0)` to my `findX()` method. But that's not what I would expect form a method that applies to an Array, for the following reasons:

#### A) The signature
`at()`  is defined as `at(index: number): Wrapper<V>;`. Throwing clearly does not mean the signature is wrong but what I argue is that the error is thrown for the wrong reason.

#### B) The semantics
If I change the case slightly and now I have only one `Child` component, I could write `findX()` as `findComponent(Child)`. Testing the case where it is not rendered I would be able to use `exists()` properly. It seems to me there is nothing inherently wrong though in sticking to the `wrapper.findAll(Child).filter(isX).at(0)` implementation also when I only have one `Child` component. Therefore I would expect the same result, that is a `BaseWrapper` to which I could ask if the element exists or not;

#### C) The similarity with `filter()`
Not only my `findX()` implementation is contingent but also `at()`'s one. It could easily be implemented as a filter function on the given `index` as they're both trying to individuate an element in an array based on some condition:

```javascript
wrapper.filter((_, index) => index === 0)
```
In this case, if the predicate fails to verify for every element, I would get an empty array. 

Returning an empty array from `at()` is not what we want – `at()` seems to be somehow similar to `Array.prototype.find` which would, in the case explored here, return `undefined`.

**Solution**

Given `find()` is already taken and that we do not want to return `undefined`, my proposal is to return an `ErrorWrapper` on which I could call the `exists()` method as easily as when I do it with `find()`.

The API of `at()` would not change, instead its internal implementation would.
It could be something along the lines of:

```javascript
at(index) {
    const normalizedIndex = index < 0 ? this.length
    return this.wrappers[normalizedIndex] || new ErrorWrapper('')
}
```

angelogulina:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

(probabaly?)

**Other information:**

### DIFF
--- a/docs/api/wrapper-array/at.md
+++ b/docs/api/wrapper-array/at.md
@@ -2,6 +2,7 @@
 
 Returns `Wrapper` at `index` passed. Uses zero based numbering (i.e. first item is at index 0).
 If `index` is negative, indexing starts from the last element (i.e. last item is at index -1).
+When none is found, returns an `ErrorWrapper`.
 
 - **Arguments:**
 
@@ -23,4 +24,7 @@ expect(secondDiv.is('div')).toBe(true)
 
 const lastDiv = divArray.at(-1)
 expect(lastDiv.is('div')).toBe(true)
+
+const nonExistentDiv = divArray.at(1000)
+expect(nonExistentDiv.exists()).toBe(false)
 ```
--- a/packages/test-utils/src/wrapper-array.js
+++ b/packages/test-utils/src/wrapper-array.js
@@ -3,6 +3,7 @@
 import type Wrapper from './wrapper'
 import type VueWrapper from './vue-wrapper'
 import { throwError } from 'shared/util'
+import ErrorWrapper from './error-wrapper'
 
 export default class WrapperArray implements BaseWrapper {
   +wrappers: Array<Wrapper | VueWrapper>
@@ -23,14 +24,9 @@ export default class WrapperArray implements BaseWrapper {
     })
   }
 
-  at(index: number): Wrapper | VueWrapper {
+  at(index: number): Wrapper | VueWrapper | ErrorWrapper {
     const normalizedIndex = index < 0 ? this.length + index : index
-    if (normalizedIndex > this.length - 1 || normalizedIndex < 0) {
-      let error = `no item exists at ${index}`
-      error += index < 0 ? ` (normalized to ${normalizedIndex})` : ''
-      throwError(error)
-    }
-    return this.wrappers[normalizedIndex]
+    return this.wrappers[normalizedIndex] || new ErrorWrapper('')
   }
 
   attributes(): void {
--- a/test/specs/wrapper-array/at.spec.js
+++ b/test/specs/wrapper-array/at.spec.js
@@ -26,33 +26,12 @@ describeWithShallowAndMount('at', mountingMethod => {
     expect(first.classes()).to.contain('index-first')
   })
 
-  it('throws error if no item exists at index', () => {
-    const index = 2
+  it('returns ErrorWrapper at index when not found', () => {
     const TestComponent = {
-      template: '<div><p /><p class="index-1"/></div>'
+      template: '<div></div>'
     }
-    const message = `[vue-test-utils]: no item exists at ${index}`
-    expect(() =>
-      mountingMethod(TestComponent)
-        .findAll('p')
-        .at(index)
-    )
-      .to.throw()
-      .with.property('message', message)
-  })
-
-  it('throws error if no item exists at negative index', () => {
-    const index = -3
-    const TestComponent = {
-      template: '<div><p /><p class="index-1"/></div>'
-    }
-    const message = `[vue-test-utils]: no item exists at -3 (normalized to -1)`
-    expect(() =>
-      mountingMethod(TestComponent)
-        .findAll('p')
-        .at(index)
-    )
-      .to.throw()
-      .with.property('message', message)
+    const all = mountingMethod(TestComponent).findAll('p')
+    const nonExistent = all.at(0)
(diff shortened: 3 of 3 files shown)


lmiller1990:
Hm, this is probably a breaking change. What do we do at the moment? Just undefined? Is there a strong reason for this breaking change?

angelogulina:
Hi @lmiller1990 and thanks for your kind reply. Yes, I think you're correct that it is a breaking change.

My mistake not to have linked explicitly the Issue where I develop the case with the compelling (at least to me 😂) reasons for wanting that.

Please, have a look [here](https://github.com/vuejs/vue-test-utils/issues/1633) and feel free to ping me for any doubt or question!

lmiller1990:
I see. The issue says "The API of at() would not change, instead its internal implementation would". Actually, I think this is a change in the public API (behavior is different).

This is probably a sensible change to make though. I have no idea if/how many people rely on this existing behavior. I would like some more opinions (extremely hard to get those, though). @afontcu  @cexbrayat any opinion?

In VTU v2, we don't even have `WrapperArray.at` - it just returns an array of `Wrapper`. Another significant breaking change we should document. @AngeloGulina do you find this API good? Do you have an opinion on just returning an Array of Wrapper instead of a WrapperArray type with `at` accessors?

afontcu:
This definitely looks like a (small) breaking change, and even though it makes sense, I'm not 100% it is worth the hassle 🤔 

Could we make it opt-in through a config parameter? Call the option `throwAtError` (or whatever), and maybe other methods could benefit from it

angelogulina:
Thanks again for the reply @lmiller1990.

#### v1

> I see. The issue says "The API of at() would not change, instead its internal implementation would". Actually, I think this is a change in the public API (behavior is different).

The current signature of `WrapperArray.at` is defined as `at(index: number): Wrapper<V>;` so the API (in this strict sense) would not change although I agree with you that there could be people relying on the current behaviour of it and I do not wish to break this lovely library for anyone 😄 

We might have two alternatives,

1. introducing a new method that is consistent with how other methods of `WrapperArray` work, namely returning the `ErrorWrapper` when a search is not successful. [`nth`](https://lodash.com/docs/4.17.15#nth) is what is generally used but I fear it might be confusing – why having two same methods that behave differently? `take` would be nicer but I the same fear applies, adding the fact that some libraries have a different API for it.

1. I'd be happy to reduce the scope of the issue/change. Conventionally, libraries that operates on `Array` have a `head` method, which I think could cover many cases. I'd propose [`first`](https://underscorejs.org/#first) as it sounds nicer 🙈  and just for the sake of it, a `last` one.

1. add a config param to `at` as suggested [here](https://github.com/vuejs/vue-test-utils/pull/1634#issuecomment-673289646)

I'd be happy to update the relevant issue and the MR if we find an agreement.

#### v2 and `at`
> Do you have an opinion on just returning an Array of Wrapper instead of a WrapperArray type with at accessors?

Thanks for asking that!

Moving from `WrapperArray` to `Array` could bring some benefits, that is one could simply use `Array` methods and/or utils to describe their test cases. In this scenario, `at` doesn't really find a place.

(As a note, I found the introduction of `BaseWrapper.exists` to be a very useful concept  with an [eminent ancestry](https://en.wikipedia.org/wiki/Existence#Existence_in_the_wide_and_narrow_senses) – but I'm fine with giving up on it for some situations, if removing `WrapperArray` brings simplification.)

#### v2 and `get`
Related to v2, I have a question about the introduction of `get`.

Move to [here](https://github.com/vuejs/vue-test-utils-next/issues/176) because it was getting too big.

angelogulina:
Thanks for the feedback @afontcu!

I just edited my reply which contains also some alternative proposals. I like yours as well, so I'm gonna add it to the list.

lmiller1990:
I am not a fan of an API like `head` or `tail`. It feels like it just adds more to learn: imo, ideally, we should just return an array of wrappers - any JS developer knows you can do `arr[0]`, `arr[1]` etc - even `at` is confusing, does it start at 0? 1? 

If I had to choose I  would go with an additive feature, suggested above... that said, the concept of WrapperArray doesn't exist in V2 (yet, and I don't think it should for the reasons outlined above) so I am not sure there is any real merit introducing it here for a small number of codebases that would benefit from it.

Not going to lie, I do feel a little bad to turn down a feature request that is so well thought about and delivered, but it's really hard to justify any kind of breaking change when the rest of the core libs have been so stable for so long, and with Vue 3 coming out, I don't expect people will want to deal with any kind of break change at this point.

On the flip side, if you feel strongly about something, now is the time to get in done in V2, before we move to a release candidate!

angelogulina:
@lmiller1990 thanks so much for the answer again 🎉 

I see your point and there's no problem in turning down the feature – I understand the concerns.
I filled an Issue for the v2 on which we could maybe elaborate a bit more on the throwing error part (if you feel like).

Gonna close this one.

PR #1701, opened 2020-09-29
docs: update vue-router documentation

blake-newman:
Update documentation to explain differences between using `localVue` and `mocks` and provide examples of the use case for each. 

A common gotcha with integration testing (which don't stub children) is when using the `mocks` approach the router instance is not available on child components

### DIFF
--- a/docs/guides/using-with-vue-router.md
+++ b/docs/guides/using-with-vue-router.md
@@ -41,16 +41,19 @@ shallowMount(Component, {
 ### Installing Vue Router with localVue
 
 ```js
-import { shallowMount, createLocalVue } from '@vue/test-utils'
+import { mount, createLocalVue } from '@vue/test-utils'
 import VueRouter from 'vue-router'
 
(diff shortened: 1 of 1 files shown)


jamiejones93 (approved):
🎉

afontcu:
Neat!

Looks like some formatting is a bit off:

```
Checking formatting...
docs/guides/using-with-vue-router.md
Code style issues found in the above file(s). Forgot to run Prettier?
```

Would you mind running Prettier and push again?  (We should def ease this process with a precommit hook or similar…)

blake-newman:
@afontcu or i was lazy and committed via github :laughing:

lmiller1990:
I never realized `mocks` was only for the component in question - I thought it would be available to all the children, too 🤔 

It sounds like this should be the default behavior.

afontcu (approved):

blake-newman:
@lmiller1990 I have created a new issue. Still believe best to document the current behaviour.

PR #1705, opened 2020-10-01
fix(setData): allow empty objects to be set fix #1704

### LINKED ISSUE #1704: setData with an empty object doesn't get set
### Subject of the issue

If you call `wrapper.setData({ someKey: {} })`, it doesn't set the `someKey` to `{}`

### Steps to reproduce

this test doesn't pass

```
const wrapper = mount({
  data() {
    return {
      someKey: { someValue: true },
    };
  },
  render(h) {
    return h('span');
  },
});

expect(wrapper.vm.$data).toEqual({ someKey: { someValue: true } });

wrapper.setData({ someKey: {} });

expect(wrapper.vm.$data).toEqual({ someKey: {} });
```

### Expected behaviour

This test passes

### Actual behaviour

The test doesn't pass, it still has the previous 

### Possible Solution

In here: https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/recursively-set-data.js I think it might make sense to see if the object has keys at all, if it doesn't have keys, you could set the value.

Haroenv:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

if someone relied on empty objects not updating, that now no longer makes sense

**Other information:**

This is a tentative fix to show one possible solution, I can clean it up if you think it makes sense

### DIFF
--- a/packages/test-utils/src/recursively-set-data.js
+++ b/packages/test-utils/src/recursively-set-data.js
@@ -5,7 +5,7 @@ export function recursivelySetData(vm, target, data) {
     const val = data[key]
     const targetVal = target[key]
 
-    if (isPlainObject(val) && isPlainObject(targetVal)) {
+    if (isPlainObject(val) && isPlainObject(targetVal) && Object.keys(val).length > 0) {
       recursivelySetData(vm, targetVal, val)
     } else {
       vm.$set(target, key, val)
--- a/packages/test-utils/src/recursively-set-data.js
+++ b/packages/test-utils/src/recursively-set-data.js
@@ -5,7 +5,11 @@ export function recursivelySetData(vm, target, data) {
     const val = data[key]
     const targetVal = target[key]
 
-    if (isPlainObject(val) && isPlainObject(targetVal) && Object.keys(val).length > 0) {
+    if (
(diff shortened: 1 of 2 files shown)


lmiller1990:
Seems like some tests are failing. LMK if I can help out with anything.

Haroenv:
I didn't update the tests yet, but I'm more wondering if my approach makes sense, and whether this is a problem that should even be fixed @lmiller1990 :)

lmiller1990:
I guess so, if Vue supports this (which it does, right, settings something in data to `{}` will not lose reactivity), it seems like VTU probably should too.

Haroenv:
Tests are passing now, although I didn't add a new test for this yet :)

lmiller1990:
Great. Can you add a test for this, too? Then we can merge it.

Haroenv:
voilà @lmiller1990, new test added :)

afontcu (approved):
🎉

Haroenv:
Thanks for your reactivity all 👍

PR #1716, opened 2020-10-15
feat: add abstract property to the core property of the component

ustbhuangyi:
Since the Transition component is an abstract component, and we use stub to simulate it. If the abstract property is not copied, and if the Transition component is used inside the tested component, the parent-child relationship of the internal component tree will be incorrect.


lmiller1990:
Hm, ok! Not super clear on this - can you share an example of a test that is incorrect before this change?

ustbhuangyi:
@lmiller1990 Here is the example repo: https://github.com/ustbhuangyi/vue-test-utils-demo

lmiller1990:
I will do a release this weekend

Issue #1702, opened 2020-09-30
IgnoredElements doesn't seem working when running the unit test


aryaroudi:
### Using IgnoredElements seems being ignored while running the test

I'm using a 3rd party library which loads some `custom elements` for my app which causing some error and warnings in my unit tests. As there isn't an article to explain how to combine your unit test with non-vue components AKA custom elements, I followed the documentation to ignore these components while testing but it seems there is a problem that Vue-Test-Utils  is ignoring the `IgnoredElements` option.

### Steps to reproduce

use a 3rd party or a Stencil custom element in your project and then write a unit test for it, while running the test the error `Unknown custom element` will show up

### Expected behaviour

The Custom Element should be ignored while the testing.

### Actual behaviour

Still getting the warning `[Vue warn]: Unknown custom element: ....`
### Possible Solution

This option should work if it sets in the `spec` file.

```typescript

const localVue = createLocalVue();
localVue.use(Vuex);
localVue.config.ignoredElements = ['custom-button'];  <----- here I set the option

describe('ticketDetails.vue', () => {
    let wrapper: Wrapper<TicketDetails>;
    let wrapprContact: Wrapper<Contacts>;
    let vuexStore: Store<any>;
    var $: any;
    beforeEach(async () => {
        vuexStore = new Vuex.Store({
            state: {},
        });
        vuexStore.registerModule(STORE_KEY, ticketingStore);
        wrapper = shallowMount(TicketDetails, {
            store: vuexStore,
            localVue,
            mocks: {
                wrapprContact,
            },
        });
        wrapper.setData({
            ticket: DETAILS.repairTicket,
            authConditions: true,
        });
        const authorization = wrapper.find('#button-authorization');
        authorization.trigger('click');
    });
....
```

lmiller1990:
Could you try stubbing the element as a workaround potentially with `stubs`?

aryaroudi:
> Could you try stubbing the element as a workaround potentially with `stubs`?

A workaround is to `import Vue from 'Vue'` and then use the ignore element, it worked but I will give `stubs` a try too

lmiller1990:
Ok this should be easy to support. You [would update here](https://github.com/vuejs/vue-test-utils/blob/dev/packages/test-utils/src/config.js) and somewhere [around here](https://github.com/vuejs/vue-test-utils/blob/dev/packages/shared/create-local-vue.js#L36) and that should be enough. Would you like to make a PR?

[...]

PR #1718, opened 2020-10-16
fix(config): merging user config with default config of Vue

aryaroudi:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
#1702

### DIFF
--- a/packages/shared/create-local-vue.js
+++ b/packages/shared/create-local-vue.js
@@ -32,8 +32,8 @@ function _createLocalVue(
     }
   })
 
-  // config is not enumerable
-  instance.config = cloneDeep(Vue.config)
+  // merge the user config with default VUE config
+  instance.config = { ...Vue.config, ...config }
 
   // if a user defined errorHandler is defined by a localVue instance via createLocalVue, register it
   instance.config.errorHandler = config.errorHandler || Vue.config.errorHandler
--- a/packages/test-utils/types/index.d.ts
+++ b/packages/test-utils/types/index.d.ts
@@ -169,7 +169,7 @@ interface VueTestUtilsConfigOptions {
   deprecationWarningHandler?: Function
 }
 
-export declare function createLocalVue (): typeof Vue
+export declare function createLocalVue (config?: Record<string, any> | null | undefined): typeof Vue
 export declare let config: VueTestUtilsConfigOptions
 
 export declare function mount<V extends Vue> (component: VueClass<V>, options?: ThisTypedMountOptions<V>): Wrapper<V>


packages/shared/create-local-vue.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 import Vue from 'vue'
-import cloneDeep from 'lodash/cloneDeep'
+import { cloneDeep, merge } from 'lodash/cloneDeep'
```
afontcu: Hm, I'm not sure you can import lodash packages from `lodash/cloneDeep` like this, but rather

```js
import { cloneDeep, merge } from 'lodash'
```

packages/shared/create-local-vue.js
```diff
@@ -33,7 +33,7 @@ function _createLocalVue(
   })
 
   // config is not enumerable
-  instance.config = cloneDeep(Vue.config)
+  instance.config = merge({}, cloneDeep(Vue.config), config)
```
afontcu: Could you add a test to assert / document this feature? 🙌

packages/shared/create-local-vue.js
lmiller1990: I think this will not tree shake well - it will include *all* of `lodash` in the final build (maybe? lodash pre-dates tree shaking).

I think what we should be doing is this:

```
import cloneDeep from 'lodash/cloneDeep'
import merge from 'lodash/merge'
```

The best way to verify this is just doing all the different imports and seeing how the final bundle changes

packages/shared/create-local-vue.js
lmiller1990: I don't even know if we need to use `merge` here - can't we just do:

```
instance.config = {...Vue.config, ...config)
```

Or something?


```

lmiller1990 (commented):
Added some comments. We do a test for this.

packages/shared/create-local-vue.js
afontcu: sure! I think it depends on bundlers, too. Also, VTU is used in dev time (not production), so build size isn't _that_ important, but if we're faced with to similar alternatives, one being lighter… let's go with it :)

aryaroudi: good point @afontcu I didn't pay attention to that

packages/shared/create-local-vue.js
aryaroudi: @lmiller1990 it depends! we can do that but `Lodash` merge is faster than spread operator and use less memory. I don't mind to use any of them, let me know what do you prefer

more info: https://www.measurethat.net/Benchmarks/Show/2978/0/lodash-merge-vs-objectassign-vs-spread

aryaroudi:
> Added some comments. We do a test for this.

I updated the comment of that line to make more sense

afontcu:
Hi! We're still missing some unit tests that covers those lines

aryaroudi:
> Hi! We're still missing some unit tests that cover those lines

Hey! I thought you guys going to add the unit test. I will try to add a test for this change

lmiller1990:
Are you able to give this a try? I can guide if needed.

aryaroudi:
> Are you able to give this a try? I can guide if needed.

yes sure, let me know what do I need to do?

lmiller1990:
That's kind of the problem - what bug does this fix? What test are you able to write that you couldn't write previously?

aryaroudi:
> That's kind of the problem - what bug does this fix? What test are you able to write that you couldn't write previously?

this was the issue if you remember. 

https://github.com/vuejs/vue-test-utils/issues/1702

lmiller1990:
This fix does not appear to work - I still get the warning about the custom elements. The only was I can get it to swallow the warning is by doing Vue.config.ignoredElements = ..., and mutating the global Vue instance. 🤔

aryaroudi:
> This fix does not appear to work - I still get the warning about the custom elements. The only was I can get it to swallow the warning is by doing Vue.config.ignoredElements = ..., and mutating the global Vue instance. 🤔

I couldn't test it on my local. it is probably because we were using the previous version (BETA) this worked `Vue.config.ignoredElements` but isn't that better we be able to set the ignore element in `localvue`?

lmiller1990:
What you are proposing makes perfect sense - but the current PR here does not seem to solve this problem, I still get the warning for unknown components. So we need to find a way to make sure the options are correctly merged. What you did seems like a step in the right direction but not quite there.

lmiller1990:
Will re-open if there is some movement on this PR.

PR #1723, opened 2020-10-23
Fix two bugs in autodestroy

pieterdd:
My colleague @kjoep did some bugfixes. I'm filling out the form on his behalf to get it merged into the mainline.

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch.
- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number) - _It's not related to an existing issue thread as far as we know_
- [x] All tests are passing: https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md#development-setup - _Awaiting build_
- [x] New/updated tests are included - _The TypeScript change is not testable. Anything else the author should cover?_

**Other information:**

PR #1724, opened 2020-10-23
feat(test-utils): add types for auto destroy methods

blake-newman:
Add types for auto destroy methods, the hook type is typed loosely to be agnostic to testing frameworks.


pieterdd:
~Related: https://github.com/vuejs/vue-test-utils/pull/1723~ Nevermind, that only covers one of the two methods.

lmiller1990:
Thanks. I will do a release this weekend

PR #1750, opened 2020-12-12
fix: does not pass data provided to mount to wrapper component

xanf:
#1661 [introduced](https://github.com/vuejs/vue-test-utils/pull/1661/files#diff-5d7b4788ebdcd7703b1df7d0bb64f8dd6a843c29459bb4fc134364530368af73R131) copying of passed options to parent wrapper. However it fails in a weird way if you are using `data` option for `mount`/`shallowMount` with external variable.

i.e.
```js 
    const defaultData = { foo: 3 }
    const wrapper = mountingMethod(TestComponent, { data: () => defaultData })
```
In this case `data` function will be invoked two times: 

* one for `Parent` component (wrapper for our `TestComponent`) - at this point `defaultData` will be "poisoned" (will become reactive, and observed by `Parent` component for changes)
* one when `Parent` wrapper will be mounted, causing `TestComponent` instance to mount - at this point Vue will complain 

```
[Vue warn]: Avoid adding reactive properties to a Vue instance or its root $data at runtime - declare it upfront in the data option.
```

because `defaultData` is treated as `$data` for Parent wrapper component already

In general, we do not need setting `data` on parent wrapper, so I just eliminate this propety before copying

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [-] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**

### DIFF
--- a/flow/options.flow.js
+++ b/flow/options.flow.js
@@ -3,6 +3,7 @@ declare type Options = {
   attachToDocument?: boolean,
   attachTo?: HTMLElement | string,
   propsData?: Object,
+  data?: Object | Function,
   mocks?: Object,
   methods?: { [key: string]: Function },
   slots?: SlotsObject,
@@ -20,6 +21,7 @@ declare type Options = {
 declare type NormalizedOptions = {
   attachTo?: HTMLElement | string,
   attachToDocument?: boolean,
+  data?: Object | Function,
   propsData?: Object,
   mocks: Object,
   methods: { [key: string]: Function },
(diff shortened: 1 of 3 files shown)


lmiller1990 (approved):
hm, interesting! I guess using an external variable for data is only really something you do in tests?

Either way, seems fine!

lmiller1990:
It's out

https://github.com/vuejs/vue-test-utils/releases/tag/v1.1.2

PR #1786, opened 2021-02-10
fix(create-local-vue): fix max range stack size (fix #1768)

### LINKED ISSUE #1768: Infinite loop on uncaught error when using createLocalVue()
### Subject of the issue

There is an infinite loop in the vue-test-utils error handler when using `createLocalVue` more than once. Everything works correctly for the first test, but subsequent tests that use their own local vue instance will cause massive noise in the console
I have tested this with version 1.1.2

### Steps to reproduce

- Create a component with an uncaught error
- Mount the component using a distinct local vue in beforeEach (createLocalVue)
- Have more than one test

### Expected behaviour

Each test should run, and produce only the errors triggered by the code

### Actual behaviour

You should see an error similar to this in the console:
```
  console.error node_modules/vue/dist/vue.runtime.common.dev.js:621
    [Vue warn]: Error in config.errorHandler: "RangeError: Maximum call stack size exceeded"

  console.error node_modules/vue/dist/vue.runtime.common.dev.js:1884
    RangeError: Maximum call stack size exceeded
        at Array.push (<anonymous>)
        at getParent (...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2799:17)
        at findAllParentInstances (...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2805:3)
        at errorHandler (...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2912:32)
        at ...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2929:5
        at Array.forEach (<anonymous>)
        at errorHandler (...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2928:26)
        at ...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2929:5
        at Array.forEach (<anonymous>)
        at errorHandler (...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2928:26)
        at ...\node_modules\@vue\test-utils\dist\vue-test-utils.js:2929:5
       (...)
```

### Possible Solution

After debugging the code, it is apparent that the first call to createLocalVue() does not have an errorHandler set. Subsequent calls, however, have their error handler set to the one in vue-test-utils. Whenever the child component has an uncaught error, it will call the vue-test-utils handler, which will call the parent's (our local vue) handler. Since it is the same handler, with the same arguments, there is infinite recursion

There are many possible solutions, but a workaround is to simply pass an empty  errorHandler function:
```
createLocalVue({ errorHandler: () => {}})
```

antoinerey:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

- This PR fixes the issue #1768. 
- It looks like it's caused by Vue instance being polluted between tests (more details on the issue itself: https://github.com/vuejs/vue-test-utils/issues/1768#issuecomment-776926894).

### DIFF
--- a/test/specs/create-local-vue.spec.js
+++ b/test/specs/create-local-vue.spec.js
@@ -231,4 +231,27 @@ describeWithShallowAndMount('createLocalVue', mountingMethod => {
       }
     }
   )
+
+  // This test is related to the issue 1768.
+  // See: https://github.com/vuejs/vue-test-utils/issues/1768
+  itSkipIf(
+    process.env.TEST_ENV === 'browser' || vueVersion < 2.6,
+    'Does not exceed maximum call stack size when no custom error handler is defined',
+    async () => {
+      const { error } = global.console
+      const spy = jest.spyOn(global.console, 'error')
+      const localVue = createLocalVue()
+
+      try {
+        mountingMethod(ComponentWithSyncError, { localVue })
+      } catch {
+        const expected = expect.stringMatching(
+          /Maximum call stack size exceeded/
+        )
+
+        global.console.error = error
+        expect(spy).not.toHaveBeenCalledWith(expected)
+      }
+    }
+  )
 })
--- a/packages/shared/create-local-vue.js
+++ b/packages/shared/create-local-vue.js
@@ -36,7 +36,7 @@ function _createLocalVue(
   instance.config = cloneDeep(Vue.config)
 
   // if a user defined errorHandler is defined by a localVue instance via createLocalVue, register it
-  instance.config.errorHandler = config.errorHandler || Vue.config.errorHandler
+  instance.config.errorHandler = config.errorHandler
 
   // option merge strategies need to be exposed by reference
   // so that merge strats registered by plugins can work properly


antoinerey:
While this PR is expected to fail, it seems that every checks are green which is quite confusing. 🤔 😁

test/specs/create-local-vue.spec.js
```diff
@@ -232,3 +233,32 @@ describeWithShallowAndMount('createLocalVue', mountingMethod => {
...
 })
+
+// TODO: Clean this up, it's only useful while debugging the issue.
+describe.only('Test', () => {
+  // When setting the array length value to `1`, then the issue disappears.
+  // However, when setting it to anything bigger, the issue appears.
+  const sets = Array.from({ length: 2 })
+
+  it.each(sets)(
+    // Note, the message may not be the final one, based on what I discover.
```
lmiller1990: can you put link to the issue? it will be difficult to remember what this test is really doing

lmiller1990:
Let's hope this does not break anything :eyes:

codfish:
@lmiller1990 what's the release schedule for the project like? When do you think we'll see this fix released into the wild?

lmiller1990:
There is no set schedule, it's just whenever I do one. I can do a release this week. Maybe we should investigate sem-release and automate things...

If no release within a few days, ping me again, but I should be able to do it soon.

PR #1792, opened 2021-02-22
docs: Add warning on docs version

afontcu:
A couple of users let me know they spent some time on VTU docs even though they were using Vue 3. I think signaling everywhere that these are docs for Vue 2 might be helpful.

disclaimer: I didn't spend much time on the actual layout, I simply kept it simple. Any suggestion is welcome ;)

imatge


lmiller1990 (approved):
Good idea

PR #1808, opened 2021-03-20
Fix VTU don't return key property to event when use trigger.('eventName.key')

voskresla:

- [x] Bugfix
- [ ] Other, please describe: fix test `convert a registered key name to a key code` in `test/specs/wrapper/trigger.spec.js`

- [ ] Yes
- [x] No

**Other information:**
If in our code exist check like this:
```
import { Key } from 'w3c-keys'
...
handleOnKeyDown(e: KeyboardEvent) {
	const tab = e.key === Key.Tab
	const escape = e.key === Key.Escape

	if (tab || escape) {
		...doSmth
	}
}
```

we can't test it by fire event like this
```
await input.trigger('keydown.tab')
```

Because VTU don't return `key` property in event when triggered by `eventName.keyName`.

Of course we can test it like this )
```
await input.trigger('keydown', {
	key: 'Tab',
})
```

### DIFF
--- a/packages/test-utils/src/create-dom-event.js
+++ b/packages/test-utils/src/create-dom-event.js
@@ -27,6 +27,7 @@ const modifiers = {
 function getOptions(eventParams) {
   const { modifier, meta, options } = eventParams
   const keyCode = modifiers[modifier] || options.keyCode || options.code
+  const key = modifier
 
   return {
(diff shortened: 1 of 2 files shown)


lmiller1990 (approved):
Thanks, sorry it took so long to review! I will do a release this week probably.

PR #1826, opened 2021-04-15
fix #1824: trigger keyboard event regression

### LINKED ISSUE #1824: trigger keybord event with key property isn't working in v1.1.4
### Subject of the issue

After the release of the 1.1.4 version, some of my tests that cover keyboard events fail. To be more specific, the tests in which I pass the key as a property of the event. This way is described in VTU documentation (https://vue-test-utils.vuejs.org/guides/dom-events.html#keyboard-example): 
```
await wrapper.trigger('keydown', {
      key: 'a'
})
```

I could replace all keys with their respective keyCodes in my tests, however since this method is specified in the official docs and it was working in version 1.1.3 and below, I belive this is a bug.

### Steps to reproduce

Let's take a simple Vue component. In template:
`<div @keyup="$_handleKeyUp"> </div>`

In script: 
```
$_handleKeyUp(event) {
            if (event.key === KEY_VALUE.SPACE || event.key === KEY_VALUE.ENTER) {
                // do sth
            }
}
```

In tests, I want to verify if that something happens. So, I trigger an event with key ENTER.

```
await root.trigger('keyup', {
           key: 'Enter'
});
```
### Expected behaviour

The test should pass.

### Actual behaviour

The test fails. This also applies to any other keys I have tested (space - ' ', left - ArrowLeft, right - ArrowRight).

It's worth mentioning that replacing key with its respective keyCode does indeed make the test pass.
```
await root.trigger('keyup', {
           keyCode: 13
});
```

Passing the key as a modifier also works:
```
await root.trigger('keyup.enter');
```

Downgrading to version 1.1.3 also makes the tests pass.
### Possible Solution

I belive this could be because of the modifications done in https://github.com/vuejs/vue-test-utils/pull/1808. 
I also noticed you don't have any test for these type of key events (with key as a parameter), or at least I could only find ones with keyCodes or modifiers; it might be worth adding a tests for this case as well to prevent further issues.

BeeMargarida:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

PR https://github.com/vuejs/vue-test-utils/pull/1808 introduced a regression where it was overriding the key with a possibly empty string. This broke the functionally of passing down a key to the trigger event. As pointed out in https://github.com/vuejs/vue-test-utils/issues/1824, also added a test to this specific use case.

### DIFF
--- a/packages/test-utils/src/create-dom-event.js
+++ b/packages/test-utils/src/create-dom-event.js
@@ -63,7 +63,7 @@ function getOptions(eventParams) {
     // Any derived options should go here
     keyCode,
     code: keyCode,
-    key
+    key: key || options.key
   }
 }
 
--- a/test/specs/wrapper/trigger.spec.js
+++ b/test/specs/wrapper/trigger.spec.js
@@ -67,6 +67,21 @@ describeWithShallowAndMount('trigger', mountingMethod => {
     })
   })
 
+  describe('causes keydown handler to fire with the appropriate key when wrapper.trigger("keydown", { key: "k" }) is fired on a Component', async () => {
+    const keydownHandler = jest.fn()
+    const wrapper = mountingMethod(ComponentWithEvents, {
+      propsData: { keydownHandler }
+    })
+
+    await wrapper.find('.keydown').trigger('keydown', { key: 'k' })
+
+    const keyboardEvent = keydownHandler.mock.calls[0][0]
+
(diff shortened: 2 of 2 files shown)


lmiller1990 (approved):
Great catch, thanks!

PR #1832, opened 2021-04-23
-added a test to check the error from the issue 1820

Bastczuak:
…happen.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -59,7 +59,7 @@ export function matches(node, selector) {
   // ignore the functional component if its a RouterView
   // Find a good explanation comment why tahts the case, please help!
   const componentInstance =
-    node[FUNCTIONAL_OPTIONS]?.name === 'RouterView'
+    node[FUNCTIONAL_OPTIONS] && node[FUNCTIONAL_OPTIONS]?.name === 'RouterView'
       ? node.child
       : node[FUNCTIONAL_OPTIONS] || node.child
 
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -56,7 +56,12 @@ export function matches(node, selector) {
     return element && element.matches && element.matches(selector.value)
   }
 
-  const componentInstance = node[FUNCTIONAL_OPTIONS] || node.child
+  // ignore the functional component if its a RouterView
+  // Find a good explanation comment why tahts the case, please help!
+  const componentInstance =
+    node[FUNCTIONAL_OPTIONS]?.name === 'RouterView'
+      ? node.child
+      : node[FUNCTIONAL_OPTIONS] || node.child
 
   if (!componentInstance) {
     return false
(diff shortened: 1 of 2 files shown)


test/specs/wrapper/find.spec.js
```diff
@@ -279,6 +280,68 @@ describeWithShallowAndMount('find', mountingMethod => {
...
+    const routes = [
+      {
+        path: '/a/b',
+        name: 'ab',
+        component: TestComponentToFind
+      }
+    ]
+    const router = new VueRouter({ routes })
+    // and this will not work with shallowMount because router view gets stubbed! How can I just test mount?
+    const wrapper = mountingMethod(
```
lmiller1990: `mountingMethod` is both `mount` and `shallowMount`. We run all tests for both. If you don't want to run the test for one, just import the one you want on line 2 from `packages/src/test-utils`. Or you can do a check with the `doNotTestIf` helpers.

Bastczuak: ok done that.

lmiller1990:
Thanks, I'll look at this today!

packages/test-utils/src/matches.js
```diff
@@ -56,7 +56,12 @@ export function matches(node, selector) {
     return element && element.matches && element.matches(selector.value)
   }
 
-  const componentInstance = node[FUNCTIONAL_OPTIONS] || node.child
+  // ignore the functional component if its a RouterView
+  // Find a good explanation comment why tahts the case, please help!
+  const componentInstance =
+    node[FUNCTIONAL_OPTIONS]?.name === 'RouterView'
```
lmiller1990: Can we have an extra check? Just to be sure:

`node[FUNCTIONAL_OPTIONS] && node[FUNCTIONAL_OPTIONS].name`

I am not sure if we have `?` enabled in this code base or not.

lmiller1990 (changes requested):
Please add an extra check! thanks.

packages/test-utils/src/matches.js
lmiller1990: Edit: we do **not** - that's why CI is failing to build. Let's do it old school for now with the `&&` check.

Bastczuak: okok

Bastczuak: sorry for that, it worked in the tests, that's why I assumed it's generally working.

Bastczuak:
@lmiller1990 ok I added the extra check.

Bastczuak:
@lmiller1990  Dude I am in despair... I can't understand why the karma test are failing and the jest test not. They are almost the same test case. I really really have no idea.

lmiller1990:
The only test failing is the one using the router. This PR makes a change that specifically adds a check against `RouterView` [on this line](https://github.com/vuejs/vue-test-utils/pull/1832/files#diff-80ab213bfb0eeafc02c6f0005a492f52eab27ef052d46ed9e17c5df0da4eefe1R62). I guess we broke something? 

I will try taking a look at this on the weekend. I'd suspect if you comment out that check, the Karma tests will pass. That's the first thing I'm going to do - figure out if this failure is due to a change we made, or not.

PR #1850, opened 2021-05-20
[WIP] fix #1839: props comparison regression

### LINKED ISSUE #1839: Props comparison regression in 1.1.3 (and later)
### Subject of the issue

After the release of 1.1.3, some tests related to props comparison started to fail. For example a test similar to the one below resulted in 

```
AssertionError [ERR_ASSERTION]: Values have same structure but are not reference-equal:
[
  'Q',
  'W',
  'E',
  'R',
  'T',
  'Y',
  'U',
  'I',
  'O',
  'P'
]
```
 
### Steps to reproduce

```
const assert = require("assert");
const testUtils = require("@vue/test-utils");
const components = require("./components");

it("should have valid characters set", () => {
        const component = testUtils(components["Keyboard"], {
            propsData: { keys: ["Q", "W", "E", "R", "T", "Y", "U", "I", "O", "P"] }
        });
        assert.deepStrictEqual(component.vm.$props.keys, [
            "Q",
            "W",
            "E",
            "R",
            "T",
            "Y",
            "U",
            "I",
            "O",
            "P"
        ]);
    });
```

### Expected behaviour

It should result in a valid comparison.

### Actual behaviour

The test fails while comparing the same array.

### Possible Solution

I checked the diff between the 1.1.2 and 1.1.3 (https://github.com/vuejs/vue-test-utils/compare/v1.1.2...v1.1.3) and some changes were made to `create-instance.js` and `wrapper.js`. I suspect the regression was introduced there.

BeeMargarida:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

As mentioned in the issue https://github.com/vuejs/vue-test-utils/issues/1839, when using the `assert` library, the props comparison that previously worked in 1.1.2 stopped working in 1.1.3. The tests added in this PR fail, however if adding them to the version 1.1.2, they work.

Tests made:
1. Run `yarn run test:unit` -> FAIL
2. Checkout to v.1.1.3 (porting the tests) and run `yarn run test:unit` -> FAIL
3. Checkout to v.1.1.2 (porting the tests) and run `yarn run test:unit` -> SUCCESS

I still don't know what caused this behaviour, but its not something related to `assert` since I maintained the same node version and dependencies installed.

### DIFF
--- a/test/specs/wrapper/props.spec.js
+++ b/test/specs/wrapper/props.spec.js
@@ -112,7 +112,7 @@ describeWithShallowAndMount('props', mountingMethod => {
     const TestComponent = {
       template: `
           <div>
-            {{ array }}
+            {{ keys }}
           </div>'
         `,
       props: {
--- a/test/specs/wrapper/props.spec.js
+++ b/test/specs/wrapper/props.spec.js
@@ -113,7 +113,7 @@ describeWithShallowAndMount('props', mountingMethod => {
       template: `
           <div>
             {{ keys }}
-          </div>'
+          </div>
         `,
       props: {
         keys: {
--- a/test/specs/wrapper/props.spec.js
+++ b/test/specs/wrapper/props.spec.js
@@ -5,6 +5,7 @@ import {
   functionalSFCsSupported
 } from '~resources/utils'
 import { itSkipIf } from 'conditional-specs'
+import { deepStrictEqual } from 'assert'
 
 describeWithShallowAndMount('props', mountingMethod => {
   it('returns true if wrapper has prop', () => {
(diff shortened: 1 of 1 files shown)


BeeMargarida:
Update: The problem was introduced with 2 changes (related to each other):
1. https://github.com/vuejs/vue-test-utils/blob/7e78174672321882d373158eb599bfbd6e9584a2/packages/create-instance/create-instance.js#L24 `options.propsData` to `currentProps`
2. This whole block: https://github.com/vuejs/vue-test-utils/blob/7e78174672321882d373158eb599bfbd6e9584a2/packages/create-instance/create-instance.js#L118-L124 if removed, the tests introduced in this PR pass

By changing these lines above, the current setProps behaviour will break. This was only to discover what caused the regression in this specific use case. I suppose not using the `options.propsData` directly is causing problems in the reference comparison on the props given.

Note: If I change the test to
```
const props = ["Q", "W", "E", "R", "T", "Y", "U", "I", "O", "P"];
 const component = testUtils(components["Keyboard"], {
     propsData: { keys: props }
});
assert.deepStrictEqual(component.vm.$props.keys, props);
```
where the props content is saved in a variable and used in the comparison, the tests work with the current code.

lmiller1990:
Thanks for this. I'll try to take a look soon.

lmiller1990:
I am sorry I did not get to go into this PR in more depth. I'm trying to clean up this repo, so I'm closing off all older PRs.

We can revisit improving the tests after we've gone over all the existing issues/PRs. Thanks for understanding.

PR #1857, opened 2021-06-08
fix #1845: add functional component check in component name match

### LINKED ISSUE #1820: v1.1.4 regression
### Subject of the issue

After upgrading from `v1.1.3` to `v1.1.4` some tests have started to fail.

### Steps to reproduce

```ts
const wrapper = mount(MyComponent, {
  propsData: {
    items: [
      { name: 'woof' },
      { name: 'bork' }
    ]
  }
});

const items = wrapper.findAllComponents({ name: 'MyComponentItem' });
const firstItem = wrapper.at(0);
//                           ^--- [vue-test-utils]: no item exists at 0

// ...
```

### Expected behaviour

Tests should pass as it was before the upgrade.

### Actual behaviour

Tests failed.

### Possible Solution

Downgrade to `v1.1.3`.

### LINKED ISSUE #1845: Bug introduced from 1.1.4 and later
### Subject of the issue

My unit tests suddenly started breaking from 1.1.4 and later

### Steps to reproduce

since updating to 1.1.4 (or greater) my unit tests are now breaking - I have a custom component (renderer) which renders form components (e.g. input, text area, checkbox etc) dynamically using a config provided via props. The component iterates through the config and renders the inputs (mapping them to appropriate vue components) - now in my test when i do the following:

```
...
const dynamicFormConfig = ['text-area', 'text-input'];

const options = merge({
    propsData: {
      dynamicFormConfig,
    },
    store,
    localVue,
  }, params.options);

textInput = mount(RendererComponent, options).findComponent({` name: 'MyTextInput' });

expect(textInput.exists()).toBe(true); 
```
i should get true, but now my tests fail with false.

### Expected behaviour

I should get true from the test expectation but now im getting false. When i downgrade to 1.1.3 everything works fine and the tests pass.

### Actual behaviour

The test now fails. (i.e. textInput.exist() returns false)

### Possible Solution

Revert the breaking change introduced in 1.1.4 (Djaler:fix/find-functional-component)

### LINKED ISSUE #1854: when the prop name of the component is an array, throw error from 1.1.4 and later
### Subject of the issue

My unit tests started breaking from 1.1.4 and later

### Steps to reproduce

```javascript
    const wrapper = mount(
      {
        props: {
          name: ["start", "end"]
        },
        template: `<Wrapper><Inner :name="name" /></Wrapper>`
      },
      {
        stubs: {
          Wrapper,
          Inner
        }
      }
    );

    // name is array
    // from 1.1.4 throw error: TypeError: str.replace is not a function
    console.log(wrapper.findComponent(Wrapper));
```
vue-test
[Demo](https://codesandbox.io/s/priceless-paper-dnpkc?file=/src/demo/__tests__/demo.spec.js)

### Expected behaviour

When i downgrade to 1.1.3, it is ok, but  **findComponent({name: 'xxx'})** has error

pimlie:
The following PR introduced a bug for finding components that contain a name prop themselves: https://github.com/vuejs/vue-test-utils/commit/3cd81d0593f56034b96f368d8ab066a855e0b204#diff-80ab213bfb0eeafc02c6f0005a492f52eab27ef052d46ed9e17c5df0da4eefe1

This PR fixes that behaviour by only using `vm.name` if the component is functional

This should fix the following issues:
Resolves #1820
Resolves #1845
Resolves #1854

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/test/resources/components/component-with-name-prop.vue
+++ b/test/resources/components/component-with-name-prop.vue
@@ -1,0 +1,12 @@
+<template>
+  <div>
+    <p class="prop-name">{{ name }}</p>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'component-with-name-prop',
+  props: ['name']
+}
+</script>
--- a/packages/test-utils/src/matches.js
+++ b/packages/test-utils/src/matches.js
@@ -3,16 +3,20 @@ import {
   COMPONENT_SELECTOR,
   FUNCTIONAL_OPTIONS
 } from 'shared/consts'
-import { isConstructor } from 'shared/validators'
+import { isConstructor, isFunctionalComponent } from 'shared/validators'
 import { capitalize, camelize } from 'shared/util'
 
 function vmMatchesName(vm, name) {
   // We want to mirror how Vue resolves component names in SFCs:
   // For example, <test-component />, <TestComponent /> and `<testComponent />
   // all resolve to the same component
-  const componentName = vm.name || (vm.$options && vm.$options.name) || ''
+  const componentName = isFunctionalComponent(vm)
+    ? vm.name
+    : vm.$options && vm.$options.name
+
   return (
     !!name &&
+    !!componentName &&
     (componentName === name ||
       // testComponent -> TestComponent
       componentName === capitalize(name) ||
--- a/test/specs/wrapper/find.spec.js
+++ b/test/specs/wrapper/find.spec.js
@@ -3,6 +3,7 @@ import { createLocalVue, shallowMount } from 'packages/test-utils/src'
 import Vue from 'vue'
 import VueRouter from 'vue-router'
 import ComponentWithChild from '~resources/components/component-with-child.vue'
+import ComponentWithNameProp from '~resources/components/component-with-name-prop.vue'
 import ComponentWithoutName from '~resources/components/component-without-name.vue'
 import ComponentWithSlots from '~resources/components/component-with-slots.vue'
 import ComponentWithVFor from '~resources/components/component-with-v-for.vue'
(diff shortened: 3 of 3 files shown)


test/resources/components/component-with-name-prop.vue
```diff
@@ -0,0 +1,12 @@
+<template>
+  <div>
```
lmiller1990, line 2: Should this be a functional component? Since the patch is related to functional component name?

lmiller1990 (commented):
Left a comment. Also it looks like the OP in one of the issues posted a reproduction: https://github.com/vuejs/vue-test-utils/issues/1845#issuecomment-849472794, the test is [here](https://github.com/GeorgieH/vue-test-utils-1845/blob/master/tests/unit/components/ViewRenderer.spec.js). Could you verify if this fixes that? If we merge this and mark is as resolving those 3 issues, they'll be closed - we only want to do that if it actually is confirmed to fix those problems.

test/resources/components/component-with-name-prop.vue
pimlie: No, the patch is related to functional components but the fix is for normal components.

Eg for a functional component when you access `vm.name` you will get the name of the component. But accessing `vm.name` for a normal component will return the name property _if_ that exists. Previously the code was just too greedy by always preferring to use `vm.name` if it exists. But given the aformentioned that will result in a find bug when you try to find a normal component that has a name prop itself

pimlie:
This pr indeed fixes 1845:

image

lmiller1990:
It's out https://github.com/vuejs/vue-test-utils/releases/tag/v1.2.1

PR #1877, opened 2021-07-13
Fix #1868 v-slots now work inside a div

### LINKED ISSUE #1868: v-slot not rendering along side default slot.
### Subject of the issue

I recently switched a project from the old slot syntax to the new v-slot syntax and my snapshot started failing. Some of the slots are not getting rendered at all.

### Steps to reproduce

I have a simplified example that showcases the issue.

My component is as follows:
```vue
<template>
  <FakeComponent>
    <template v-slot:test>
      <TestComponent :msg="msg">
        <div>
          default slot of test component
        </div>
      </TestComponent>
    </template>
    <div>
      default slot of fake component
    </div>
  </FakeComponent>
</template>

<script>
export default {
  name: "HelloWorld",
  props: {
    msg: String,
  },
};
</script>

```
It is populating the test and default slot FakeComponent. According to the vue2 docs you can but do not have to specify the default slot in a template.
My test is as follows:
```js
import { shallowMount } from "@vue/test-utils";
import HelloWorld from "@/components/HelloWorld.vue";

describe("HelloWorld.vue", () => {
  it("renders props.msg when passed", () => {
    const msg = "new message";
    const wrapper = shallowMount(HelloWorld, {
      propsData: { msg },
    });
    expect(wrapper).toMatchSnapshot();
  });
});

```

### Expected behaviour

I would expect the test slot to be rendered in the snap shot as it is with the old syntax.

```js
exports[`HelloWorld.vue renders props.msg when passed 1`] = `
<fakecomponent>
  <testcomponent slot="test" msg="new message">
    <div>
      default slot of test component
    </div>
  </testcomponent>
  <div>
    default slot of fake component
  </div>
</fakecomponent>
`;
```

NOTE: to get the result above I simply changed the component to this:
```vue
<template>
  <FakeComponent>
    <TestComponent :msg="msg" slot="test">
      <div>
        default slot of test component
      </div>
    </TestComponent>
    <div>
      default slot of fake component
    </div>
  </FakeComponent>
</template>
```

### Actual behaviour

```js
exports[`HelloWorld.vue renders props.msg when passed 1`] = `
<fakecomponent>
  <div>
    default slot of fake component
  </div>
</fakecomponent>
`;
```

rikbrowning:

- [ X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X ] No

- [X ] It's submitted to the `dev` branch.
- [X ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [X ] All tests are passing: https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md#development-setup
- [ X] New/updated tests are included

- [ X] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -125,7 +125,7 @@ export function createStubFromComponent(
           ? context.children
           : this.$options._renderChildren ||
               getScopedSlotRenderFunctions(this).map(x =>
-                this.$options.parent._vnode.data.scopedSlots[x]({})
+                this.$vnode.data.scopedSlots[x]({})
               )
       )
     }
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -87,7 +87,6 @@ function getScopedSlotRenderFunctions(ctx: any): Array<string> {
   return []
 }
 
-
 export function createStubFromComponent(
   originalComponent: Component,
   name: string,
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -77,15 +77,8 @@ function getScopedSlotRenderFunctions(ctx: any): Array<string> {
   // In Vue 2.6+ a new v-slot syntax was introduced
   // scopedSlots are now saved in parent._vnode.data.scopedSlots
   // We filter out _normalized, $stable and $key keys
-  if (
-    ctx &&
-    ctx.$options &&
-    ctx.$options.parent &&
-    ctx.$options.parent._vnode &&
-    ctx.$options.parent._vnode.data &&
-    ctx.$options.parent._vnode.data.scopedSlots
-  ) {
-    const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
+  if (ctx && ctx.$vnode && ctx.$vnode.data && ctx.$vnode.data.scopedSlots) {
+    const slotKeys: Array<string> = ctx.$vnode.data.scopedSlots
     return keys(slotKeys).filter(
       x => x !== '_normalized' && x !== '$stable' && x !== '$key'
     )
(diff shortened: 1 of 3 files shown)


packages/create-instance/create-component-stubs.js
```diff
@@ -77,15 +77,8 @@ function getScopedSlotRenderFunctions(ctx: any): Array<string> {
...
-  if (
-    ctx &&
-    ctx.$options &&
-    ctx.$options.parent &&
-    ctx.$options.parent._vnode &&
-    ctx.$options.parent._vnode.data &&
-    ctx.$options.parent._vnode.data.scopedSlots
-  ) {
-    const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
+  if (ctx && ctx.$vnode && ctx.$vnode.data && ctx.$vnode.data.scopedSlots) {
```
lmiller1990, line 80: Hey, sorry for the delay on this... so basically `ctx.$options.parent._vnode.data.scopedSlots` references the same thing as `ctx.$vnode.data.scopedSlots`? Do you see any potential risk/regression here?

rikbrowning: Really sorry I am only getting to this now. They main difference is we are not going to the parent for reference of the current vnode. I am unsure if there was a specific reason we were doing this?

lmiller1990: I have no idea, this part of the code base is incredibly complex.

I wonder if we can do this in a way that changes less code to reduce risk. A lot of code bases depend on `shallowMount` and `stubs` working how they currently do (which is what this code is relate dto) and I don't want to accidental break anything.

rikbrowning: I guess this code change is address the fact that `parent` might actually be a normal DOM node and therefore not have a _vnode property. This is the case with the original bug report. Instead we are using the $vnode on the context to get access to the scopedSlots rather than relying on a `parent`

PR #1871, opened 2021-07-07
feat: add generic for wrapper.find() and wrapper.get()

standbyoneself:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [ x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ x] It's submitted to the `dev` branch.
- [x ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [x ] All tests are passing: https://github.com/vuejs/vue-test-utils/blob/dev/.github/CONTRIBUTING.md#development-setup
- [x ] New/updated tests are included

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

fix: #1870 

This feature will allow to target already the need type of the Element e.g. SVGElement, HTMLElement or other

Example: 
```js
const input = wrapper.find<HTMLInputElement>('input')
console.log(input.element.value)
```

PR #1885, opened 2021-07-23
fix: revert default element typings to `HTMLElement`

alecgibson:
The type definitions were updated to allow generic element typing in
https://github.com/vuejs/vue-test-utils/pull/1871

However, this "loosened" the default type from `HTMLElement` to
`Element. This was actually a breaking change.

For example, consumers may have had this test code:

```ts
wrapper.element.click()
```

But `click()` only exists on `HTMLElement`, not on `Element`, so test
compilation fails.

This change moves the default type back to `HTMLElement`. If we want to
loosen this requirement in the future, it should be considered a
breaking change.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/test-utils/types/index.d.ts
+++ b/packages/test-utils/types/index.d.ts
@@ -75,7 +75,7 @@ interface BaseWrapper {
   selector: Selector | void
 }
 
-export interface Wrapper<V extends Vue | null, el extends Element = Element> extends BaseWrapper {
+export interface Wrapper<V extends Vue | null, el extends HTMLElement = HTMLElement> extends BaseWrapper {
   readonly vm: V
   readonly element: el
   readonly options: WrapperOptions
@@ -83,7 +83,7 @@ export interface Wrapper<V extends Vue | null, el extends Element = Element> ext
   get<R extends Vue> (selector: VueClass<R>): Wrapper<R>
   get<R extends Vue> (selector: ComponentOptions<R>): Wrapper<R>
   get<Props = DefaultProps, PropDefs = PropsDefinition<Props>>(selector: FunctionalComponentOptions<Props, PropDefs>): Wrapper<Vue>
-  get<el extends Element>(selector: string): Wrapper<Vue, el>
+  get<el extends HTMLElement>(selector: string): Wrapper<Vue, el>
   get (selector: RefSelector): Wrapper<Vue>
   get (selector: NameSelector): Wrapper<Vue>
 
@@ -96,7 +96,7 @@ export interface Wrapper<V extends Vue | null, el extends Element = Element> ext
   find<R extends Vue> (selector: VueClass<R>): Wrapper<R>
   find<R extends Vue> (selector: ComponentOptions<R>): Wrapper<R>
   find<Props = DefaultProps, PropDefs = PropsDefinition<Props>>(selector: FunctionalComponentOptions<Props, PropDefs>): Wrapper<Vue>
-  find<el extends Element>(selector: string): Wrapper<Vue, el>
+  find<el extends HTMLElement>(selector: string): Wrapper<Vue, el>
   find (selector: RefSelector): Wrapper<Vue>
   find (selector: NameSelector): Wrapper<Vue>
 
--- a/packages/test-utils/types/test/wrapper.ts
+++ b/packages/test-utils/types/test/wrapper.ts
@@ -69,6 +69,7 @@ found = wrapper.find({ name: 'my-button' })
 selector = found.selector
 
 wrapper.find<HTMLInputElement>('input').element.value
+wrapper.find('div').element.click()
 
 let array = wrapper.findAll('.bar')
 selector = array.selector


lmiller1990:
Ah, I see. 

That said, you can find things other than HTMLElement with `find` and `get` (like SVGElement). So while this is breaking change, it's a necessary one, to make the types correct.

Maybe we should revert, then release the "loosened" types as a minor (so 1.3, in this case).

Regarding your example, is `wrapper.element.click()` a thing? I don't recall an `element` property on the root wrapper. For other elements, you could do `wrapper.find<HTMLButtonElement>` etc.

What do you think? I do agree this kind of change is a bit inconvenient, and should not have been a patch release. Even if we revert, you will have this problem when you update to the next minor release (1.3 - we cannot do a 2.x, that's already used for Vue 3).

alecgibson:
@lmiller1990 `Wrapper` exposes an `element` property [here](https://github.com/vuejs/vue-test-utils/blob/d56d9453ca7143d97c5483e287d626c999cd30bd/packages/test-utils/types/index.d.ts#L80) in the types and [here](https://github.com/vuejs/vue-test-utils/blob/d56d9453ca7143d97c5483e287d626c999cd30bd/packages/test-utils/src/wrapper.js#L34) in the actual code.

The other thing to do to not make this breaking is to just keep `HTMLElement` as the (sensible) default. Any consumers who need to override the type (eg for SVG elements — don't know how many other examples there are?) can do so explicitly (as they would need to anyway, if they need SVG-specific methods, etc.).

That being said, I can understand if this is what you want to do to be more "correct". Just a bit surprising on a patch version in particular.

As a side-note, even Microsoft's official type definitions seem to be a little bit confused on the matter of casting verbosity vs spec compliance: https://github.com/microsoft/TypeScript/issues/4689

lmiller1990:
@alecgibson seems VTU v2 (for Vue 3) is using `Element` https://github.com/vuejs/vue-test-utils-next/blob/master/src/baseWrapper.ts#L7. We should make them consistent, seems like `Element` is the best choice, but I definitely see why this was frustrating - sorry about that.

I agree doing this as a patch release was a bad idea, it should definitely have been a minor release. I'm not sure there's value in reverting it, only to re-release it as a minor in the near future.

For `get` and `find` - I wonder if we could just use the types from VTU v2 for `get` and `find`: https://github.com/vuejs/vue-test-utils-next/blob/master/src/interfaces/wrapperLike.ts - they seem a lot more well defined.

alecgibson:
Yes definitely a shame; this will definitely make the TS syntax for this library pretty verbose — (lots of `wrapper.find<HTMLElement>()` everywhere), but if that's how you all want it then it's a valid choice.

lmiller1990:
Hang on - why can't we just rework this PR to use the similar types from VTU v2?

It can actually infer the element type, eg:

```ts
mount({}).find('div').element.click()
```

image

What do you think? It won't work with `#selector`, though. So maybe it's not that useful. I do think that `Element` is more correct, and we should go with it moving forward. Apologies on the unexpected breaking change.

lmiller1990:
It won't work with `#selector`, though. So maybe it's not that useful.

image

alecgibson:
Yes it's fine if you reference by element tag, but I personally tend to avoid that to keep tests from being too fragile (and often looking things up by just eg `div` are usually not specific enough).

The scope of this PR was just to make the change un-breaking; so if you want to adopt the v3 types, that should probably be a different change.

lmiller1990:
I understand now. Fair enough - I agree that this should not have been a patch, definitely not ideal. Happy to discuss more on general type improvements, the Vue 3 lib is a good place to look for ideas since we also have a `tsd` test suite there to validate the types. If we do want to make a change like this, we should at least make it consistent across both libs. In this particular case, I think `Element` is the best choice, although I do agree it's a bit verbose. Thanks for your input and understanding.

alecgibson:
As a side-note, it sounds like using this sort of generic typing where the type isn't used in the parameter list is an [anti-pattern](https://github.com/microsoft/TypeScript/pull/42067#issuecomment-754860862), and I guess now I think about it, I don't really understand the value of doing this over just forcing consumers to cast:

```ts
const button = wrapper.find('.button') as HTMLButtonElement;
```

lmiller1990:
Hm, I don't fully understand "Type parameters that aren't used in the parameter list are an anti-pattern". Why is this an anti pattern? Does this mean we should just not have any generic here at all? I could be misunderstand this. If we have made a bad decision with this type, we can change it. We should have a good reason as to why, and make sure V2 also has the same types, or it'll be hard to upgrade.

I think the difference between casting with `as HTMLButtonElement` and the generic is when using a generic, we can enforce you must pass something that `extends Element`. Using `as XXX` allows you to write any type, which might not be valid.

alecgibson:
> Why is this an anti pattern?

I think it's an anti-pattern, because a purist would argue that this use of generics offers no "real" type safety. For example, a classic example of a generic might be the `Array`:

```ts
const arr = new Array<number>();
arr.push(1);
const item: number = arr.pop();
```

The generic here offers type-safety, because it ties everything about the class together: you definitely can only add a `number`, and always are guaranteed to get a `number`.

If the `Array` (hypothetically) let you override its generic types on its instance methods, you could write something like:

```ts
const arr = new Array(); // Defaults to any
arr.push<number>(1);
const item = arr.pop<string>();
```

...then you've not gained anything from the type system, and you're just abusing the fact that the methods let you override the expected type — essentially you've just cast the result — which then begs the question of what value is the type system adding here.

If we follow our `wrapper.find<HTMLElement>() {}` to its logical extreme, it's akin to writing:

```ts
function foo<T extends any = any>(): T {}
const bar = foo<string>();
```

...which _really_ offers no more type safety than this:

```ts
function foo(): any {}
const bar = foo() as string;
```

> Using `as XXX` allows you to write any type, which might not be valid

That is not the case:

Screen Shot 2021-07-28 at 08 38 47

**However**, all that being said, as a pragmatist, I do think if we're going to be forced to constantly cast our `find()` results, then this syntax:

```ts
wrapper.find<HTMLElement>('.foo').element.click()
```

is arguably "nicer" than this syntax:

```ts
(wrapper.find('.foo').element as HTMLElement).click()
// Or alternatively...
(<HTMLElement> wrapper.find('.foo').element).click()
```

PR #1910, opened 2021-10-03
fix: restore chaining and CSS selectors for findComponent

### LINKED ISSUE #904: Request to reinstate ability to use CSS selectors in findComponent
It said in the release notes that if we felt strongly about this feature we could open an issue, so that's what I'm doing. I don't think it makes sense to remove the ability to use CSS selectors in findComponent just to make VTU2 more in line with VTU1, and I think there are some notable benefits to using CSS selectors to find components. 

The most significant benefit (in my opinion) is the ability to more specifically select components. Say I have a component like so, that imports a Spinner component and then shows it on two buttons depending on which is clicked
```vue
<template>
  <button @click="buttonASpinning = true">A <Spinner id="btnASpinner" v-if="buttonASpinning" /></button>
  <button @click="buttonBSpinning = true">B <Spinner id="btnBSpinner" v-if="buttonBSpinning" /></button>
</template>

<script>
import { defineComponent, ref } from 'vue'
import Spinner from '@/components/Spinner.vue'

export default defineComponent({
  components: { Spinner },
  setup() {
    const buttonASpinning = ref(false)
    const buttonBSpinning = ref(false)
    return { buttonASpinning, buttonBSpinning }
  },
})
</script>
```
Without CSS selectors I have to use `findAllComponents(Spinner)[1]` to access Spinner B, which will break if the buttons switch positions, if there are any other spinners in any other components before this component on the page, etc. With large components or views this can become a nightmare of remembering state for completely unrelated parts of the page. 

It was also just nice to have another option for selecting components that didn't come with the downsides of the other three. Selecting by constructor requires you to import the constructor into your test, selecting by name requires you to know the name of the component which isn't always readily available or even present for 3rd party components, and selecting by ref requires you to add refs to components that don't already have them just for testing. Personally I like to use data attributes for selectors in tests since they can be easily removed in production.

xanf:
This PR is basically two features from `vue-test-utils-next`:
- [feat(find): allow chaining find with findComponent](https://github.com/vuejs/vue-test-utils-next/pull/897)
- also it fixes https://github.com/vuejs/vue-test-utils-next/issues/904 (having this consistent across both VTU versions will be in great help for migration

- [x] Bugfix
- [] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/test-utils/src/wrapper.js
+++ b/packages/test-utils/src/wrapper.js
@@ -16,7 +16,8 @@ import {
   isPhantomJS,
   nextTick,
   warn,
-  warnDeprecated
+  warnDeprecated,
+  isVueWrapper
 } from 'shared/util'
 import { isElementVisible } from 'shared/is-visible'
 import find from './find'
@@ -275,17 +276,6 @@ export default class Wrapper implements BaseWrapper {
     this.__warnIfDestroyed()
 
     const selector = getSelector(rawSelector, 'findComponent')
-    if (!this.vm && !this.isFunctionalComponent) {
-      throwError(
-        'You cannot chain findComponent off a DOM element. It can only be used on Vue Components.'
-      )
-    }
-
-    if (selector.type === DOM_SELECTOR) {
-      throwError(
-        'findComponent requires a Vue constructor or valid find object. If you are searching for DOM nodes, use `find` instead'
-      )
-    }
 
     return this.__find(rawSelector, selector)
   }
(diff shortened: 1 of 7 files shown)


xanf:
@lmiller1990 could you please cut a release if/when this one is merged :) Thank you

xanf:
@lmiller1990 sorry, just a friendly ping. This is a significant change greatly simplifying our migration to Vue3

packages/test-utils/src/wrapper.js
```diff
@@ -327,28 +317,25 @@ export default class Wrapper implements BaseWrapper {
...
-    return this.__findAll(rawSelector, selector)
+
+    return this.__findAll(rawSelector, selector, isVueWrapper)
   }
 
-  __findAll(rawSelector: Selector, selector: Object): WrapperArray {
+  __findAll(
+    rawSelector: Selector,
+    selector: Object,
+    filterFn: Function = () => true
```
lmiller1990, line 327: Is there any benefit to passing `filterFn` as an argument? It looks like `__findAll` is only called in one place, so `filterFn` will always be `isVueWrapper`. So should we just remove the argument entirely?

Then we would not need `filter` at all.

lmiller1990 (commented):
One comment, more of a question!

packages/test-utils/src/wrapper.js
lmiller1990: @xanf if we can resolve this one we can merge/release, can you take a look? Sorry on the delay, I've been mostly focused on V2.

xanf: @lmiller1990 

~~:heavy_check_mark: Done, sorry for delay, was focused on `v1 -> v2` VTU migration :)~~

This one is actually used, I missed it :) It is used here: https://github.com/vuejs/vue-test-utils/pull/1910/files#diff-5c5637ddf50cba4f86e96ebb7854bb165fadf227f94d056fef19923be0a60e48R321

The reason for this is that we need to filter only component instances there :)

After releasing this I will do a big check-up if there are some low-hanging fruits in terms of v1/v2 behavior differences :)

PR #2013, opened 2022-09-30
feat: make isVisible more comprehensive (backport from v2) (fix #1456)

### LINKED ISSUE #1456: isVisible doesn't check visibility due to CSS classes
Making test for components like this
```
<template>
    <div>
        <h1>Header</h1>
        <button class="hidden">Test</button>
    </div>
</template>

<script>
  export default {
    name: "Foo"
  }
</script>

<style scoped>
    .hidden {
        display: none;
    }
</style>
```
```
describe('Foo.vue', () => {
  test('Button', () => {
    const wrapper = shallowMount(Foo, {});
    console.log(wrapper.isVisible());
    const but = wrapper.find('button');
    console.log(window.getComputedStyle(but.element).display);
    console.log(but.isVisible());
  });
});
```

Call of isVisible will be the same as before

IGx89:
Enhancement to isVisible to take all styling into account. This is a backport of https://github.com/vuejs/test-utils/pull/454.

One notable change between this and the version in v2 is that it only works on elements that are part of the DOM (part of an attached tree). This is due to getComputedStyle being defined as only working in that scenario, per https://lists.w3.org/Archives/Public/www-style/2018May/0031.html. Chrome and Safari match the spec, but other browsers (including jsdom) allow it to still work. v2 only runs under jsdom while v1 can run under Chrome, which is why I had to narrow it here to only work when the element is part of the DOM. I imagine someday jsdom will implement that behavior as well, in which case v2+ will need to function this way as well.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

CC @lmiller1990

### DIFF
--- a/packages/shared/is-visible.js
+++ b/packages/shared/is-visible.js
@@ -5,13 +5,20 @@
  */
 
 function isStyleVisible(element) {
-  const { display, visibility, opacity } = element.style
+  if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) {
+    return false
+  }
+
+  // Per https://lists.w3.org/Archives/Public/www-style/2018May/0031.html
+  // getComputedStyle should only work with connected elements.
+  const { display, visibility, opacity } = element.isConnected
+    ? getComputedStyle(element)
+    : element.style
   return (
     display !== 'none' &&
     visibility !== 'hidden' &&
     visibility !== 'collapse' &&
-    opacity !== '0' &&
-    opacity !== 0
+    opacity !== '0'
   )
 }
 
--- a/packages/shared/is-visible.js
+++ b/packages/shared/is-visible.js
@@ -18,7 +18,8 @@ function isStyleVisible(element) {
     display !== 'none' &&
     visibility !== 'hidden' &&
     visibility !== 'collapse' &&
-    opacity !== '0'
+    opacity !== '0' &&
+    opacity !== 0
   )
 }
 
--- a/test/specs/wrapper/isVisible.spec.js
+++ b/test/specs/wrapper/isVisible.spec.js
@@ -176,4 +176,19 @@ describeWithShallowAndMount('isVisible', mountingMethod => {
 
     expect(wrapper.find('.child.ready').isVisible()).toEqual(true)
   })
+
+  it('returns false if element has class with opacity: 0', async () => {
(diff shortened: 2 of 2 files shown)


IGx89:
@lmiller1990, just making sure this is still on your radar

packages/shared/is-visible.js
```diff
@@ -5,13 +5,20 @@
...
+  // getComputedStyle should only work with connected elements.
+  const { display, visibility, opacity } = element.isConnected
+    ? getComputedStyle(element)
+    : element.style
   return (
     display !== 'none' &&
     visibility !== 'hidden' &&
     visibility !== 'collapse' &&
-    opacity !== '0' &&
-    opacity !== 0
```
lmiller1990: We don't need this last check anymore? The `!== 0`?

lmiller1990 (commented):
One question. Will reply a bit slowly, sorry about that.

packages/shared/is-visible.js
IGx89: I removed it since you removed it in v2 (https://github.com/vuejs/test-utils/commit/9f343b049a815c9832445067b9a419fbc448d428), but can easily add it back if you'd feel more comfortable doing that? It appears to always be a string value, but can't say 100% sure.

IGx89: @lmiller1990 in the interest of getting this merged ASAP (it's the last thing blocking my project's migration to vitest) I've added back the `opacity !== 0` check. Let me know if this looks good now!

lmiller1990: Hi, I'm sorry about the slow reply, I was away from computer/internet during Oct, but I am back now.

I agree, let's get this merged.

IGx89: 🎊  No problem, thanks a lot!

lmiller1990: It's out, grab `@vue/test-utils@1.3.2` or `@vue/test-utils@legacy`.

lmiller1990: https://github.com/vuejs/vue-test-utils/releases/tag/v1.3.2

PR #2025, opened 2022-11-04
fix #2024 - Accessing `HTMLElement` and `SVGElement` through `window`

### LINKED ISSUE #2024: Error: `ReferenceError: SVGElement is not defined` when `isVisible()` is called on an SVG element.
### Subject of the issue

A test will fails and throw `ReferenceError: SVGElement is not defined` when calling `isVisible()` on an SVG element. 

### Expected behaviour

Can correctly verify if an SVG Element is visible.

### Actual behaviour

Throws an error when verifying if an SVG Element is visible.

### Possible Solution

The issue was introduced [here](https://github.com/vuejs/vue-test-utils/pull/2013) with the purpose to make `isVisible` more comprehensive. This was ported [from jest-dom](https://github.com/testing-library/jest-dom) as it is declared on the function comment but looking at js-dom repository they don't have `instanceof SVGElement` but, instead `instanceof window.SVGElement` which makes sense. Also, I have tried this fix and it worked so I will open PR with this suggested change and let you guys decide.

3rdvision:
Parent issue: https://github.com/vuejs/vue-test-utils/issues/2024

The method `isStyleVisible` was ported [from jest-dom](https://github.com/testing-library/jest-dom) as it is declared on the function comment but looking at js-dom code, they don't have `instanceof SVGElement` but, instead, `instanceof window.SVGElement` which makes sense and fixes issue #2024.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/test-utils/dist/vue-test-utils.iife.js
+++ b/packages/test-utils/dist/vue-test-utils.iife.js
@@ -9015,10 +9015,7 @@ var VueTestUtils = (function (exports, Vue, vueTemplateCompiler) {
    */
 
   function isStyleVisible(element) {
-    if (
-      !(element instanceof window.HTMLElement) &&
-      !(element instanceof window.SVGElement)
-    ) {
+    if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) {
       return false
     }
 
--- a/packages/test-utils/dist/vue-test-utils.iife.js
+++ b/packages/test-utils/dist/vue-test-utils.iife.js
@@ -9015,7 +9015,10 @@ var VueTestUtils = (function (exports, Vue, vueTemplateCompiler) {
    */
 
   function isStyleVisible(element) {
-    if (!(element instanceof HTMLElement) && !(element instanceof SVGElement)) {
+    if (
+      !(element instanceof window.HTMLElement) &&
+      !(element instanceof window.SVGElement)
+    ) {
       return false
     }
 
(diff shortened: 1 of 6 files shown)


packages/test-utils/dist/vue-test-utils.iife.js
```diff
@@ -9015,7 +9015,10 @@ var VueTestUtils = (function (exports, Vue, vueTemplateCompiler) {
    */
```
lmiller1990: We usually don't modify the dist files by hand. Can you revert those? I think we only need to change is-visible.js.

The dist files are generated and committed when we do a release. Thanks.

lmiller1990 (commented):
One comment. thanks!

packages/test-utils/dist/vue-test-utils.iife.js
3rdvision: Got it :+1:

lmiller1990:
Cool, thanks.

I will do a release in a few days, I want to try and get a couple more things in, thanks for your patience 👍

lmiller1990:
Sorry this took a long time, life got busy. I did a release with this patch: https://github.com/vuejs/vue-test-utils/releases/tag/v1.3.4

3rdvision:
> Sorry this took a long time, life got busy. I did a release with this patch: https://github.com/vuejs/vue-test-utils/releases/tag/v1.3.4

No problem. Thank you!

PR #2068, opened 2023-04-30
fix: make scoped slots rendering consistent for stubs

xanf:
This MR fixes extremely weird stubbing behavior. Currently scoped slot stubs will be rendered **only** for components, which are located at root level of other component. Basically:

```html
<some-cmp>foo</some-cmp>
```
will have scoped slots rendered on stub, while

```html
<div><some-cmp>foo</some-cmp>
````

will have only default slot rendered. 
This is caused because slot lookup was performed on **parent** instead of instance itself

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

This is definitely a breaking change: now we try to render more scoped slots than before. Basically, that means that some tests which rely that slots were not previously rendered will fail. However, I see this definitely a better thing than current confusing approach

**Other information:**

### DIFF
--- a/packages/create-instance/create-component-stubs.js
+++ b/packages/create-instance/create-component-stubs.js
@@ -77,16 +77,8 @@ function getScopedSlotRenderFunctions(ctx: any): Array<string> {
   // In Vue 2.6+ a new v-slot syntax was introduced
   // scopedSlots are now saved in parent._vnode.data.scopedSlots
   // We filter out _normalized, $stable and $key keys
-  if (
-    ctx &&
-    ctx.$options &&
-    ctx.$options.parent &&
-    ctx.$options.parent._vnode &&
-    ctx.$options.parent._vnode.data &&
-    ctx.$options.parent._vnode.data.scopedSlots
-  ) {
-    const slotKeys: Array<string> = ctx.$options.parent._vnode.data.scopedSlots
-    return keys(slotKeys).filter(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/create-instance/create-component-stubs.js [hunk 1]
```diff
@@ -130,9 +122,15 @@ export function createStubFromComponent(
         context
           ? context.children
           : this.$options._renderChildren ||
-              getScopedSlotRenderFunctions(this).map(x =>
-                this.$options.parent._vnode.data.scopedSlots[x]({})
-              )
+              getScopedSlotRenderFunctions(this)
+                .map(x => {
+                  let result = null
+                  try {
```
xanf, line 128: This `try catch` is best-effort attempt to mitigate migration from existing state. Since we're now rendering more scoped slots on stubs than previously there is higher chance of blow up. 

While it might create some confusion (hey, some my slots are rendering, some are not) considering close EOL of Vue.js 2 I'm thinking of this one as acceptable trade-off

For example at GitLab codebase, without `try - catch` this MR entirely blows over 700 tests, with try catch - less than 100, however I'm open for discussion on removing it

lmiller1990:
Sorry, just saw this one @xanf. I'll review it now.

lmiller1990 (approved):

